PR #27, opened 2023-04-09
Refactor main desktop view

imGok:
- Add padding to main view (side bar + chat list) and side view (chat details)
- Change background colors
- Add chat header for desktop with tab name 
- Add expandable tile to chat list 

(Added space with Element)
image


lib/pages/chat_list/chat_list_body.dart
```diff
@@ -69,6 +71,138 @@ class ChatListViewBody extends StatelessWidget {
                   ActiveFilter.messages,
                 }.contains(controller.activeFilter) &&
                 client.storiesRooms.isNotEmpty;
+            if (PlatformInfos.isDesktop || PlatformInfos.isWeb) {
+              return ExpandablePanel(
```
hoangdat, line 75: Should be separate in other widget.

linux/flutter/generated_plugin_registrant.cc
```diff
@@ -35,6 +36,9 @@ void fl_register_plugins(FlPluginRegistry* registry) {
   g_autoptr(FlPluginRegistrar) flutter_secure_storage_linux_registrar =
       fl_plugin_registry_get_registrar_for_plugin(registry, "FlutterSecureStorageLinuxPlugin");
   flutter_secure_storage_linux_plugin_register_with_registrar(flutter_secure_storage_linux_registrar);
+  g_autoptr(FlPluginRegistrar) flutter_webrtc_registrar =
+      fl_plugin_registry_get_registrar_for_plugin(registry, "FlutterWebRTCPlugin");
+  flutter_web_r_t_c_plugin_register_with_registrar(flutter_webrtc_registrar);
```
hoangdat, line 41: can you explain more about this change?

linux/flutter/generated_plugins.cmake
```diff
@@ -9,6 +9,7 @@ list(APPEND FLUTTER_PLUGIN_LIST
   emoji_picker_flutter
   file_selector_linux
   flutter_secure_storage_linux
+  flutter_webrtc
```
hoangdat, line 12: idem

linux/flutter/generated_plugin_registrant.cc
imGok: it seems to have been generated when I compiled for desktop applications, I'm not sure if it's needed

hoangdat: Please remove it

lib/config/themes.dart
```diff
@@ -128,6 +128,9 @@ abstract class FluffyThemes {
           onSecondary: brightness == Brightness.light
               ? AppColor.secondaryPrimary
               : AppColor.colorDEEEFF,
+          onPrimaryContainer: brightness == Brightness.light
+              ? Colors.white
+              : const Color.fromARGB(255, 25, 25, 26)
```
sherlockvn, line 133: You should convert color to same format, the heximal form

lib/pages/chat_list/chat_list_view/chat_list_view_style.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:flutter/material.dart';
+
+class ChatListViewStyle {
+  static BorderRadius containerBorderRadius = BorderRadius.circular(16);
```
sherlockvn, line 4: You can use like this to take advantage of const
  static const BorderRadius containerBorderRadius = BorderRadius.all(Radius.circular(16));

lib/config/themes.dart
imGok: I can't use heximal, flutter only accepts integer in parameter I can convert it to 32 bits if you prefer

hoangdat:
> overflowed issue

IMO, you have too little items in the list

hoangdat:
IMO, you missed my comment: https://github.com/linagora/twake-on-matrix/pull/27#issuecomment-1502739712.

linux/flutter/generated_plugin_registrant.cc
imGok: done it

lib/config/themes.dart
imGok: done it

PR #61, opened 2023-04-11
#31 fix: reply text style

### LINKED ISSUE #31: [Chat] [ChatView] The text of reply message from sender is not correct design
OS: Android / IOS
Device: Mobile
Twake 1.10.0

*Precondition: 
- There is existed a message conversation
- There is existed a replied message

### **Actual:** 
The text of reply message is having:
- color = #1b1b1f
- size = 13.9
- weight,style,height = NULL
Image

### **Expected:** 
The text of reply message is having:
- color = #000000
- size = 16
- weight,style,height = 400,normal,20

Image

imGok:
This Closes #31

Light : 
imageDark : 
image


hoangdat:
please ref the ticket in commit message

hoangdat:
can you demo your modification?

PR #62, opened 2023-04-11
fix: fab buttons size icon

imGok:


lib/pages/chat_list/select_chat_list_fab/select_chat_list_fab.dart
```diff
@@ -53,24 +53,18 @@ class _SelectChatListFloatingActionButtonState extends State<SelectChatListFloat
               TwakeFloatingButton(
                 buttonText: L10n.of(context)!.groups,
                 svgString: CustomSVGIcons.groupsIcon,
-                svgHeight: SelectChatListFabStyle.fabButtonSize,
```
sherlockvn, line 56: Why do you remove height and width ? I think if you want to have consistent size of fab icon, you can you BoxFit.fitHeight like my new PR

sherlockvn (approved):

lib/pages/chat_list/select_chat_list_fab/select_chat_list_fab.dart
hoangdat: can you ref you PR? @sherlockvn

PR #65, opened 2023-04-13
Feature/change bundle id ios

ManhNTX:


ios/Runner/Info.plist
```diff
@@ -6,6 +6,8 @@
 	<array>
 		<string>im.fluffychat.app</string>
```
hoangdat, line 7: why not change it?

ManhNTX: because this is id of TaskSchedulerPermitted

hoangdat (approved):

PR #83, opened 2023-05-13
Discovery

hoangdat:


lib/data/network/interceptor/dynamic_url_interceptor.dart
```diff
@@ -0,0 +1,19 @@
...
+
+class DynamicUrlInterceptors extends InterceptorsWrapper {
+  String? _baseUrl;
+
+  void changeBaseUrl(String? url) {
+    _baseUrl = url;
+  }
+
+  String? get baseUrl => _baseUrl;
+
```
sherlockvn, line 11: IMO, use String? baseUrl for simplicity.

.github/workflows/tests.yaml
```diff
@@ -54,7 +54,9 @@ jobs:
           ssh-private-key: ${{ secrets.SSH_KEY }}
 
       - name: Run widget test
-        run: flutter test
+        run: |
```
sherlockvn, line 57: typo error ?

lib/data/network/interceptor/dynamic_url_interceptor.dart
hoangdat: 👍

PR #128, opened 2023-06-02
TW-111 scroll down after sending a message

### LINKED ISSUE #111: Auto scroll down after sent a message in Chat

imGok:
This Closes #111

scroll_down

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -309,6 +309,8 @@ class ChatController extends State<Chat> {
       });
 
   Future<void> send() async {
+    scrollDown();
+
     if (sendController.text.trim().isEmpty) return;
     _storeInputTimeoutTimer?.cancel();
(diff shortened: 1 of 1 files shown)


lib/pages/chat/chat.dart
```diff
@@ -349,6 +349,8 @@ class ChatController extends State<Chat> {
       editEvent = null;
       pendingText = '';
     });
+
+    scrollDown();
```
sherlockvn, line 353: I think put it in first line of the send() method have better UX because inside send() method also have await, so as a user i usually want to move right after i click send

sherlockvn (approved):

lib/pages/chat/chat.dart
hoangdat: same idea for me

hoangdat (approved):

Issue #102, opened 2023-05-19
Apply new design for App Bar in Chat screen


hoangdat:
### DoD:
- [x] Typing
- [ ] Waiting for network status in No Connection
- [x] Online status: online, offline, offline for how long (in Dialog)

imGok:
did not find design for Typing app bar could you provide one please

PR #137, opened 2023-06-06
TW-102 - App Bar Style Offline & Typing

imGok:
- Add No Connection on app bar 
- Change typing animation
This is related to #102 

offlinetwake
image

### DIFF
--- a/lib/pages/chat/chat_app_bar_title.dart
+++ b/lib/pages/chat/chat_app_bar_title.dart
@@ -1,9 +1,11 @@
 import 'package:fluffychat/pages/chat/chat_app_bar_title_style.dart';
 import 'package:fluffychat/utils/room_status_extension.dart';
 import 'package:fluffychat/utils/string_extension.dart';
+import 'package:fluffychat/widgets/twake_components/twake_loading/twake_loading_indicator.dart';
 import 'package:flutter/material.dart';
 
 import 'package:flutter_gen/gen_l10n/l10n.dart';
+import 'package:linagora_design_flutter/linagora_design_flutter.dart';
 import 'package:matrix/matrix.dart';
 import 'package:vrouter/vrouter.dart';
 
@@ -100,7 +102,7 @@ class ChatAppBarTitle extends StatelessWidget {
                     overflow: TextOverflow.ellipsis,
                     style: Theme.of(context).textTheme.titleMedium?.copyWith(
                       color: Theme.of(context).colorScheme.onSurface,
-                      letterSpacing: ChatAppBarTitleStyle.letterSpacingRoomName
+                      letterSpacing: ChatAppBarTitleStyle.letterSpacingRoomName,
                     ),
                   ),
                   _buildStatusContent(context, room),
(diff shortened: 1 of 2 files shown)


lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -113,37 +142,49 @@ class ChatAppBarTitle extends StatelessWidget {
     );
   }
 
-  _buildStatusContent(BuildContext context, Room room) {
+  bool hasNoConnection(SyncStatusUpdate status) {
+    return status.status == SyncStatus.error &&
+        (status.error?.exception is SocketException ||
+            status.error?.exception is MatrixConnectionException);
```
hoangdat, line 148: why not use `[connection_pls](https://pub.dev/packages/connectivity_plus)`? it will be a passive listener.

hoangdat: right here we need to depend on the `sync update event`? How the `sync` request work? Repeat request to `sync`? What is the interval?

hoangdat: Moreover, how about reuse it in other screen?

lib/pages/chat/chat_app_bar_title_style.dart
```diff
@@ -12,4 +12,7 @@ class ChatAppBarTitleStyle {
 
   static double get letterSpacingRoomName => 0.15;
   static double get letterSpacingStatusContent => 0.5;
+
+  static Color get flashingCircleBrightColor => const Color(0xFFcad1d9);
```
hoangdat, line 16: it is 
image

We also defined it in LinagoraSysColor. Please use it
```dart
LinagoraSysColors.material().tertiary
```
or
```dart
LinagoraSysColors.material().tertiaryDark
```

lib/widgets/twake_components/twake_loading/twake_loading_indicator.dart
```diff
@@ -2,23 +2,23 @@ import 'package:fluffychat/widgets/twake_components/twake_loading/status_loading
 import 'package:flutter/widgets.dart';
 import 'package:flutter/cupertino.dart';
 
-class TypingIndicator extends StatefulWidget {
-  const TypingIndicator({
+class TwakeLoadingIndicator extends StatefulWidget {
```
hoangdat, line 5: Can we do it as a `StatelessWidget`?

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -13,19 +18,44 @@ import 'package:fluffychat/utils/adaptive_bottom_sheet.dart';
...
+  @override
+  void dispose() {
+    _onSyncSub.cancel();
+    super.dispose();
+  }
+
   @override
   Widget build(BuildContext context) {
-    final room = controller.room;
+    final client = Matrix.of(context).client;
```
hoangdat, line 49: how about to use `StreamBuilder`?

hoangdat:
- For this PR: only keep the typing, online/offline status of user
- Create another PR: for network connection

lib/widgets/twake_components/twake_loading/twake_loading_indicator.dart
imGok: Because it's an animation widget with differents states, I followed the official tutorial https://docs.flutter.dev/cookbook/effects/typing-indicator

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -113,37 +115,34 @@ class ChatAppBarTitle extends StatelessWidget {
     );
   }
 
-  _buildStatusContent(BuildContext context, Room room) {
+  Widget _buildStatusContent(BuildContext context, Room room) {
+    final TextStyle statusTextStyle = Theme.of(context).textTheme.labelSmall!.copyWith(
```
sherlockvn, line 119: you should use with `labelSmall?.copyWith....`, style of Text can have null

hoangdat (approved):

PR #138, opened 2023-06-06
fix: check if scroll is attached before scrolling on sending

imGok:
- Check if the elements are attached to scroll view before scrolling (preventing bugs)

Chat without message
scrollsansmessage

Chat with messages
scrollavecmessages


hoangdat:
please dont forget to have the demostration

hoangdat (approved):

PR #139, opened 2023-06-06
Style Chat List Item

imGok:
- Add correct M3 styles
- Fix Unread colors on text content and last event timer

correctstylestile

### DIFF
--- a/assets/l10n/intl_ar.arb
+++ b/assets/l10n/intl_ar.arb
@@ -637,7 +637,7 @@
             "link": {}
         }
     },
-    "isTyping": "يكتب…",
+    "isTyping": "يكتب",
     "@isTyping": {
         "type": "text",
         "placeholders": {}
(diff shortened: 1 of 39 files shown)


assets/l10n/intl_ar.arb
```diff
@@ -637,7 +637,7 @@
             "link": {}
         }
     },
-    "isTyping": "يكتب…",
+    "isTyping": "يكتب",
```
hoangdat, line 640: so interesting, what is it? why you change like that?

hoangdat: you want to put the text stable and animate the flashingCircle?

imGok: I removed every "..." because these strings as used in chat list item (with ...) and in app bar (with flashing circle) so it's easier to manage like so

hoangdat:
Screenshot 2023-06-08 at 16 29 22

PR #144, opened 2023-06-08
feat: add No Connection feature on app bar

imGok:
- Add Network Connectivity Service
- Update rich presence status badge according to connection
- Update app bar status bar according to connection

offlinefeature

### DIFF
--- a/lib/di/global/get_it_initializer.dart
+++ b/lib/di/global/get_it_initializer.dart
@@ -1,5 +1,6 @@
 import 'package:fluffychat/di/global/configuration/tom_configuration_di.dart';
 import 'package:fluffychat/di/global/hive_di.dart';
+import 'package:fluffychat/di/global/network_connectivity_di.dart';
 import 'package:fluffychat/di/global/network_di.dart';
 import 'package:get_it/get_it.dart';
 
@@ -17,6 +18,7 @@ class GetItInitializer {
   void setUp() {
     NetworkDI().bind();
     HiveDI().bind();
+    NetworkConnectivityDI().bind();
     ToMConfigurationDI().bind();
   }
 }
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1,7 +1,9 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:fluffychat/di/global/get_it_initializer.dart';
 import 'package:fluffychat/pages/forward/forward.dart';
+import 'package:fluffychat/utils/network_connection_service.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
@@ -46,6 +48,8 @@ class Chat extends StatefulWidget {
 }
 
 class ChatController extends State<Chat> {
+  final NetworkConnectionService networkConnectionService = getIt.get<NetworkConnectionService>();
+
   Room? room;
 
   Client? sendingClient;
--- a/lib/pages/chat/chat_app_bar_title.dart
+++ b/lib/pages/chat/chat_app_bar_title.dart
@@ -1,3 +1,4 @@
+import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:fluffychat/pages/chat/chat_app_bar_title_style.dart';
 import 'package:fluffychat/utils/room_status_extension.dart';
 import 'package:fluffychat/utils/string_extension.dart';
@@ -68,27 +69,6 @@ class ChatAppBarTitle extends StatelessWidget {
                     ),
                   ),
                 ),
-                Positioned(
-                  right: 0,
-                  top: 0,
-                  child: room.isDirectChat == true
-                      ? Container(
-                          width: ChatAppBarTitleStyle.statusSize,
-                          height: ChatAppBarTitleStyle.statusSize,
-                          decoration: BoxDecoration(
-                            border: Border.all(
-                              color: ChatAppBarTitleStyle.statusBorderColor,
-                              width: ChatAppBarTitleStyle.statusBorderSize,
-                            ),
-                            color:
-                                room.directChatPresence?.currentlyActive == true
-                                    ? ChatAppBarTitleStyle.currentlyActiveColor
-                                    : ChatAppBarTitleStyle.currentlyInactiveColor,
-                            shape: BoxShape.circle,
-                          ),
-                        )
-                      : const SizedBox.shrink(),
-                ),
               ],
             ),
             const SizedBox(width: 12),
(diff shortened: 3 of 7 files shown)


lib/data/hive/hive_collection_network_database.dart
```diff
@@ -0,0 +1,139 @@
...
+  static Future<HiveCollectionNetworkDatabase> databaseBuilder() async {
+    Logs().d('Open Hive for Network Connectivity...');
+    HiveAesCipher? hiverCipher;
+    try {
+      // Workaround for secure storage is calling Platform.operatingSystem on web
+      if (kIsWeb) throw MissingPluginException();
+
+      const secureStorage = FlutterSecureStorage();
+      final containsEncryptionKey =
+          await secureStorage.read(key: FlutterHiveCollectionsDatabase.cipherStorageKey) != null;
```
nqhhdev, line 35:
```suggestion
      final containsEncryptionKey = await secureStorage.read(key: FlutterHiveCollectionsDatabase.cipherStorageKey) != null;
```

lib/data/hive/hive_collection_network_database.dart
```diff
@@ -0,0 +1,139 @@
...
+
+      // workaround for if we just wrote to the key and it still doesn't exist
+      final rawEncryptionKey = await secureStorage.read(key: FlutterHiveCollectionsDatabase.cipherStorageKey);
+      if (rawEncryptionKey == null) throw MissingPluginException();
+
+      hiverCipher = HiveAesCipher(base64Url.decode(rawEncryptionKey));
+    } on MissingPluginException catch (_) {
+      const FlutterSecureStorage()
+          .delete(key: FlutterHiveCollectionsDatabase.cipherStorageKey)
+          .catchError((_) {});
```
nqhhdev, line 54:
```suggestion
      const FlutterSecureStorage()
        .delete(key: FlutterHiveCollectionsDatabase.cipherStorageKey)
        .catchError((_) {});
```

lib/data/hive/hive_collection_network_database.dart
```diff
@@ -0,0 +1,139 @@
...
+      Logs().i('Hive encryption is not supported on this platform');
+    } catch (e, s) {
+      const FlutterSecureStorage()
+          .delete(key: FlutterHiveCollectionsDatabase.cipherStorageKey)
+          .catchError((_) {});
```
nqhhdev, line 59: duplicate lines 52-54. 

- We can write func handle this case and can be reuse it? 
- What do you think?

lib/utils/network_connection_service.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:fluffychat/data/hive/hive_collection_network_database.dart';
+import 'package:fluffychat/data/hive/network/network_connectivity_hive_obj.dart';
+import 'package:fluffychat/di/global/get_it_initializer.dart';
+import 'package:matrix/matrix.dart';
+
+class NetworkConnectionService {
+  final Connectivity connectivity;
+  bool enableShowToastDisconnected = true;
+  late StreamSubscription<ConnectivityResult> subscription;
+  late final HiveCollectionNetworkDatabase _networkDatabase = getIt.get<HiveCollectionNetworkDatabase>();
```
nqhhdev, line 13: This is why it doesn't work. 

Pls check again

lib/widgets/fluffy_chat_app.dart
```diff
@@ -36,12 +38,19 @@ class FluffyChatApp extends StatefulWidget {
 class FluffyChatAppState extends State<FluffyChatApp> {
   bool? columnMode;
   String? _initialUrl;
+  late final networkConnectionService = getIt.get<NetworkConnectionService>();
 
   @override
-  void initState() {
+  void initState() async {
```
nqhhdev, line 44: In Flutter, the initState method is not allowed to be declared as async because it needs to return void and cannot be marked as async. initState is a lifecycle method in Flutter that is called when the widget is inserted into the widget tree and is typically used for initialization tasks.

lib/widgets/fluffy_chat_app.dart
```diff
@@ -36,12 +38,19 @@ class FluffyChatApp extends StatefulWidget {
...
     super.initState();
-    _initialUrl =
-        widget.clients.any((client) => client.isLogged()) ? '/rooms' : '/home';
+    _initialUrl = widget.clients.any((client) => client.isLogged()) ? '/rooms' : '/home';
+    networkConnectionService.onInit();
+  }
+
+  @override
+  void dispose() async {
```
nqhhdev, line 51: same `initState`

hoangdat:
Why we need to store connection state? We use it for what?

lib/di/global/get_it_initializer.dart
```diff
@@ -15,8 +16,9 @@ class GetItInitializer {
   GetItInitializer._internal();
 
   void setUp() {
-    NetworkDI().bind();
```
hoangdat, line 18: Wwhy neeed to remove it?

lib/pages/chat/chat.dart
```diff
@@ -136,6 +138,8 @@ class ChatController extends State<Chat> {
 
   bool showEmojiPicker = false;
 
+  final NetworkConnectionService networkConnectionService = getIt.get<NetworkConnectionService>();
```
hoangdat, line 141: I don't see the usage

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -122,30 +123,48 @@ class ChatAppBarTitle extends StatelessWidget {
...
-        overflow: TextOverflow.ellipsis,
-        style: statusTextStyle,
-      );
-    } else {
-      return Row(
-        children: [
-          Text(
-            room.getLocalizedTypingText(context),
+    return StreamBuilder(
+      stream: controller.networkConnectionService.getStreamInstance(),
```
hoangdat, line 127: Why not put the service directly?

lib/utils/network_connection_service.dart
```diff
@@ -0,0 +1,60 @@
...
+    _listenNetworkConnectionChanged();
+  }
+
+  void onDispose() {
+    networkStreamController.close();
+    networkStreamSubscription.cancel();
+  }
+
+  void _setupNetworkStream() {
+    networkStreamController = StreamController<ConnectivityResult>.broadcast();
```
hoangdat, line 24: Why we need it? Why not directly listen from connectivity?

lib/di/global/get_it_initializer.dart
imGok: I just change the order because Hive db should be in first

lib/pages/chat/chat.dart
imGok: mistake from cleaning

lib/pages/chat/chat_app_bar_title.dart
imGok: you mean in the widget ?

lib/utils/network_connection_service.dart
imGok: Because onConnectivityChanged stream only emits when there is a change. When we change view, the stream needs to have the current status. So IMO it's better to have one stream that is synchronized then copied in every streambuilder (which manage his own lifeycle) we need.

We could used checkConnectivity() to initialize and merge with onChanged but then we would have needed FutureBuilder in a streambuilder in every widget which is not convenient.

I'm open to any suggestions

lib/di/global/get_it_initializer.dart
hoangdat: No. Be careful in the order of DI. The best way is don't touch

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -115,37 +102,86 @@ class ChatAppBarTitle extends StatelessWidget {
     );
   }
 
-  Widget _buildStatusContent(BuildContext context, Room room) {
+  StreamBuilder<ConnectivityResult> _buildStatusBadge(Room room) {
```
hoangdat, line 105: remove it, because ATM, we dont have it

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -115,37 +102,86 @@ class ChatAppBarTitle extends StatelessWidget {
...
-    } else {
-      return Row(
-        children: [
-          Text(
-            room.getLocalizedTypingText(context),
+    return StreamBuilder<ConnectivityResult>(
+      stream: controller.networkConnectionService.getStreamInstance(),
+      builder: (context, snapshot) {
+        if (!snapshot.hasData) {
+          return const SizedBox.shrink();
```
hoangdat, line 147: what happen if no event in connectivity? we can not show the typing,...?

hoangdat (approved):

lib/data/hive/hive_collection_network_database.dart
```diff
@@ -0,0 +1,139 @@
...
+    }
+    Logs().d('Hive for ToM is ready');
+    return db;
+  }
+
+  static Future<String> _findDatabasePath({String? path = 'tom_db'}) async {
+    if (!kIsWeb) {
+      Directory directory;
+      try {
+        if (Platform.isLinux) {
```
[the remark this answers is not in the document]
sherlockvn: oh sorry, i review but forget to push comments

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -115,37 +95,51 @@ class ChatAppBarTitle extends StatelessWidget {
...
-        style: statusTextStyle,
-      );
-    } else {
-      return Row(
-        children: [
-          Text(
-            room.getLocalizedTypingText(context),
+    return StreamBuilder<ConnectivityResult>(
+      stream: controller.networkConnectionService.getStreamInstance(),
+      builder: (context, snapshot) {
```
hoangdat, line 107: what happen if connection state not change, but the typing/online state change?

imGok: After trying differents scenarios there is indeed a memory leak when disconnecting because typing ephemeral variable is not resynchronized. I'm working on it

imGok: bug is unconsistent and seems related to spam disconnect / reconnect. I got no issue when user is disconnecting while someone is typing 
disconnectduringtyping

Typing when disconnected : 
typingduringdisconnected

hoangdat: WHAT??? why we need to care about `typing` when disconnected?

hoangdat: not only testing on the `disconnected`, please test on the `connected`

hoangdat: please focus on the stream

hoangdat:
1. Connection available: typing/online/offline state

2. Connection not available: `no netwrok connection`

hoangdat (approved):

PR #147, opened 2023-06-09
TW-66 Enable Firebase Push notification for Android

hoangdat:
- change appId 
- hardcode pushGateway
- change parse method of message from Push 
- ADR 
- notification icon

WIP DEMO:

https://github.com/linagora/twake-on-matrix/assets/6462404/81e020ae-268a-4578-ae1d-ffa41a7bc6bb


lib/config/app_config.dart
nqhhdev, line 1: i have a question:
- Why we don't create an env for config?

nqhhdev (approved):

lib/config/app_config.dart
hoangdat: the code base :))))

PR #153, opened 2023-06-12
TW-129 : feat: add avatar on contact tab

### LINKED ISSUE #129: Update avatar for Contact in Contact/Chat/ContactPicker Screen
### Desc:
- Contact screen connect directly to ToM so all the contact don't have avatar correctly
- Because we also reuse avatar for each contact we work with many times, we should cache this globally
- What is `MxcImage`?

### DoD:
- [ ] all contact screens need to have avatar correctly: `New Private Chat`, `New Group Chat`, `Contact tab`
- [ ] avatar cache is applied for all screens
- [x] add postman collection: done by Khai

imGok:
avatarcontact

This Closes #129

### DIFF
--- a/lib/pages/new_private_chat/widget/expansion_contact_list_tile.dart
+++ b/lib/pages/new_private_chat/widget/expansion_contact_list_tile.dart
@@ -1,8 +1,9 @@
 import 'package:fluffychat/presentation/model/presentation_contact.dart';
 import 'package:fluffychat/pages/new_private_chat/widget/contact_status_widget.dart';
+import 'package:fluffychat/widgets/avatar/avatar.dart';
+import 'package:fluffychat/widgets/matrix.dart';
 import 'package:fluffychat/widgets/twake_components/twake_chip.dart';
 import 'package:flutter/material.dart';
-import 'package:linagora_design_flutter/avatar/round_avatar.dart';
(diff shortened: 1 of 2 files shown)


lib/pages/new_private_chat/widget/expansion_contact_list_tile.dart
```diff
@@ -21,15 +21,19 @@ class ExpansionContactListTile extends StatelessWidget {
...
         crossAxisAlignment: CrossAxisAlignment.start,
         children: [
-          RoundAvatar(
-            text: contact.displayName ?? "@",
-            size: 48,
+          FutureBuilder<Uri?>(
+            future: getAvatarUrl(context),
+            builder: (context, snapshot) {
+              return Avatar(mxContent: snapshot.data, name: contact.displayName);
+            },
```
hoangdat, line 32: Be careful with the `FutureBuilder`, it maybe makes the blink blink effect

sherlockvn (approved):

lib/pages/new_private_chat/widget/expansion_contact_list_tile.dart
imGok: Mxcimage puts a default avatar before data is loaded. It's already handled
I could remove the fade animation if you want tho

blinkavatar

hoangdat (approved):

PR #194, opened 2023-06-22
hot-fix/fix error cannot send images in IOS

sherlockvn:


hoangdat (approved):

lib/domain/usecase/send_images_interactor.dart
```diff
@@ -17,6 +17,7 @@ class SendImagesInteractor {
         await room.sendImageFileEvent(
           txIdMapToImageInfo.value1[txId]!,
           fakeImageEvent: txIdMapToImageInfo.value2[txId],
+          shrinkImageMaxDimension: 1600,
```
hoangdat, line 20: Wwhat is it?

sherlockvn: we can remove it

Issue #159, opened 2023-06-16
Cache the profile of users  in storage


hoangdat:
### Desc:
because client use the profile of user in many places. Nice to cache it in local. And also get the change event

- Currently we cached profile in memory 
 

### Docs:

https://spec.matrix.org/v1.6/client-server-api/#events-on-change-of-profile-information

![Image](https://user-images.githubusercontent.com/6462404/245759990-9311f4dd-c741-49c3-bc27-105597a38176.png)

### DoD:
- [ ] profiles are cached
- [ ] profiles cached will be updated when receive events on change of profile

Connected for example to changing avatar:
When one of the users changes his avatar another one doesn't see the changes when his application is not open during the change

shepilov:
@hoangdat is it still relevant?

DianaPotokina:
@hoangdat Dat? Is it still relevant?

pm-McFly:
An implementation is needed for this as it can became really heavy on network usage if fetching users avatar too often...

hoangdat:
We need to verify:
- [ ] we don't download profile of user frequently (every time to open chat with this guy, ...) 
- [ ] well update after some guys update profile (username, avatar)

PR #195, opened 2023-06-22
WIP fix: fetch avatar after update

imGok:
`client.dart - getProfileFromUserId()` only caches the profile on initialization, the avatar becomes outdated after changing during the session

Solution : need to do our own cache profiles (#159)


hoangdat:
hi @imGok, in my point of view, you can listen the `Presence` event via `/sync` method to update the cache.

lib/widgets/matrix.dart
```diff
@@ -455,6 +464,14 @@ class MatrixState extends State<Matrix> with WidgetsBindingObserver {
     createVoipPlugin();
   }
 
+  void initProfilesService() {
```
sherlockvn, line 467: you didn't call this method

hoangdat:
TBD: Group Info first -> select invite.

PR #259, opened 2023-07-03
TW-245: fix-bug/remove blink-blink effect when mentions user or room

sherlockvn:
https://github.com/linagora/twake-on-matrix/assets/43041967/ba6e1d5c-cbcc-4f32-b04b-253078acfa61

https://github.com/linagora/twake-on-matrix/assets/43041967/fe532df2-aa8a-4995-a844-71b9da934762

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -1,8 +1,10 @@
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/config/setting_keys.dart';
+import 'package:fluffychat/pages/chat/chat.dart';
 import 'package:fluffychat/pages/image_viewer/image_viewer.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/matrix_locals.dart';
 import 'package:fluffychat/utils/url_launcher.dart';
+import 'package:fluffychat/widgets/pill.dart';
 import 'package:flutter/material.dart';
 
 import 'package:flutter_gen/gen_l10n/l10n.dart';
@@ -15,14 +17,18 @@ class HtmlMessage extends StatelessWidget {
   final String html;
   final int? maxLines;
   final Room room;
+  final Event event;
   final TextStyle? defaultTextStyle;
   final TextStyle? linkStyle;
   final double? emoteSize;
   final Widget? bottomWidgetSpan;
+  final ChatController? chatController;
 
   const HtmlMessage({
     Key? key,
     required this.html,
+    this.chatController,
+    required this.event,
     this.maxLines,
     required this.room,
     this.defaultTextStyle,
@@ -163,6 +169,16 @@ class HtmlMessage extends StatelessWidget {
         }
         return {};
       },
-    );
+      pillBuilder: (identifier, url, onTap, getMxcUrl) {
+        if (chatController == null) {
+          return null;
+        }
+        return Pill(
+          identifier: identifier, 
+          chatController: chatController!,
+          url: url, 
+          onTap: onTap,
+        );
+      });    
   }
 }
--- a/lib/widgets/pill.dart
+++ b/lib/widgets/pill.dart
@@ -1,0 +1,61 @@
+import 'package:fluffychat/pages/chat/chat.dart';
+import 'package:fluffychat/presentation/extensions/room_extension.dart';
+import 'package:fluffychat/widgets/matrix.dart';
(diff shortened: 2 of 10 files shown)


lib/pages/chat/events/html_message.dart
```diff
@@ -163,6 +169,20 @@ class HtmlMessage extends StatelessWidget {
         }
         return {};
       },
-    );
+      pillBuilder: (identifier, url, onTap, getMxcUrl) {
+        debugPrint('identifier: $identifier, url: $url');
+        if (event!.status == EventStatus.sending) {
```
nqhhdev, line 174: if `event == null` what is happen?

lib/pages/chat/events/html_message.dart
```diff
@@ -163,6 +169,20 @@ class HtmlMessage extends StatelessWidget {
...
+          chatController!.storePillSendingEventId(event!.eventId);
+        }
+        if (event!.status == EventStatus.sending || chatController!.isSyncEventForSendingPill(event: event!)) {
```
nqhhdev, line 177: case A: ` if (event!.status == EventStatus.sending)`
case B: ` if (event!.status == EventStatus.sending || chatController!.isSyncEventForSendingPill(event: event!))`

If (event!.status == EventStatus.sending) Will this case happen first? Not clear to me, plss specifically explain to me

lib/pages/chat/events/html_message.dart
sherlockvn: event can't be null, that is for sure, `event` is nullable because other class also use this HtmlMessage and don't need event

lib/pages/chat/events/html_message.dart
sherlockvn: When the event's status is 'Sending,' it indicates that the message is currently being sent. I need to store its eventId, and it will later appear in the 'synced' event once the message is successfully sent. This is why I have 'Case A.' 'Case B' refers to two scenarios: when the event is in the 'Sending' status or when its status is 'synced' (synced specifically for this message).

lib/pages/chat/events/html_message.dart
```diff
@@ -163,6 +169,20 @@ class HtmlMessage extends StatelessWidget {
         }
         return {};
       },
-    );
+      pillBuilder: (identifier, url, onTap, getMxcUrl) {
+        debugPrint('identifier: $identifier, url: $url');
```
hoangdat, line 173: please use the same convention log of fluffy

lib/pages/chat/events/html_message.dart
hoangdat: but still need to keep the thing in safe. respect the input to keep us safe in the future also

lib/pages/chat/events/html_message.dart
hoangdat: deserve a method for it :)

lib/presentation/mixins/pill_cache_mixin.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:matrix/matrix.dart';
+
+mixin PillCacheMixin {
+  final Map<String, dynamic> mentionCache = {};
```
hoangdat, line 4: a lot of place we cache for profile in memory. Not a safe way. please consistent it.

imGok (approved):

lib/widgets/pill.dart
```diff
@@ -0,0 +1,61 @@
...
+        decoration: BoxDecoration(
+          color: Colors.grey[800],
+          borderRadius:
+              BorderRadius.all(Radius.circular(avatarSize + padding)),
+        ),
+        child: Row(
+          crossAxisAlignment: CrossAxisAlignment.center,
+          mainAxisSize: MainAxisSize.min,
+          children: <Widget>[
+            if (avatarUrl?.toString().isNotEmpty == true)
```
dab246, line 47: **Attention**: The null variable's `toString()` function always returns the value `null`. So the application will crash when `avatarUrl=null`

```suggestion
if (avatarUrl != null && avatarUrl.toString().isNotEmpty)
```

dab246: Oh, srr, you can still write `if(avatarUrl?.toString().isNotEmpty == true)` it's perfectly true if you used the nullsafety operator `?` :))

nqhhdev:
- Pls rebase your PR

nqhhdev (approved):

Issue #202, opened 2023-06-23
Improve UX in chat screen


sherlockvn:
- When user is chatting in chat Screen, there is no way to unfocus chat input bar, except the system pop
- Overflow in empty group chat in small screen device

![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/0730b4cf-adec-4c38-8027-64ba96c9ca25)

QuangHoang1210:
Passed

PR #263, opened 2023-07-03
TW-202: Fix overflow for `DirectChatEmptyView` and `GroupChatEmptyView`

nqhhdev:
- Issue: #202 

- Resolved:

- Fix overflow for `DirectChatEmptyView` and `GroupChatEmptyView`

https://github.com/linagora/twake-on-matrix/assets/99852347/7031610f-a795-4ad9-b534-6ba4681fc26a

### DIFF
--- a/lib/pages/chat/chat_event_list.dart
+++ b/lib/pages/chat/chat_event_list.dart
@@ -1,5 +1,6 @@
 import 'package:fluffychat/pages/chat/direct_chat_empty_view.dart';
 import 'package:fluffychat/pages/chat/group_chat_empty_view.dart';
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 
 import 'package:flutter_gen/gen_l10n/l10n.dart';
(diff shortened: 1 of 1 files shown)


hoangdat (approved):

lib/pages/chat/chat_event_list.dart
```diff
@@ -41,10 +42,22 @@ class ChatEventList extends StatelessWidget {
         EventTypes.CallInvite,
       }.contains(e.type),
     )) {
-      return Center(
-        child: controller.room?.isDirectChat ?? true 
-          ? const DirectChatEmptyView()
-          : const GroupChatEmptyView(),
+      return Column(
```
dab246, line 45: IMO, we don't need `Column`

```suggestion
return Center(
  child: SingleChildScrollView(
    padding: const EdgeInsets.symmetric(vertical: 16.0),
    controller: controller.scrollController,
    physics: const ClampingScrollPhysics(),
    child: controller.room?.isDirectChat ?? true
      ? const DirectChatEmptyView()
      : const GroupChatEmptyView(),
  ),
);
```

nqhhdev: I tried it, and wrong

```
The ParentDataWidget Expanded(flex: 1) wants to apply ParentData of type FlexParentData to a RenderObject, which has been set up to accept ParentData of incompatible type BoxParentData.

Usually, this means that the Expanded widget has the wrong ancestor RenderObjectWidget. Typically, Expanded widgets are placed directly inside Flex widgets.
The offending Expanded is currently placed inside a Center widget.
```

PR #295, opened 2023-07-09
fix-bug: can't see message of user havenot accept invite

sherlockvn:
# Description:
When user A invite a user B to a direct chat, if user B have not join the chat, user B can't receive new message from user A
# Solution: 
When user B receive the invite from user A, user B will be automatically joined chat with user A. In chat list screen, only display the the direct chat have at least one message and group chat have invited

https://github.com/linagora/twake-on-matrix/assets/43041967/72c2d7ff-6df1-49d9-98fa-45eddd31eff2

### DIFF
--- a/lib/pages/chat_list/chat_list.dart
+++ b/lib/pages/chat_list/chat_list.dart
@@ -232,6 +232,17 @@ class ChatListController extends State<ChatList>
     });
   }
 
+  void joinInviteDirectChats() {
+    Future.wait(Matrix.of(context).client.rooms.map(joinRoom));
+  }
+
+  Future<void> joinRoom(Room room) async {
+    final isRoomInvited = !room.isArchived && room.membership == Membership.invite;
+    if (isRoomInvited && room.isDirectChat) {
+      await room.join();
+    }
+  }
+
   void onSearchEnter(String text) {
     if (text.isEmpty) {
       cancelSearch(unfocus: false);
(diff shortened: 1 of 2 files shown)


nqhhdev:
- Could you help me explain for issue step by step?
- Write your solution in PR, I think It will be easier to understand for reviewers 🤙

lib/pages/chat_list/chat_list.dart
```diff
@@ -232,6 +232,19 @@ class ChatListController extends State<ChatList>
     });
   }
 
+  void joinInviteDirectChats() {
+    Matrix.of(context).client.rooms.forEach((room) {
```
dab246, line 236: **Attention**: Be careful when using `forEach` for `async` functions. It may work, but sometimes the results are not what you expect.

- Can we do the `joinRoom` in parallel?

lib/pages/chat_list/chat_list.dart
```diff
@@ -232,6 +232,17 @@ class ChatListController extends State<ChatList>
     });
   }
 
+  void joinInviteDirectChats() {
+    Future.wait(Matrix.of(context).client.rooms.map(joinRoom));
```
hoangdat, line 236: I am not a fan of execute a heavy work inside a `StreamBuilder` or any `build` function, it is the messy between the UI and logic. Lagging in UI for sure

hoangdat: and, it is not a good choice, because in the future, maybe use as spam

hoangdat:
Can I close it? Khải

hoangdat:
Can I close it? Khải

PR #296, opened 2023-07-09
Fix small bug

hoangdat:
- Notification for Android
- move to Twake in some components

image

IMG_1658

### DIFF
--- a/ios/FluffyChat Share/Info.plist
+++ b/ios/FluffyChat Share/Info.plist
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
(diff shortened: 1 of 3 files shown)


nqhhdev (approved):

nqhhdev:
I think if there was a small demo it would be awesomeeeeeeeee ✋ ✋

sherlockvn (approved):

lib/config/app_config.dart
```diff
@@ -54,7 +54,7 @@ abstract class AppConfig {
   static const String pushNotificationsChannelName = 'Twake push channel';
   static const String pushNotificationsChannelDescription =
       'Push notifications for Twake';
-  static const String pushNotificationsAppId = 'com.twake.twake';
+  static const String pushNotificationsAppId = 'com.linagora.ios.twake';
```
hoangdat, line 57: please separate for android and ios

Issue #268, opened 2023-07-03
[UI] create group chat always a private


hoangdat:
### DoD:

PR #306, opened 2023-07-12
feat: remove encryption and public/private

imGok:
creategrpsettings
image
image

#268

### DIFF
--- a/lib/pages/new_group/widget/expansion_participants_list.dart
+++ b/lib/pages/new_group/widget/expansion_participants_list.dart
@@ -89,9 +89,6 @@ class _ExpansionParticipantsListState extends State<ExpansionParticipantsList> {
   void toggleExpansionList() {
     setState(() {
       isExpanded = !isExpanded;
-      if (isExpanded) {
-        widget.newGroupController.autoScrollWhenExpandParticipants();
-      }
     });
   }
 }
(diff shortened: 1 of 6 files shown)


lib/pages/new_group/widget/expansion_participants_list.dart
```diff
@@ -23,7 +23,7 @@ class ExpansionParticipantsList extends StatefulWidget {
 }
 
 class _ExpansionParticipantsListState extends State<ExpansionParticipantsList> {
-  bool isExpanded = false;
+  bool isExpanded = true;
```
sherlockvn, line 26: Why do you change it, i set it default to be false because, when have a long list, it will auto scroll down a little bit, so user can't see group image, so i set default to false. Please check the case when have a lot of contacts also

hoangdat: IMO we should found another way to solve that, we need to expand it by default

imGok: I changed it because imo it was better UI now we removed settings tiles

imGok: not working atm with many contacts

sherlockvn: wjhat is not working ?

imGok: @sherlockvn it looks like it was a build error, it's working now

imGok: expandablelist

hoangdat:
Please keep the fixed size for the `Group information`. Only scroll for the list of participant @imGok

imGok:
@hoangdat 
expandablelistaa

nqhhdev (approved):

PR #327, opened 2023-07-18
feat: change delete icon to archive

### LINKED ISSUE #326: Remove Delete Chat button and Chagne it to Archive for chats

imGok:
image

This Closes #326

### DIFF
--- a/lib/pages/chat_list/chat_list_header.dart
+++ b/lib/pages/chat_list/chat_list_header.dart
@@ -144,7 +144,7 @@ class ChatListHeader extends StatelessWidget implements PreferredSizeWidget {
                 ),
               if (selectMode == SelectMode.select)
                 IconButton(
-                  icon: const Icon(Icons.delete_outlined),
+                  icon: const Icon(Icons.archive_outlined),
                   tooltip: L10n.of(context)!.archive,
                   onPressed: controller.archiveAction,
                 ),


nqhhdev (approved):

hoangdat:
please also change the function from `delete` to `archive` also, Twake will not support `delete`

imGok:
> please also change the function from `delete` to `archive` also, Twake will not support `delete`

Current action is `room.leave()`

API call : 
```dart
  /// This API stops a user participating in a particular room.
  ///
  /// If the user was already in the room, they will no longer be able to see
  /// new events in the room. If the room requires an invite to join, they
  /// will need to be re-invited before they can re-join.
  ///
  /// If the user was invited to the room, but had not joined, this call
  /// serves to reject the invite.
  ///
  /// The user will still be allowed to retrieve history from the room which
  /// they were previously allowed to see.
  ///
  /// [roomId] The room identifier to leave.
  ///
  /// [reason] Optional reason to be included as the `reason` on the subsequent
  /// membership event.
  Future<void> leaveRoom(String roomId, {String? reason}) async {
    final requestUri = Uri(
        path: '_matrix/client/v3/rooms/${Uri.encodeComponent(roomId)}/leave');
    final request = Request('POST', baseUri!.resolveUri(requestUri));
    request.headers['authorization'] = 'Bearer ${bearerToken!}';
    request.headers['content-type'] = 'application/json';
    request.bodyBytes = utf8.encode(jsonEncode({
      if (reason != null) 'reason': reason,
    }));
    final response = await httpClient.send(request);
    final responseBody = await response.stream.toBytes();
    if (response.statusCode != 200) unexpectedResponse(response, responseBody);
    final responseString = utf8.decode(responseBody);
    final json = jsonDecode(responseString);
    return ignore(json);
  }
```

and archived is just a room you left 
```dart
  /// Is the room archived
  bool get isArchived => membership == Membership.leave;
```

drminh2807 (approved):

hoangdat (approved):

PR #338, opened 2023-07-20
chore: Enable flutter analyze check

drminh2807:


nqhhdev:
Why do we need?

sherlockvn:
> _ No description provided. _

Be careful to use it, because some place need to ignore linter to work as expected

scripts/code_analyze.sh
```diff
@@ -3,5 +3,6 @@ flutter pub get && flutter pub run build_runner build --delete-conflicting-outpu
 flutter pub run import_sorter:main --no-comments
 flutter format lib/ test/
 flutter pub get
-flutter analyze --no-fatal-infos --no-fatal-warnings
+flutter analyze
```
dab246, line 6: IMO, `flutter analyze --no-fatal-infos --no-fatal-warnings` should be left as is. 
Treating `infos` or `warnings` as `fatal` is too strict at a time when the project is still in its early stages.

dab246 (commented):
Suggest using [`GitHub Action — Analyze Dart`](https://github.com/marketplace/actions/analyze-dart) to analyze the project.

hoangdat:
waiting for `super-thursday`: 3 Aug 23

tk-nguyen:
FYI, the CI is running analyze on every PR, so in a way we're already using code analysis

Issue #276, opened 2023-07-04
[M1] Reduce the padding for message bubble when select messages


hoangdat:

hoangdat:
waiting for designer

nqhhdev:
@hoangdat any update on this ticket from Desinger?

blockchain-advisory:
I don't really see how design can help in this case, when you select multiple messages, don't change anything. Just preserve the same paddings inside the bubble as well as the space between bubbles.

hoangdat:
### DoD:
- padding between bubble in selection mode should be the same in normal mode (no select)

PR #341, opened 2023-07-20
TW-276: Remove padding when selected msg

nqhhdev:
- Issues: 
  - #276 
  - Can't show camera when clicking on camera icon

- Resolved :

https://github.com/linagora/twake-on-matrix/assets/99852347/85d77f0c-5321-408c-b2f7-72ae32b6a48d

### DIFF
--- a/lib/pages/chat/events/message.dart
+++ b/lib/pages/chat/events/message.dart
@@ -481,7 +481,7 @@ class Message extends StatelessWidget {
 
   Widget _messageSelectedWidget(BuildContext context, Widget child) {
     return Container(
-      padding: EdgeInsets.symmetric(horizontal: selected ? 8 : 0, vertical: selected ? 4 : 0),
+      padding: EdgeInsets.symmetric(vertical: selected ? 1 : 0, horizontal: selected ? 8 : 0),
       color: selected
         ? LinagoraSysColors.material().secondaryContainer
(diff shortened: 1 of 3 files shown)


lib/pages/chat/events/message.dart
```diff
@@ -481,7 +481,7 @@ class Message extends StatelessWidget {
 
   Widget _messageSelectedWidget(BuildContext context, Widget child) {
     return Container(
-      padding: EdgeInsets.symmetric(horizontal: selected ? 8 : 0, vertical: selected ? 4 : 0),
+      padding: EdgeInsets.symmetric(vertical: selected ? 1 : 0, horizontal: selected ? 8 : 0),
```
drminh2807, line 484:
```suggestion
      padding: selected
          ? const EdgeInsets.symmetric(vertical: 1, horizontal: 8)
          : EdgeInsets.zero,
```

drminh2807 (changes requested):

lib/pages/chat/events/message.dart
nqhhdev: What is different?

drminh2807: just coding style, easier to understand, anyway use `const` is better right?

hoangdat (approved):

Issue #342, opened 2023-07-20
Implement Notification for iOS to connect directly with APN


hoangdat:
### DoD:

hoangdat:
- Show common notification
- trying to show full notification in `foreground`, `background`

PR #345, opened 2023-07-21
TW-342 handle notificaton on ios

drminh2807:
# Issue
- #342 

# Completed
- Handle notification when app in Foreground, Background and Terminated

# Result 

https://github.com/linagora/twake-on-matrix/assets/12546908/d12af5f9-7689-409b-8ecd-9dae404cab8b

### DIFF
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -6,11 +6,14 @@ let apnTokenKey = "apnToken"
 @UIApplicationMain
 @objc class AppDelegate: FlutterAppDelegate {
   var twakeApnChannel: FlutterMethodChannel?
+  var initialNotiInfo: Any?
   override func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
-    self.twakeApnChannel = createApnChannel()
+    twakeApnChannel = createApnChannel()
+    initialNotiInfo = launchOptions?[.remoteNotification]
+    
     GeneratedPluginRegistrant.register(with: self)
     if #available(iOS 10.0, *) {
       UNUserNotificationCenter.current().delegate = self
@@ -23,12 +26,17 @@ let apnTokenKey = "apnToken"
         let twakeApnChannel = FlutterMethodChannel(
             name: "twake_apn",
             binaryMessenger: controller.binaryMessenger)
-        twakeApnChannel.setMethodCallHandler({
-          (call: FlutterMethodCall, result: @escaping FlutterResult) -> Void in
-            if call.method == "getToken" {
+        twakeApnChannel.setMethodCallHandler { [weak self ] call, result in
+            switch call.method {
+            case "getToken":
                 result(UserDefaults.standard.string(forKey: apnTokenKey))
+            case "getInitialNoti":
+                result(self?.initialNotiInfo)
+                self?.initialNotiInfo = nil
+            default:
+                break
             }
-        })
+        }
         return twakeApnChannel
     }
     
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -47,22 +47,18 @@ let apnTokenKey = "apnToken"
     }
     
     override func userNotificationCenter(_ center: UNUserNotificationCenter,
-                                willPresent notification: UNNotification,
-      withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+                                         willPresent notification: UNNotification,
+                                         withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         let userInfo = notification.request.content.userInfo
         twakeApnChannel?.invokeMethod("willPresent", arguments: userInfo)
-        switch UIApplication.shared.applicationState {
-        case .active:
-            // Do not show notification when app in foreground
-            return completionHandler([])
-        default:
-            return completionHandler([.alert, .badge, .sound])
-        }
+        let isRemoteNotification = userInfo["aps"] != nil
+        // Hide remote noti when in foreground, decrypted noti will show by dart
+        completionHandler(isRemoteNotification ? [] : [.alert, .badge, .sound])
     }
 
     override func userNotificationCenter(_ center: UNUserNotificationCenter,
-                                didReceive response: UNNotificationResponse,
-                                withCompletionHandler completionHandler: @escaping () -> Void) {
+                                         didReceive response: UNNotificationResponse,
+                                         withCompletionHandler completionHandler: @escaping () -> Void) {
         let userInfo = response.notification.request.content.userInfo
         twakeApnChannel?.invokeMethod("didReceive", arguments: userInfo)
         completionHandler()
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -13,7 +13,7 @@ let apnTokenKey = "apnToken"
     self.twakeApnChannel = createApnChannel()
     GeneratedPluginRegistrant.register(with: self)
     if #available(iOS 10.0, *) {
-      UNUserNotificationCenter.current().delegate = self as? UNUserNotificationCenterDelegate
+      UNUserNotificationCenter.current().delegate = self
     }
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
   }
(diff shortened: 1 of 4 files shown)


nqhhdev:
IMO: pls create PR name is: `TW-342 handle notificaton on ios #345`

ios/Runner/AppDelegate.swift
```diff
@@ -37,4 +45,26 @@ let apnTokenKey = "apnToken"
         // Save the token to use in the future
         UserDefaults.standard.set(token, forKey: apnTokenKey)
     }
+    
+    override func userNotificationCenter(_ center: UNUserNotificationCenter,
+                                willPresent notification: UNNotification,
+      withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
```
hoangdat, line 51: formating

ios/Runner/AppDelegate.swift
```diff
@@ -37,4 +45,26 @@ let apnTokenKey = "apnToken"
...
+        let userInfo = notification.request.content.userInfo
+        twakeApnChannel?.invokeMethod("willPresent", arguments: userInfo)
+        switch UIApplication.shared.applicationState {
+        case .active:
+            // Do not show notification when app in foreground
```
hoangdat, line 56: why not to put back to `dart` logic to handle the same logic with android? How about the notification when you stay in chat list, in other chat?

ios/Runner/Info.plist
```diff
@@ -114,5 +114,7 @@
 	<false/>
 	<key>io.flutter.embedded_views_preview</key>
 	<true/>
+	<key>FirebaseAppDelegateProxyEnabled</key>
+	<false/>
```
hoangdat, line 118: it also work with `request Permission`?

ios/Runner/AppDelegate.swift
```diff
@@ -23,12 +26,17 @@ let apnTokenKey = "apnToken"
...
             binaryMessenger: controller.binaryMessenger)
-        twakeApnChannel.setMethodCallHandler({
-          (call: FlutterMethodCall, result: @escaping FlutterResult) -> Void in
-            if call.method == "getToken" {
+        twakeApnChannel.setMethodCallHandler { [weak self ] call, result in
+            switch call.method {
+            case "getToken":
                 result(UserDefaults.standard.string(forKey: apnTokenKey))
+            case "getInitialNoti":
+                result(self?.initialNotiInfo)
```
nqhhdev, line 34:
```
if let notificationInfo = initialNotiInfo as? [String: Any] {
    // Handle initialNotiInfo not nil
} else {
    // Handle initialNotiInfo is nil
}      
```

IMO: `result(self?.initialNotiInfo)` It can result if `initialNotiInfo ` is nil? Can you explain this case? WDYT my code ?

ios/Runner/AppDelegate.swift
nqhhdev: IMO: In foreground still shows notification and pls handle in this case

lib/utils/background_push.dart
```diff
@@ -75,25 +75,28 @@ class BackgroundPush {
...
+    } else if (Platform.isIOS) {
+      apnChannel.setMethodCallHandler((call) async {
+        Logs().v('[Push] Received APN call: $call');
+        if (call.method == 'willPresent') {
+          onReceiveNotification(call.arguments);
+        } else if (call.method == 'didReceive') {
+          onReceiveNotification(call.arguments);
+          // Go to room because user selected the notification
+          goToRoom(call.arguments['room_id']);
+        }
```
nqhhdev, line 97: - if `willPresent ` => Why need handle this case ?
- if `didReceive ` => Show local notification => Click on noti => go to Room right?

IMO: I'm not sure, can you explain this case when we handle it at `dart` layer?

lib/utils/background_push.dart
```diff
@@ -75,25 +75,28 @@ class BackgroundPush {
...
+      });
+      getIosInitialNoti();
```
nqhhdev, line 99: Why did you put it here? Why not at the top?
```
else if (Platform.isIOS) {
  getIosInitialNoti();
  apnChannel.setMethodCallHandler...
}
```

ios/Runner/AppDelegate.swift
drminh2807: done

ios/Runner/Info.plist
drminh2807: Still working

ios/Runner/AppDelegate.swift
drminh2807: checked in dart code, empty mean no initial noti

ios/Runner/AppDelegate.swift
drminh2807: done

lib/utils/background_push.dart
drminh2807: It changed, let me know if you still need explain

lib/utils/background_push.dart
drminh2807: no worries, both func just initial, one for listeners and one for initial noti
Android do same thing, no impact

hoangdat (approved):

nqhhdev (approved):

lib/utils/background_push.dart
```diff
@@ -296,9 +297,43 @@ class BackgroundPush {
...
+    Logs().d('BackgroundPush::onMessage(): $message');
+    final notification = _parseMessagePayload(message);
+    pushHelper(
+      notification,
+      client: client,
+      l10n: l10n,
+      activeRoomId: router?.currentState?.pathParameters['roomid'],
+      onSelectNotification: onSelectNotification,
+    );
+    Logs().d('BackgroundPush::onMessage(): finished pushHelper');
```
imGok, line 328: perhaps you could log when pushHelper is done with `whenComplete` ?

lib/utils/background_push.dart
```diff
@@ -75,25 +75,26 @@ class BackgroundPush {
...
       );
+      fcmSharedIsolate?.setListeners(onMessage: (message) {
+        onReceiveNotification(message);
+      });
+    } else if (Platform.isIOS) {
+      apnChannel.setMethodCallHandler((call) async {
+        Logs().v('[Push] Received APN call: $call');
+        if (call.method == 'willPresent') {
+          onReceiveNotification(call.arguments);
+        } else if (call.method == 'didReceive') {
```
imGok, line 93: IMO we should have an enumeration for these "MethodCall"

lib/utils/background_push.dart
drminh2807: @imGok  I just move the old listener handler to function to reuse

lib/utils/background_push.dart
drminh2807: I don't think it's necessary

ios/Runner/AppDelegate.swift
```diff
@@ -23,12 +26,17 @@ let apnTokenKey = "apnToken"
         let twakeApnChannel = FlutterMethodChannel(
             name: "twake_apn",
             binaryMessenger: controller.binaryMessenger)
-        twakeApnChannel.setMethodCallHandler({
-          (call: FlutterMethodCall, result: @escaping FlutterResult) -> Void in
-            if call.method == "getToken" {
+        twakeApnChannel.setMethodCallHandler { [weak self ] call, result in
```
sherlockvn, line 29: just curious, why do you use `weak` here

ios/Runner/AppDelegate.swift
```diff
@@ -37,4 +45,22 @@ let apnTokenKey = "apnToken"
...
+        // Hide remote noti when in foreground, decrypted noti will show by dart
+        completionHandler(isRemoteNotification ? [] : [.alert, .badge, .sound])
+    }
+
+    override func userNotificationCenter(_ center: UNUserNotificationCenter,
+                                         didReceive response: UNNotificationResponse,
+                                         withCompletionHandler completionHandler: @escaping () -> Void) {
+        let userInfo = response.notification.request.content.userInfo
+        twakeApnChannel?.invokeMethod("didReceive", arguments: userInfo)
+        completionHandler()
```
sherlockvn, line 64: no need to call `completionHandler` here right ?

ios/Runner/AppDelegate.swift
```diff
@@ -37,4 +45,22 @@ let apnTokenKey = "apnToken"
         // Save the token to use in the future
         UserDefaults.standard.set(token, forKey: apnTokenKey)
     }
+    
+    override func userNotificationCenter(_ center: UNUserNotificationCenter,
+                                         willPresent notification: UNNotification,
+                                         withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+        let userInfo = notification.request.content.userInfo
+        twakeApnChannel?.invokeMethod("willPresent", arguments: userInfo)
```
sherlockvn, line 53: just curious, when this method get called ?

ios/Runner/AppDelegate.swift
drminh2807: for swift, always use [weak self] in escaping closure to avoid memory leak
it won't increase the count in ARC
actually, AppDelegate is a singleton, so it isn't necessary
even if it's weak self it's okay

ios/Runner/AppDelegate.swift
drminh2807: We need it
**completionHandler**
The block to execute when you have finished processing the user’s response. You must execute this block at some point after processing the user's response to let the system know that you are done. The block has no return value or parameters.
https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649501-usernotificationcenter

ios/Runner/AppDelegate.swift
drminh2807: check should show a noti when app in foreground
```
Asks the delegate how to handle a notification that arrived while the app was running in the foreground.
```
https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter

ios/Runner/AppDelegate.swift
nqhhdev: IMO: You should write ADR for this ticket

sherlockvn (approved):

imGok (approved):

hoangdat (approved):

PR #358, opened 2023-07-23
Hotfix: Change path name for get file Android

nqhhdev:


lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
```diff
@@ -11,9 +11,8 @@ import 'package:share_plus/share_plus.dart';
 import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:fluffychat/utils/size_string.dart';
 
-const String fileName = "/storage/emulated/0/Download";
-
 extension MatrixFileExtension on MatrixFile {
+  static const String _downloadDirectoryPath = "/storage/emulated/0/Download";
```
hoangdat, line 15: maybe it only works with emulator, not work with real device,

lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
```diff
@@ -44,7 +43,7 @@ extension MatrixFileExtension on MatrixFile {
   }
 
   Future<Directory> getDownloadDirectoryAndroid() async {
-    final defaultDownloadDirectory = Directory(fileName);
+    final defaultDownloadDirectory = Directory(_downloadDirectoryPath);
```
hoangdat, line 46: should use `getDirectory....`

lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
sherlockvn: Why don't you use path_provider package to get path ?

sherlockvn (approved):

drminh2807 (approved):

hoangdat:
@nqhhdev why we need this fix? what problem?

hoangdat (approved):

dab246 (approved):
lgtm

PR #437, opened 2023-08-07
TW-418: Cannot get active room id and go to room screen with roomid

nqhhdev:
- Issue: #418 

- Resolved:

https://github.com/linagora/twake-on-matrix/assets/99852347/53a65683-c3da-4656-80be-d199d86236f6

### DIFF
--- a/docs/adr/0004-cannot-get-current-location-and-current-context.md
+++ b/docs/adr/0004-cannot-get-current-location-and-current-context.md
@@ -1,0 +1,55 @@
+# 4. Better way to get the current location and current context
+
+Date: 2023-08-08
+
+
+## Status
+
+- Issue: [#418](https://github.com/linagora/twake-on-matrix/issues/418)
+
+## Context
+
+```
+ Can't get `activeRoomId`
+ Can't go to new screen when in any state or screen position without context
+```
+
+## Root cause
+
+1. Can't get current `location`
+2. Can't get current `BuildContext`
+```
+* Reason: 
+- Cannot anymore use router.location to determine where am I in the application and there are certain scenarios 
+where I cannot provide a context to GoRouterState.of(context).location, deep link or firebase messages that 
+will pop or push some other routes from or to the navigation stack depending on the current location and state.
+```
+
+## Decision
+
+- I recommend for guys need read and understand about `go_router` and I found a similar issue raised by the community. Follow it it's really good.
+- [#12983](https://github.com/flutter/flutter/issues/129833) 
+- I think it's current workaround for this issue. Wait for the next version of `go_router` to support this issue.
+
+1. I use `GoRouter` inside `BackgroundPush` service to get current location and current context.
+
+2. Func get current `location` and get `activeRoomID`
+
+```
+String? get roomId {
+  return currentContext != null
+    ? GoRouterState.of(currentContext!).pathParameters['roomid']
+    : '';
+}
+```
+
+3. Func get current `BuildContext` with `GlobalKey<NavigatorState>` from `GoRouterDelegate` inside `GoRouter` class
+
+```
+BuildContext? get currentContext => router?.routerDelegate.navigatorKey.currentState?.context;
+```
+
+## Consequences
+
+1. User click on Notification, it not go to correct screen and no action
+2. Block func click on Notification
--- a/lib/utils/background_push.dart
+++ b/lib/utils/background_push.dart
@@ -50,7 +50,7 @@ class BackgroundPush {
   final FlutterLocalNotificationsPlugin _flutterLocalNotificationsPlugin =
       FlutterLocalNotificationsPlugin();
   Client client;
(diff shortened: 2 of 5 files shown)


lib/utils/background_push.dart
```diff
@@ -50,16 +50,15 @@ class BackgroundPush {
   final FlutterLocalNotificationsPlugin _flutterLocalNotificationsPlugin =
       FlutterLocalNotificationsPlugin();
   Client client;
-  BuildContext? context;
-  GlobalKey<NavigatorState>? router;
+  GoRouter? router;
```
hoangdat, line 53: Why we need to add this here?

lib/utils/background_push.dart
```diff
@@ -545,6 +542,14 @@ class BackgroundPush {
     }
   }
 
+  String? get roomId {
+    final lastMatch = router?.routerDelegate.currentConfiguration.last;
```
drminh2807, line 546: you should create an extension for GoRouter then we can reuse
- one func return pathParameters
- one func return currentContext

nqhhdev: IMO: IMO: Workaround is ugly. On another screen we have `context` and can use it. I think don't need it. Because in this case, It's class and doesn't have `BuildContext` so needs a global `context` in here. Just only herreee

nqhhdev: > IMO: 
- Workaround is ugly. On another screen we have `context` and can use it. I think don't need it. Because in this case, It's class and doesn't have `BuildContext` so needs a global `context` in here. Just only herreee.
- You cannot arbitrarily use `GoRouter` everywhere

lib/utils/background_push.dart
nqhhdev: IMO: 
- In this class, we need a `BuildContext` to get current `location`.
- We don't have to pass the Gouter into the class. We only need a `GlobalKey<NavigatorState> ` from `GoRouterDelegate`
- I updated it, it's very clearly pls check it again

ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -628,11 +626,10 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "Apple Development";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				CODE_SIGN_STYLE = Automatic;
```
drminh2807, line 630: What ?

lib/utils/background_push.dart
```diff
@@ -50,16 +50,15 @@ class BackgroundPush {
   final FlutterLocalNotificationsPlugin _flutterLocalNotificationsPlugin =
       FlutterLocalNotificationsPlugin();
   Client client;
-  BuildContext? context;
-  GlobalKey<NavigatorState>? router;
+  GlobalKey<NavigatorState>? globalKey;
```
drminh2807, line 53:
```suggestion
  GlobalKey<NavigatorState>? navigatorKey;
```

ios/Runner.xcodeproj/project.pbxproj
nqhhdev: REMOVED :)

lib/utils/background_push.dart
nqhhdev: IMO: 
- I think `navigatorKey ` is wrongly, just `NavigatorState `

drminh2807: GlobalKey of NavigatorState = navigatorKey

drminh2807: or just keep `router` as the origin

nqhhdev: I'll change `navigatorKey ` to `globalNavigatorKey`

docs/adr/0004-cannot-get-current-location-and-current-context.md
```diff
@@ -0,0 +1,55 @@
...
+ Can't go to new screen when in any state or screen position without context
+```
+
+## Root cause
+
+1. Can't get current `location`
+2. Can't get current `BuildContext`
+```
+* Reason: 
+- Cannot anymore use router.location to determine where am I in the application and there are certain scenarios
```
sherlockvn, line 23: I think you use chatGPT to make it easier to understand.

sherlockvn: because i can't understand what do you mean :)))

drminh2807 (approved):

drminh2807 (approved):

hoangdat (approved):

PR #443, opened 2023-08-09
Update dart lang to 3.x.x

sherlockvn:
### Update:
- update `linagora_design_flutter` to latest
- update dart language to at least 3.0.0

### DIFF
--- a/lib/mixin/comparable_presentation_search_mixin.dart
+++ b/lib/mixin/comparable_presentation_search_mixin.dart
@@ -1,6 +1,6 @@
 import 'package:fluffychat/presentation/model/search/presentation_search.dart';
 
-class ComparablePresentationSearchMixin {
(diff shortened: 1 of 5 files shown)


lib/mixin/comparable_presentation_search_mixin.dart
```diff
@@ -1,6 +1,6 @@
 import 'package:fluffychat/presentation/model/search/presentation_search.dart';
 
-class ComparablePresentationSearchMixin {
+mixin ComparablePresentationSearchMixin {
```
hoangdat, line 3: why you did it as a `class` before?

sherlockvn: because class allows more versatile than mixin, but it's not allow now

drminh2807 (approved):

hoangdat (approved):

PR #466, opened 2023-08-16
[DON"T REVIEW]TW-459: reduce frames in chat list

sherlockvn:
### Before refactoring:

https://github.com/linagora/twake-on-matrix/assets/43041967/ca29e492-5cef-4f90-a912-82714f135067

### After refactoring:

https://github.com/linagora/twake-on-matrix/assets/43041967/c28981d0-0dc5-4afa-9a5d-490fe5bbad9f

### Conclusion:
The average and max ms/frame before refactoring is lower than after refactoring, so the performance before is better

### DIFF
--- a/lib/pages/chat_list/chat_list.dart
+++ b/lib/pages/chat_list/chat_list.dart
@@ -64,9 +64,8 @@ class ChatList extends StatefulWidget {
 
 class ChatListController extends State<ChatList>
     with
-        TickerProviderStateMixin,
-        RouteAware,
-        ComparablePresentationContactMixin {
+      RouteAware,
+      ComparablePresentationContactMixin {
   StreamSubscription? _intentDataStreamSubscription;
 
   StreamSubscription? _intentFileStreamSubscription;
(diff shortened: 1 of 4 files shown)


lib/pages/chat_list/chat_list.dart
```diff
@@ -79,6 +78,8 @@ class ChatListController extends State<ChatList>
 
   String? activeSpaceId;
 
+  final chatListIndexes = <String, int>{};
```
hoangdat, line 81: what is it

sherlockvn: It's a method for optimize the chat list when chat list is reordered

sherlockvn:
> Your PR is very very hard for me.

describe your problem

nqhhdev:
>  ADR + video Demo

lib/pages/chat_list/chat_list_body.dart
```diff
@@ -31,224 +28,200 @@ class ChatListViewBody extends StatelessWidget {
     final userSearchResult = controller.userSearchResult;
     final client = Matrix.of(context).client;
 
-    return PageTransitionSwitcher(
```
nqhhdev, line 34: Why did you delete it? It's better about performance

sherlockvn: Why it's better for performance ?. I think currently we don't use the animation, so just remove it for now

PR #469, opened 2023-08-16
fix: reduce width of bubble messages

### LINKED ISSUE #411: Bubble width too large in direct chat
### Desc:

- In direct chat
- Message bubble expand too much in width

Image

### DoD:
- [ ] need to have margin left for own message 
- [ ] need to have margin right for received message

imGok:
Design : 
web version (620px)
image
mobile (272px)
image

Direct message : 
image
image

Group : 
image
image

Web :
image

This Closes #411

### DIFF
--- a/lib/pages/chat/events/message.dart
+++ b/lib/pages/chat/events/message.dart
@@ -117,8 +117,10 @@ class Message extends StatelessWidget {
               children: [
                 // if (ownMessage && event.messageType == MessageTypes.Image)
                 //   ReplyIconWidget(isOwnMessage: ownMessage),
-                Expanded(
-                  flex: MessageStyle.messageFlexMobile,
+                Container(
+                  constraints: BoxConstraints(
+                    maxWidth: MessageStyle.messageBubbleWidth(context),
+                  ),
                   child: Column(
                     crossAxisAlignment: CrossAxisAlignment.start,
                     mainAxisSize: MainAxisSize.min,
--- a/lib/pages/chat/events/message/message_style.dart
+++ b/lib/pages/chat/events/message/message_style.dart
@@ -1,5 +1,6 @@
 import 'package:fluffychat/config/app_config.dart';
 import 'package:flutter/material.dart';
+import 'package:fluffychat/utils/extension/build_context_extension.dart';
 
 class MessageStyle {
   static final bubbleBorderRadius = BorderRadius.circular(20);
@@ -47,4 +48,17 @@ class MessageStyle {
   static double get forwardContainerSize => 40.0;
   static Color? forwardColorBackground(context) =>
       Theme.of(context).colorScheme.surfaceTint.withOpacity(0.08);
+
+  static const double messageBubbleLargeMaxWidth = 620.0;
+  static const double messageBubbleWatchMaxWidth = 240.0;
+  static const double messageBubbleMobileRatioMaxWidth = 0.80;
+  static double messageBubbleWidth(BuildContext context) {
+    return context.responsiveValue<double>(
+      desktop: messageBubbleLargeMaxWidth,
+      tablet: messageBubbleLargeMaxWidth,
+      mobile:
+          MediaQuery.of(context).size.width * messageBubbleMobileRatioMaxWidth,
+      watch: messageBubbleWatchMaxWidth,
+    );
+  }
 }


sherlockvn (approved):

lib/config/themes.dart
```diff
@@ -8,6 +8,7 @@ import 'app_config.dart';
 
 abstract class FluffyThemes {
   static const double columnWidth = 360.0;
+  static const double bubbleMessageMaxWidth = 272.0;
```
hoangdat, line 11: no, I dont think it is a good solution, how about in web? bigger screen?

lib/pages/chat/events/message.dart
```diff
@@ -180,9 +180,8 @@ class Message extends StatelessWidget {
                                                       AppConfig
                                                           .bubbleSizeFactor,
                                             ),
-                                      constraints: const BoxConstraints(
-                                        maxWidth:
-                                            FluffyThemes.columnWidth * 1.5,
+                                      constraints: BoxConstraints(
+                                        maxWidth: MediaQuery.of(context).size.width * 0.8,
```
nqhhdev, line 184: IMO: Should use `Expanded(flex: ...)`, it good for this case

imGok: you are right it's more reliable

lib/pages/chat/events/message/message_style.dart
```diff
@@ -33,7 +33,10 @@ class MessageStyle {
         : const Color.fromARGB(239, 36, 36, 36);
   }
 
-  static int get messageFlexMobile => 7;
+  static int messageFlexMobile(bool othersMessageNotInGroupChat) {
+    return othersMessageNotInGroupChat ? 3 : 5;
```
nqhhdev, line 37: IMO: With your rate quite small for this monitor

sherlockvn: What about in landscape mode in mobile

imGok: IMO it's fine like this doesn't take much space in landscape mode
I did 3 for others message in group chat because there is avatar 
image
demo_msg_width222

nqhhdev: IMO: It's okay

hoangdat: hope it bigger in portrait screen. ( I talked in details with Julian in meeting)

imGok: demodemo
this is 4 instead of 3 flex

drminh2807 (approved):

lib/pages/chat/events/message/message_style.dart
hoangdat: IMO, it is still a lot of space in view. Try to keep the UI good but need to prioritize the main function of communication.
We need to align the left/right and also manage the maximum width. WDYT?

imGok: for me we should put a fixed margin and update  it if the device is mobile or web

imGok: image
image

sherlockvn (approved):

lib/pages/chat/events/message/message_style.dart
hoangdat: how about demo in web or landscape?

imGok: web version doesn't use this style, it looks like before 
image

imGok:
I removed flex property and put a fixed width that switch between web and mobile version according to figmas design, read description

sherlockvn:
In the new design, 
In Mobile portrait mode, The maxWidth for a message in chat is 272, and if it has an avatar, the maxWidth of the message also is 272.

In the Web version or mobile landscape, the maxWidth for a message is 572
- Currently, you should focus on text messages, not images or videos.
- If you find a lot of mobile devices that have a width is less than or approximately 272, I think you should use ratio. Otherwise, I think use number is fine.

imGok:
@sherlockvn based on android devices, small phones are old ones. So IMO fixed pixels width shouldn't be a problem
image
here is landscape view 
image

imGok:
web width remains 620 on figma for me 
image
I put 272 for mobile

sherlockvn (approved):

lib/pages/chat/events/message/message_style.dart
```diff
@@ -47,4 +48,17 @@ class MessageStyle {
   static double get forwardContainerSize => 40.0;
   static Color? forwardColorBackground(context) =>
       Theme.of(context).colorScheme.surfaceTint.withOpacity(0.08);
+
+  static const double messageBubbleLargeMaxWidth = 620.0;
+  static const double messageBubbleWatchMaxWidth = 240.0;
+  static const double messageBubbleMobileRatioMaxWidth = 0.80;
+  static double messageBubbleWidth(BuildContext context) {
+    return context.responsiveValue<double>(
```
hoangdat, line 56: cool function, but the breakpoint is not exactly what we define

hoangdat (approved):

Issue #502, opened 2023-08-23
[New design] Show error message


hoangdat:
<img width="824" alt="image" src="https://github.com/linagora/twake-on-matrix/assets/6462404/da711758-2398-47b3-80b7-25e99372f335">

<img width="807" alt="image" src="https://github.com/linagora/twake-on-matrix/assets/6462404/360628b8-746c-4462-91f7-c95a2cb7f4d5">

PR #519, opened 2023-08-24
TW-502 [New design] Show error message

drminh2807:
# Issue
- #502 

# Demo
Simulator Screenshot - iPhone 14 Plus - 2023-08-24 at 18 08 45

### DIFF
--- a/lib/pages/chat/seen_by_row.dart
+++ b/lib/pages/chat/seen_by_row.dart
@@ -27,34 +27,43 @@ class SeenByRow extends StatelessWidget {
       controller.timeline!,
       eventId: eventId,
     );
+    if (eventStatus == EventStatus.error) {
+      return Icon(
+        Icons.error,
+        color: Colors.red,
+        size: MessageTimeStyle.seenByRowIconSize,
+      );
+    }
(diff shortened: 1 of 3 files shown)


hoangdat (approved):

lib/pages/chat/seen_by_row.dart
```diff
@@ -27,34 +27,43 @@ class SeenByRow extends StatelessWidget {
...
+        size: MessageTimeStyle.seenByRowIconSize,
+      );
+    }
+    return Icon(
+      Icons.done_all,
+      color: timelineOverlayMessage
+          ? Colors.white
+          : Theme.of(context).colorScheme.primary,
+      size: MessageTimeStyle.seenByRowIconSize,
+    );
```
nqhhdev, line 65: IMO: Use switch case for this
WDYT ?

sherlockvn (approved):

lib/pages/chat/seen_by_row.dart
hoangdat: some condition is not relevant to the same variable. I have no problem with current code

hoangdat (approved):

PR #532, opened 2023-08-28
Added Docker image builds workflow

tk-nguyen:
2 images with be created: linagora/twake-web on Docker Hub and ghcr.io/linagora/twake-web on Github Container Registry.

Image tag with the latest commit is `main`.

Image tags with tagged release are `release` and `<semver>`.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,0 +1,115 @@
+on:
+  push:
+    branches:
+      - "main"
+    tags:
+      - "v*.*.*"
+
+name: Build Docker images
+
+permissions:
+  pull-requests: read
+  contents: read
+  packages: write
+
+jobs:
+  build_dev_image:
+    name: Build development image
+    if: github.ref_type == 'branch' && github.ref == 'refs/heads/main'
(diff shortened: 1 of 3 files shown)


hoangdat (approved):

nqhhdev (approved):

.github/workflows/image.yaml
```diff
@@ -0,0 +1,115 @@
...
+            default
+
+  build_release_image:
+    name: Build release image
+    if: github.ref_type == 'tag' && startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    environment: prod
+
+    steps:
+      - name: Set up Docker Buildx
```
drminh2807, line 73: duplicate steps
I think we should reuse jobs
https://docs.github.com/en/actions/using-workflows/reusing-workflows

sherlockvn (approved):

.github/workflows/image.yaml
tk-nguyen: No, this is for reusing whole jobs instead of just steps.

PR #543, opened 2023-08-29
TW-99: Clean ToM DB on logout

### LINKED ISSUE #99: Clean ToM DB when logout
### DoD:
- [ ] all the data of `HiveCollectionToMDatabase` must be clear after logout

imGok:
logoutremove

This Closes #99

### DIFF
--- a/lib/data/hive/hive_collection_tom_database.dart
+++ b/lib/data/hive/hive_collection_tom_database.dart
@@ -72,11 +72,8 @@ class HiveCollectionToMDatabase {
     try {
       await db.open();
     } catch (e, s) {
-      Logs().w('Unable to open Hive. Delete database and storage key...', e, s);
-      const FlutterSecureStorage()
-          .delete(key: FlutterHiveCollectionsDatabase.cipherStorageKey);
+      Logs().w('Unable to open Hive.', e, s);
       await db.clear().catchError((_) {});
-      await Hive.deleteFromDisk();
       rethrow;
     }
     Logs().d('Hive for ToM is ready');
@@ -123,6 +120,10 @@ class HiveCollectionToMDatabase {
   }
 
   Future<void> clear() async {
+    Logs().w('Delete database and storage key...');
+    await const FlutterSecureStorage()
+        .delete(key: FlutterHiveCollectionsDatabase.cipherStorageKey);
     await tomConfigurationsBox.clear();
+    await Hive.deleteFromDisk();
   }
 }
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -5,6 +5,7 @@ import 'dart:io';
 import 'package:adaptive_dialog/adaptive_dialog.dart';
 import 'package:collection/collection.dart';
 import 'package:desktop_notifications/desktop_notifications.dart';
+import 'package:fluffychat/data/hive/hive_collection_tom_database.dart';
 import 'package:fluffychat/data/network/interceptor/authorization_interceptor.dart';
 import 'package:fluffychat/data/network/interceptor/dynamic_url_interceptor.dart';
 import 'package:fluffychat/di/global/get_it_initializer.dart';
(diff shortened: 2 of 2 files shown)


nqhhdev:
Pls add demo for this case.

drminh2807 (approved):

lib/widgets/matrix.dart
```diff
@@ -186,6 +189,11 @@ class MatrixState extends State<Matrix> with WidgetsBindingObserver {
     return candidate;
   }
 
+  Future<void> logout() async {
+    await client.logout();
+    await hiveCollectionToMDatabase.clear();
```
sherlockvn, line 194: what if `hiveCollectionToMDatabase` is not available ?

lib/widgets/matrix.dart
```diff
@@ -351,13 +359,13 @@ class MatrixState extends State<Matrix> with WidgetsBindingObserver {
           ),
         );
 
-        if (state != LoginState.loggedIn) {
-          navigatorContext?.go('/rooms');
-        }
+        navigatorContext?.go('/rooms');
```
sherlockvn, line 362: what if log in failed ?

lib/widgets/matrix.dart
hoangdat: because it was init in `main`?

lib/widgets/matrix.dart
imGok: it is the case for logout with multiple clients, so if it fails it should redirect to room with another client. btw I didn't change anything on this part just removed the useless condition

hoangdat (approved):

hoangdat:
But I still see the `tom db` here, but the `fluffy` is totally cleared

image

imGok:
image

lib/widgets/matrix.dart
```diff
@@ -351,13 +353,13 @@ class MatrixState extends State<Matrix> with WidgetsBindingObserver {
...
       } else {
         setUpToMServicesInLogin(c);
         if (state == LoginState.loggedIn) {
           navigatorContext?.go('/rooms');
+        } else {
+          await getIt.get<HiveCollectionToMDatabase>().clear();
```
hoangdat, line 362: Why do it here? Why not do it when logout?

imGok: because "onLoginStateChanged" catches everytime the client logout or login, IMO it's better flow and cleaner

lib/data/hive/hive_collection_tom_database.dart
```diff
@@ -123,6 +120,10 @@ class HiveCollectionToMDatabase {
   }
 
   Future<void> clear() async {
-    await tomConfigurationsBox.clear();
```
hoangdat, line 126: why you remove it?

hoangdat: `clear` must be `clear`

lib/widgets/matrix.dart
hoangdat: 1. How `HiveCollectionsDatabase` work?

imGok: `HiveCollectionsDatabase (FlutterHiveCollectionsDatabase)` is initialized on the Matrix `createClient` function and destroyed in the logout function (`client.dart - 562 - clear()`)

We can call Client.clear

imGok: https://github.com/linagora/twake-on-matrix/pull/543/commits/d5ba17909f7b93a60f998960eba4abbbb7315ac9

lib/data/hive/hive_collection_tom_database.dart
imGok: I put clear function again

lib/widgets/matrix.dart
```diff
@@ -351,13 +353,14 @@ class MatrixState extends State<Matrix> with WidgetsBindingObserver {
...
+          await c.clear();
```
hoangdat, line 363: maybe we dont need it? WDYT?

imGok: For me we don't need it because it's already call in logout action

hoangdat: please remove it

imGok:
@hoangdat 
-> Login to Linagora
-> Logout
-> Login to tom dev
logoutandlogin

Issue #589, opened 2023-09-08
[WEB] Support download File for Web


nqhhdev:
### 
- [ ] Download a file in web

PR #565, opened 2023-09-04
TW-435: Save Gallery Action

imGok:
### Zoom/Unzoom on double tap
zooom

### Save image 
- Android
downloademo

- IOS

https://github.com/linagora/twake-on-matrix/assets/99852347/2d9763a7-3a76-4409-aa1b-81db99eaedfc

- WEB
zoeozomwebaa

downloadwebaaa

PR #593, opened 2023-09-08
TW-589 : support download file web

imGok:
depends on PR #565 
#589 
downloadfileinweb

types

### DIFF
--- a/lib/pages/chat/chat_context_menu_actions.dart
+++ b/lib/pages/chat/chat_context_menu_actions.dart
@@ -6,7 +6,7 @@ enum ChatContextMenuActions {
   copyMessage,
   pinMessage,
   forward,
-  download;
+  downloadFile;
 
   String getTitle(BuildContext context) {
     switch (this) {
@@ -18,7 +18,7 @@ enum ChatContextMenuActions {
         return L10n.of(context)!.pinMessage;
       case ChatContextMenuActions.forward:
         return L10n.of(context)!.forward;
-      case ChatContextMenuActions.download:
+      case ChatContextMenuActions.downloadFile:
         return L10n.of(context)!.downloadFile;
     }
   }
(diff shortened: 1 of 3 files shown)


nqhhdev:
PLS demo other files, ex: .pdf, .text, .odt, all file type we support

lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
```diff
@@ -38,8 +38,8 @@ extension MatrixFileExtension on MatrixFile {
     return;
   }
 
-  void downloadImageInWeb(BuildContext context) {
-    Logs().d("MatrixFileExtension()::downloadImageInWeb()::download on Web");
+  void downloadFileInWeb(BuildContext context) {
```
nqhhdev, line 41: - download file and image are the same?

- Pls create a method handle `html.AnchorElement`.

lib/pages/chat/chat_context_menu_actions.dart
```diff
@@ -5,7 +5,8 @@ enum ChatContextMenuActions {
   select,
   copyMessage,
   pinMessage,
-  forward;
+  forward,
+  download;
```
nqhhdev, line 9: What can download? file, image, ....
```suggestion
  downloadFile;
```

hoangdat:
still the same, please make user a step to choose destination.

lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
imGok: > download file and image are the same?
Since we download files from bytes, type doesn't matter IMO

> Pls create a method handle html.AnchorElement.
Why creating an other method for this part, imo it's overkill the method is readable. 
Could you give me a methods name for that ?

hoangdat:
And the demo for the permission also

nqhhdev:
> And the demo for the permission also

IMO: Don't need for web

drminh2807 (approved):

PR #602, opened 2023-09-11
config default homeserver when debug

sherlockvn:


drminh2807:
@sherlockvn  what do you think if we can use `config.json`

sherlockvn:
> config.json

i only see the config.sample.json file.

drminh2807:
> > config.json
> 
> i only see the config.sample.json file.

you only see config.sample.json file is correct
 it is template file
 common behavior when setup a project env:
 - Duplicate config.sample.json
 - Change name to config.json
 - content in config.json will override default config
 - config.json should be in .gitignore

sherlockvn:
> > > config.json
> > 
> > 
> > i only see the config.sample.json file.
> 
> you only see config.sample.json file is correct it is template file common behavior when setup a project env:
> 
> * Duplicate config.sample.json
> * Change name to config.json
> * content in config.json will override default config
> * config.json should be in .gitignore

Thanks, I try it when have time.

sherlockvn:
Closed because i handled it in another ticket

PR #607, opened 2023-09-11
Tw 499/only show accepted event defined user story

sherlockvn:
### Blocker:
https://github.com/linagora/matrix-dart-sdk/pull/12
### What i'm doing:
- As you can see, when you first create the group chat, a lot of events which is not necessary for users. So, I have to remove that when first create the group chat and direct chat
Screenshot 2023-09-16 at 00 05 42

### Solution:
- "Bạn đã tham gia cuộc trò chuyện", we can remove this event simply by checking the `senderId` and the `state_key` (receiverId). If senderId equals state_key, remove the event.
- "Bạn đã kích hoạt mã hoá đầu cuối", we can remove this event by checking type of event is `Encryption` type.
- "Bạn đã thay đổi tên của nhóm" we can remove this event by checking the previous content of event (prev_content) contains old name of room or not
- "Bạn đã thay đổi avatar của nhóm" the same
- "Bạn mời ai đó vào phòng", because the event content when first create the chat and later is not different at all. So, no condition to remove the events when create the chat. But we can do it by change the `inviteUser` APIs by adding default reason = `Welcome` and when creating the chat, the reason is null.

### Demo: 

https://github.com/linagora/twake-on-matrix/assets/43041967/36b8079f-1a7d-4792-8edd-21ef100816f8

https://github.com/linagora/twake-on-matrix/assets/43041967/e75c5f16-c702-4dae-b58a-4807303ebed6

### DIFF
--- a/lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
@@ -27,7 +27,7 @@ extension IsStateExtension on Event {
           type != EventTypes.RoomMember ||
           room.joinRules != JoinRules.public ||
           content.tryGet<String>('membership') == 'ban' ||
-          stateKey != senderId);
+      !isSomeoneChangeDisplayName() &&
 
   static const Set<String> importantStateEvents = {
     EventTypes.Encryption,
@@ -42,4 +42,9 @@ extension IsStateExtension on Event {
         EventTypes.Sticker,
         EventTypes.Encrypted
       }.contains(type);
+  bool isSomeoneChangeDisplayName() {
+    return stateKey != null &&
+        prevContent?['displayname'] != null &&
+        prevContent!['displayname'] != content['displayname'];
+  }
 }
--- a/lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
@@ -28,6 +28,7 @@ extension IsStateExtension on Event {
           room.joinRules != JoinRules.public ||
           content.tryGet<String>('membership') == 'ban' ||
       !isSomeoneChangeDisplayName() &&
+      !isSomeoneChangeAvatar() &&
 
   static const Set<String> importantStateEvents = {
     EventTypes.Encryption,
@@ -47,4 +48,11 @@ extension IsStateExtension on Event {
         prevContent?['displayname'] != null &&
         prevContent!['displayname'] != content['displayname'];
   }
+  bool isSomeoneChangeAvatar() {
+    return stateKey != null &&
+        prevContent?["membership"] == 'join' &&
+        prevContent?['avatar_url'] != null &&
+        prevContent?['avatar_url'] != content['avatar_url'];
+  }
+
 }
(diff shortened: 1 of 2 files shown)


hoangdat:
How about if someone change personal avatar?

sherlockvn:
> How about if someone change personal avatar?

Done.

assets/l10n/intl_en.arb
```diff
@@ -2701,5 +2701,6 @@
     "media": "Media",
     "files": "Files",
     "links": "Links",
-    "downloads": "Downloads"
+    "downloads": "Downloads",
+    "setMeAsAdmin": "You have been made an admin"
```
imGok, line 2705:
```suggestion
    "setMeAsAdmin": "You have been appointed admin"
    or   
    "setMeAsAdmin": "You have been made admin" 
```

lib/utils/matrix_sdk_extensions/matrix_locals.dart
```diff
@@ -311,4 +296,12 @@ class MatrixLocals extends MatrixLocalizations {
 
   @override
   String get unknownUser => l10n.user;
+
+  @override
+  String setMeAsAdmin() => l10n.setMeAsAdmin;
```
imGok, line 301: where is this used ? thought it was for toast

sherlockvn: Now, when some one set me as admin, the event "You now are the admin" in chat will be show

lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
```diff
@@ -27,19 +27,75 @@ extension IsStateExtension on Event {
...
+      !isActivateEndToEndEncryption() &&
+      !isInviteWhenCreate();
 
   static const Set<String> importantStateEvents = {
     EventTypes.Encryption,
     EventTypes.RoomCreate,
     EventTypes.RoomMember,
     EventTypes.RoomTombstone,
     EventTypes.CallInvite,
+    EventTypes.RoomName,
```
sherlockvn, line 45: Here i enable RoomName and RoomAvatar events, because if not, when someone change group name or group avatar, we can't get notify

lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
```diff
@@ -27,19 +27,75 @@ extension IsStateExtension on Event {
...
   };
 
   bool get isState => !{
         EventTypes.Message,
         EventTypes.Sticker,
         EventTypes.Encrypted
       }.contains(type);
+
+  bool isSomeoneChangeDisplayName() {
+    return stateKey != null &&
```
sherlockvn, line 56: check if someone changes their displayName, the `stateKey` is the userId of user have changed avatar, and then compare the old displayname with new displayname

lib/utils/matrix_sdk_extensions/filtered_timeline_extension.dart
```diff
@@ -27,19 +27,75 @@ extension IsStateExtension on Event {
...
+        prevContent?['displayname'] != null &&
+        prevContent!['displayname'] != content['displayname'];
+  }
+
+  bool isGroupNameChangeWhenCreate() {
```
sherlockvn, line 61: check if the event is the group name event when user create the group, so the previous name of room must be null

hoangdat (approved):

nqhhdev (approved):

nqhhdev:
Pls rebase your PR

Issue #596, opened 2023-09-10
[CHAT] Change the logic of deliver status


nqhhdev:
### Des
```
GIVEN The recipient and sender text each other
WHEN Both sent many messages at the same time
THEN Message is not correct according to status
```
### DoD

- [x] sending is clock icon
- [x] sent is 2 tick silver 
- [x]  As long as there is one reader, there will be 2 primary tick icons (No count, reader > 1)
- [x]  error is icon error

![Image](https://github.com/linagora/twake-on-matrix/assets/99852347/877ce525-a56c-48df-81cb-4ab8386153b3)

PR #628, opened 2023-09-13
TW-596: change delivery status logic

imGok:
#596 

delvierdtest

### DIFF
--- a/lib/pages/chat/seen_by_row.dart
+++ b/lib/pages/chat/seen_by_row.dart
@@ -5,8 +5,8 @@ import 'package:matrix/matrix.dart';
 enum MessageStatus {
   sending,
   sent,
-  hasBeenSeenByOne,
-  hasBeenSeenByAll,
+  hasBeenSeen,
+  error,
 }
 
 class SeenByRow extends StatelessWidget {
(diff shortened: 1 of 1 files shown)


hoangdat:
@imGok pleaes use video instead of GIF, easy to understand and scroll to the position we want

hoangdat:
https://github.com/linagora/twake-on-matrix/assets/6462404/c52d6dca-6dcd-41dd-880c-ae1a9fcffc66

something weird here. please check

imGok:
> Screen.Recording.2023-09-15.at.15.04.11.mov 
> something weird here. please check

waiting for tom server to work again because I need multiple clients

hoangdat:
> waiting for tom server to work again because I need multiple clients

can not understand, why it was `double blue` then come to `double gray` and then `double blue`. What is the problem?

imGok:
> > waiting for tom server to work again because I need multiple clients
> 
> can not understand, why it was `double blue` then come to `double gray` and then `double blue`. What is the problem?

smth crashed not related to my PR so I pressed "continue" and it looks like this in the demo. But I'll check more deeply today

drminh2807 (approved):

imGok:
I confirm it was a debug issue, tested it multiple times in build command. doesn't crash 

https://github.com/linagora/twake-on-matrix/assets/48354990/b775bf9b-3e18-4326-b583-1230f5edcf4b

hoangdat (approved):

Issue #609, opened 2023-09-12
Apply error page


nqhhdev:
### Des

- Open app with link wrong (ex: `https://..../#/abc123`, `https://..../#/rooms/12n312n3jk1/123213`) => show error page

### DoD

- [ ] Apply for new error page (404) 

#### Mobile

https://www.figma.com/file/uLw3iWnOEGaRlJnfPz4Pv0/Twake-%5BMatrix%5D?type=design&node-id=13581-2707&mode=dev

![Image](https://github.com/linagora/twake-on-matrix/assets/99852347/d8495727-fbfe-4783-beb2-e1508907d192)

#### Web

https://www.figma.com/file/uLw3iWnOEGaRlJnfPz4Pv0/Twake-%5BMatrix%5D?type=design&node-id=13581-2590&mode=dev

![Image](https://github.com/linagora/twake-on-matrix/assets/99852347/cf6d59ed-b6f6-41cd-8781-86389920b2a3)

PR #648, opened 2023-09-15
TW-609: Apply Error Page

imGok:
#609 
Web :

image

Demo video : 

https://github.com/linagora/twake-on-matrix/assets/48354990/0b9269f9-2fe2-4466-af94-556c8ca4331c
Responsive : 

https://github.com/linagora/twake-on-matrix/assets/48354990/d3c6c236-66ff-4f03-9daf-1a39100bade7

Mobile : 
image


lib/pages/chat/chat_view.dart
```diff
@@ -115,15 +116,8 @@ class ChatView extends StatelessWidget {
...
-        appBar: AppBar(
-          toolbarHeight: 120,
-          title: Text(L10n.of(context)!.oopsSomethingWentWrong),
-        ),
-        body: Center(
-          child: Text(L10n.of(context)!.youAreNoLongerParticipatingInThisChat),
-        ),
-      );
+      context.go('/error');
+      return const LinearProgressIndicator();
```
nqhhdev, line 120: Why don't you `return const SizedBox()`

lib/pages/error_page/error_page_style.dart
```diff
@@ -0,0 +1,49 @@
...
+
+class ErrorPageStyle {
+  static ColorFilter backgroundColorFilter(BuildContext context) =>
+      ColorFilter.mode(
+        Theme.of(context).colorScheme.primaryContainer,
+        BlendMode.srcATop,
+      );
+
+  static const double backgroundIconWidthWeb = 448.0;
+  static const double backgroundIconWidthMobile = 280.0;
```
nqhhdev, line 12: Screenshot 2023-09-18 at 00 52 07

web: 768 x 288

lib/pages/error_page/error_page_style.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  static const EdgeInsets textPadding = EdgeInsets.symmetric(vertical: 32.0);
+
+  static TextStyle? titleTextStyle(BuildContext context) =>
+      Theme.of(context).textTheme.headlineSmall!.copyWith(
```
nqhhdev, line 17: Screenshot 2023-09-18 at 00 54 49

M3/headline/large not headlineSmall

lib/pages/error_page/error_page_style.dart
```diff
@@ -0,0 +1,49 @@
...
+            fontWeight: FontWeight.w600,
+            color: Theme.of(context).colorScheme.onSurface,
+          );
+  static TextStyle? descriptionTextStyle(BuildContext context) =>
+      Theme.of(context).textTheme.labelLarge!.copyWith(
```
nqhhdev, line 22: M3/body/large

sherlockvn (approved):

lib/pages/chat/chat_view.dart
imGok: I could, but thought a progress was more suitable

lib/pages/error_page/error_page_style.dart
imGok: didn't see it was different styles for web and mobile

lib/pages/chat/chat_view.dart
```diff
@@ -115,15 +116,8 @@ class ChatView extends StatelessWidget {
...
-          toolbarHeight: 120,
-          title: Text(L10n.of(context)!.oopsSomethingWentWrong),
-        ),
-        body: Center(
-          child: Text(L10n.of(context)!.youAreNoLongerParticipatingInThisChat),
-        ),
-      );
+      context.go('/error');
+      return const SizedBox.shrink();
     }
```
nqhhdev, line 121: IMO: Move it to `initState`.

nqhhdev: Screenshot 2023-09-19 at 08 28 09

imGok: why is it better for you ?

nqhhdev: Using Navigator inside the build method of a Flutter widget is generally not recommended and can lead to unexpected behavior or errors. The build method should be responsible for describing the UI hierarchy based on the current state and should not trigger navigation actions directly. Instead, you should initiate navigation in response to user interactions or specific events.

Here's why it's not a good practice:

Infinite Loop: Placing navigation logic inside the build method can lead to an infinite loop. When Navigator is called, it triggers a rebuild of the widget tree. If the navigation action itself causes the widget to rebuild (for example, because it changes the route or state), it can lead to an infinite loop of rebuilds.

Build Method Frequency: The build method can be called frequently, even when there are no user interactions. Placing navigation logic inside build might cause unintended navigation actions to occur frequently.

drminh2807: @nqhhdev  agree with you

imGok: done

lib/pages/error_page/error_page.dart
```diff
@@ -0,0 +1,95 @@
...
+        automaticallyImplyLeading: false,
+        elevation: 0,
+        backgroundColor: Theme.of(context).scaffoldBackgroundColor,
+      ),
+      extendBodyBehindAppBar: true,
+      body: Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [
+            PlatformInfos.isMobile
```
drminh2807, line 25:
```suggestion
     ResponsiveUtils.isMobile
```
- Also should create private class

drminh2807 (approved):

hoangdat (approved):

Issue #580, opened 2023-09-07
`Empty Chat` display after first create chat from draft chat


hoangdat:
https://github.com/linagora/twake-on-matrix/assets/6462404/8998668a-21dc-45c9-a4ac-57f6379eea24

### DoD:
- [ ] remove `empty chat` in appbar
- [ ] fallback case from argument when create direct chat

PR #660, opened 2023-09-20
TW-580 : feat enhance chat app bar from draft

imGok:
demo with throttling (slow connection)
## Old : 

https://github.com/linagora/twake-on-matrix/assets/48354990/51f0b707-4b15-4d09-af9a-da1e1fd334c5

## New : 
Direct chat

https://github.com/linagora/twake-on-matrix/assets/48354990/04efab4e-1c16-448a-807f-47409e27162e

Room

https://github.com/linagora/twake-on-matrix/assets/48354990/625af1bb-481e-4d3f-88df-b48d65adc5ea

External ID 

https://github.com/linagora/twake-on-matrix/assets/48354990/9d95b744-e167-49b5-84f9-c32d1d1976b8

#580

### DIFF
--- a/lib/config/go_routes/go_router.dart
+++ b/lib/config/go_routes/go_router.dart
@@ -343,6 +343,15 @@ abstract class AppRoutes {
                   final extra = state.extra as ChatRouterInputArgument;
                   switch (extra.type) {
                     case ChatRouterInputArgumentType.draft:
+                      if (extra.data is String?) {
+                        return NoTransitionPage(
+                          child: Chat(
+                            roomId: state.pathParameters['roomid']!,
+                            key: Key(state.pathParameters['roomid']!),
+                            roomName: extra.data as String?,
+                          ),
+                        );
+                      }
                       return NoTransitionPage(
                         child: Chat(
                           roomId: state.pathParameters['roomid']!,
--- a/lib/domain/app_state/room/create_new_group_chat_state.dart
+++ b/lib/domain/app_state/room/create_new_group_chat_state.dart
@@ -13,11 +13,15 @@ class CreateNewGroupChatLoading extends Success {
 
 class CreateNewGroupChatSuccess extends Success {
   final String roomId;
+  final String? groupName;
 
-  const CreateNewGroupChatSuccess({required this.roomId});
+  const CreateNewGroupChatSuccess({
+    required this.roomId,
+    this.groupName,
+  });
 
   @override
-  List<Object?> get props => [roomId];
+  List<Object?> get props => [roomId, groupName];
 }
 
 class CreateNewGroupChatFailed extends Failure {
--- a/lib/pages/chat/chat_app_bar_title.dart
+++ b/lib/pages/chat/chat_app_bar_title.dart
@@ -19,11 +19,13 @@ class ChatAppBarTitle extends StatelessWidget {
   final TextEditingController sendController;
   final Stream<ConnectivityResult> getStreamInstance;
   final VoidCallback onPushDetails;
+  final String? roomName;
 
   const ChatAppBarTitle({
     Key? key,
     required this.actions,
     this.room,
+    this.roomName,
     required this.selectedEvents,
     required this.isArchived,
     required this.sendController,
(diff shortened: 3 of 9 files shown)


hoangdat (changes requested):
- what happen with the case of external matrix account? dathpham:matrix.org?
- what happen with the case of group chat?

lib/config/go_routes/go_router.dart
```diff
@@ -374,6 +375,8 @@ abstract class AppRoutes {
                         child: Chat(
                           roomId: state.pathParameters['roomid']!,
                           key: Key(state.pathParameters['roomid']!),
+                          presentationContact:
+                              extra.data as PresentationContact?,
```
drminh2807, line 379:
```suggestion
                              extra.data is PresentationContact ? extra.data : null ,
```

imGok:
@hoangdat I put demo for room aswell
For external user I didn't find a way to test it, do you know a way to text to external id ? 
But according to the code, if it doesn't have display name it will show matrix id instead

lib/domain/app_state/room/create_new_group_chat_state.dart
```diff
@@ -13,8 +13,9 @@ class CreateNewGroupChatLoading extends Success {
 
 class CreateNewGroupChatSuccess extends Success {
   final String roomId;
+  final List<String>? invite;
```
hoangdat, line 16: why we need the `list of invite` here?

hoangdat:
for the case of creating a group, you created a room with name is `test`, but when you go to chat view, name is bla bla bla. WTF?

lib/config/go_routes/go_router.dart
```diff
@@ -370,6 +371,27 @@ abstract class AppRoutes {
...
+                        return NoTransitionPage(
+                          child: Chat(
+                            roomId: state.pathParameters['roomid']!,
+                            key: Key(state.pathParameters['roomid']!),
+                            invitesIds: [contact.matrixId!],
+                          ),
+                        );
+                      }
+                      if (extra.data is List<String>?) {
+                        final membersNames = extra.data as List<String>;
```
sherlockvn, line 385: be careful, because it still can be null

lib/domain/app_state/room/create_new_group_chat_state.dart
sherlockvn: he need it to create the name for group, but I think it's not a good idea. How about the roomName directly?

lib/pages/new_group/new_group.dart
```diff
@@ -266,8 +267,18 @@ class NewGroupController extends ContactsSelectionController<NewGroup>
     );
   }
 
-  void _goToRoom(BuildContext context, String roomId) {
-    context.go("/rooms/$roomId");
+  void _goToRoom(
+    BuildContext context,
```
sherlockvn, line 271: can you remove context here ?

sherlockvn: use context of new_group is enough

imGok:
updated demo for group name

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -29,6 +30,7 @@ class ChatAppBarTitle extends StatelessWidget {
     required this.sendController,
     required this.getStreamInstance,
     required this.onPushDetails,
+    required this.roomName,
```
sherlockvn, line 33:
```suggestion
   this.roomName,
```

lib/domain/app_state/room/create_new_group_chat_state.dart
```diff
@@ -13,11 +13,15 @@ class CreateNewGroupChatLoading extends Success {
 
 class CreateNewGroupChatSuccess extends Success {
   final String roomId;
+  final String? groupName;
 
-  const CreateNewGroupChatSuccess({required this.roomId});
+  const CreateNewGroupChatSuccess({
+    required this.roomId,
+    required this.groupName,
```
sherlockvn, line 20:
```suggestion
   this.groupName,
```

hoangdat (approved):

drminh2807 (approved):

sherlockvn:
You forget to `flutter pub get`, because I don't see any change in pubspec.lock. Please do it again then we can merge

PR #678, opened 2023-09-26
Only run pages workflow on web changes

tk-nguyen:
This should limit the deploy pages workflow to only run on changes in `web` folder.

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -1,5 +1,7 @@
 on:
   pull_request:
+    paths:
+      - "web/**"
 
 env:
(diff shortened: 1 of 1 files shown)


drminh2807:
@tk-nguyen I'm not agree that
we still need rebuild the web page when we change the flutter code side
include pubspec.yaml, pubspec.lock, *.dart, ...

tk-nguyen:
Okay, but running pages deploy on every PR is quite time consuming.

Maybe not only on changes in `web`?

drminh2807:
no, it is not. you should decrease the build time, the trigger rule is correct for now

tk-nguyen:
Closing this then.

PR #681, opened 2023-09-26
hot-fix: change folder of dio interceptor cache

sherlockvn:
### Problem:
- The hive database is initialized 2 times, first inside `setUpDioHiveCache` method, then later initialized inside FlutterHiveCollectionsDatabase class. So, the Hive will not be initialized the second time, making some boxes not created.
- Boxes are not created is the second initialization will not be closed.
### Resolved: 
- We just need to initialize the Hive database at different folder, and the problem will be resolved. So i changed the path of dio hive cache

### Demo:

https://github.com/linagora/twake-on-matrix/assets/43041967/afa927f9-41d3-4cc1-89dd-9e901c161fe9

### DIFF
--- a/lib/data/network/dio_cache_option.dart
+++ b/lib/data/network/dio_cache_option.dart
@@ -1,8 +1,6 @@
 import 'package:dio_cache_interceptor/dio_cache_interceptor.dart';
 import 'package:dio_cache_interceptor_hive_store/dio_cache_interceptor_hive_store.dart';
-import 'package:flutter/foundation.dart';
 import 'package:matrix/matrix.dart';
-import 'package:path_provider/path_provider.dart';
 
 class DioCacheOption {
   static const String _hiveBoxName = "twake_dio_cache_hive_store";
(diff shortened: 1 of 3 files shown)


lib/data/network/dio_cache_option.dart
```diff
@@ -24,7 +25,7 @@ class DioCacheOption {
     if (kIsWeb) return null;
     final appDir = await getApplicationDocumentsDirectory();
     Logs().d('DioCacheOption::_getAppDirPath() appDirPath ${appDir.path}');
-    return appDir.path;
+    return '${appDir.path}$_dioCacheFolder';
```
drminh2807, line 28: https://pub.dev/documentation/path_provider/latest/path_provider/getApplicationDocumentsDirectory.html
getApplicationDocumentsDirectory can throw exception , please add try catch for that

drminh2807 (changes requested):

lib/data/network/dio_cache_option.dart
sherlockvn: done

drminh2807 (approved):

lib/data/network/dio_cache_option.dart
```diff
@@ -22,9 +25,16 @@ class DioCacheOption {
...
+    try {
+      appDir = await getApplicationDocumentsDirectory();
+    } catch (e) {
+      Logs().e('DioCacheOption::getApplicationDocumentsDirectory() failed.');
+      return null;
+    }
+
     Logs().d('DioCacheOption::_getAppDirPath() appDirPath ${appDir.path}');
-    return appDir.path;
+    return '${appDir.path}$_dioCacheFolder';
```
nqhhdev, line 37:
```suggestion
    return '_dioCacheFolder${appDir.path}$';
```

drminh2807: ? @nqhhdev  are you sure ?

drminh2807 (approved):

nqhhdev (approved):

PR #682, opened 2023-09-26
fix: Search hide loading after no result

drminh2807:
# Before

https://github.com/linagora/twake-on-matrix/assets/12546908/76948018-a895-473a-8074-cc7ad906b346

# After

https://github.com/linagora/twake-on-matrix/assets/12546908/57dc9ac0-1f07-4c5d-a90c-328a07d0d115


sherlockvn (approved):

drminh2807:
> How about the case to go `Direction.b`?

both directions working correctly, updated description

hoangdat (approved):

PR #684, opened 2023-09-26
fix: Highlight file event

drminh2807:
# Demo
Simulator Screenshot - iPhone 15 Pro - 2023-09-26 at 16 12 25

### DIFF
--- a/lib/pages/chat/events/message_content.dart
+++ b/lib/pages/chat/events/message_content.dart
@@ -177,6 +177,7 @@ class MessageContent extends StatelessWidget with PlayVideoActionMixin {
                 MessageDownloadContent(
                   event,
                   onFileTapped: controller.onFileTapped,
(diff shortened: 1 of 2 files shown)


lib/pages/chat/events/message_content.dart
```diff
@@ -502,10 +507,13 @@ class _MessageVideoBuilder extends StatelessWidget {
 
   final DownloadVideoEventCallback handleDownloadVideoEvent;
 
+  final ValueNotifier<Either<Failure, Success>> searchStatus;
```
hoangdat, line 510: it is `stateless`. How about the initialize state for search? What happen if the chat not in `search`

drminh2807: made it optional

hoangdat (approved):

Issue #321, opened 2023-07-18
[web] Send file in Draft chat screen


sherlockvn:
### DES

```
GIVEN user have not have direct chat with user B before
WHEN user A open new empty chat screen, it will redirect user A to empty chat
THEN send file, voice and call is not added yet.
```

### DOD
- [ ] refactor to reuse with composer in Chat screen
- [ ] send File
- [ ] send voice
- [ ] make a call
- [ ] Add information of users in app bar

imGok:
Why not refactoring chat screen so we can use it as a draft screen too ?

sherlockvn:
OKay i will discuss with you in this afternoon

PR #694, opened 2023-09-27
TW-321 : Send files on draft web

imGok:
#321 

https://github.com/linagora/twake-on-matrix/assets/48354990/1ca456c4-d26d-4838-a1a4-def2d999c196

### DIFF
--- a/lib/pages/chat_draft/draft_chat.dart
+++ b/lib/pages/chat_draft/draft_chat.dart
@@ -1,9 +1,11 @@
 import 'dart:async';
 
 import 'package:emoji_picker_flutter/emoji_picker_flutter.dart';
+import 'package:file_picker/file_picker.dart';
 import 'package:fluffychat/di/global/get_it_initializer.dart';
 import 'package:fluffychat/domain/app_state/direct_chat/create_direct_chat_success.dart';
 import 'package:fluffychat/domain/usecase/create_direct_chat_interactor.dart';
+import 'package:fluffychat/domain/usecase/send_file_on_web_interactor.dart';
 import 'package:fluffychat/pages/chat/chat.dart';
 import 'package:fluffychat/pages/chat_draft/draft_chat_view.dart';
 import 'package:fluffychat/presentation/mixins/common_media_picker_mixin.dart';
@@ -13,6 +15,7 @@ import 'package:fluffychat/presentation/model/chat/chat_router_input_argument.da
 import 'package:fluffychat/presentation/model/presentation_contact.dart';
 import 'package:fluffychat/presentation/model/presentation_contact_constant.dart';
 import 'package:fluffychat/utils/network_connection_service.dart';
+import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:fluffychat/widgets/matrix.dart';
 import 'package:flutter/material.dart';
 import 'package:fluttertoast/fluttertoast.dart';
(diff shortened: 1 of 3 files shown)


lib/pages/chat_draft/draft_chat.dart
```diff
@@ -260,6 +259,40 @@ class DraftChatController extends State<DraftChat>
     );
   }
 
+  @override
+  void sendFileOnWebAction(
+    BuildContext context, {
+    Room? room,
+  }) async {
+    if (room == null) {}
```
drminh2807, line 267: remove it

lib/pages/chat_draft/draft_chat.dart
```diff
@@ -260,6 +259,40 @@ class DraftChatController extends State<DraftChat>
     );
   }
 
+  @override
+  void sendFileOnWebAction(
+    BuildContext context, {
+    Room? room,
```
sherlockvn, line 265: remove this also

lib/pages/chat_draft/draft_chat.dart
```diff
@@ -260,6 +259,40 @@ class DraftChatController extends State<DraftChat>
...
+        sendFileOnWebInteractor.execute(
+          room: newRoom,
+          filePickerResult: result,
+        );
+      },
+    );
+  }
+
+  @override
+  Future<void> sendImages(
```
sherlockvn, line 285: why need to move over here, please make it less changes as little as possible to avoid error

lib/presentation/mixins/send_files_mixin.dart
```diff
@@ -64,9 +64,9 @@ mixin SendFilesMixin {
     final result = await FilePicker.platform.pickFiles(
       withData: true,
     );
-    if (result == null && result?.files.isEmpty == true) return;
+    if (result == null || result.files.isEmpty) return;
 
-    sendFileOnWebInteractor.execute(room: room!, filePickerResult: result!);
+    sendFileOnWebInteractor.execute(room: room!, filePickerResult: result);
```
sherlockvn, line 69: Why do you change it, the old version is correct

lib/pages/chat_draft/draft_chat.dart
imGok: To override it has to have the same signature

lib/presentation/mixins/send_files_mixin.dart
imGok: No it wasn't because, the room was created even if we cancel file picker, we can't have result == null and result empty at the same time

lib/pages/chat_draft/draft_chat.dart
imGok: Because we need to create the room before sending the image

sherlockvn: no i means this method look exactly the old code

lib/presentation/mixins/send_files_mixin.dart
sherlockvn: okay you are correct 👍

sherlockvn:
Okay the demo looks great. If you can make the send button into circular progress indicator when create room, it would be great.

hoangdat (approved):

imGok:
> Okay the demo looks great. If you can make the send button into circular progress indicator when create room, it would be great.

About the circular progress, IMO it deserves an other PR / issue because it will add a lot of code out of the scope

sherlockvn (approved):

nqhhdev (approved):

PR #716, opened 2023-10-02
hot-fix/cannot-load-more-in-contacts-tab

sherlockvn:
### Problem:
- The use of duplicated `SingleChildScrollView` make the scroll listener don't know which one it will listen to

### Resolve:
- Remove one of `SingleChildScrollView`

### Demo mobile:

https://github.com/linagora/twake-on-matrix/assets/43041967/3cd90c2a-39a0-40c2-9df9-f07d5a36b8f2

### Demo web:

https://github.com/linagora/twake-on-matrix/assets/43041967/e1cbb514-12d5-464b-b304-917a4dd280b3

### DIFF
--- a/lib/pages/contacts_tab/contacts_tab_body_view.dart
+++ b/lib/pages/contacts_tab/contacts_tab_body_view.dart
@@ -17,76 +17,86 @@ class ContactsTabBodyView extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return SingleChildScrollView(
-      controller: controller.scrollController,
-      child: Padding(
-        padding: const EdgeInsets.all(4.0),
-        child: controller.contactsNotifier == null
-            ? null
-            : ValueListenableBuilder(
-                valueListenable: controller.contactsNotifier!,
-                builder: (context, value, child) => value.fold(
-                  (failure) => const EmptyContactBody(),
-                  (success) {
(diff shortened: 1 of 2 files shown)


lib/pages/contacts_tab/contacts_tab_body_view.dart
```diff
@@ -16,78 +16,75 @@ class ContactsTabBodyView extends StatelessWidget {
...
-                                    contact: contact,
-                                  );
-                                },
-                                child: ExpansionContactListTile(
+                  return Padding(
+                    padding: const EdgeInsets.symmetric(horizontal: 4.0),
+                    child: Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: success.data
+                          .map<Widget>(
```
drminh2807, line 50: please use SliverList for that, also the load more indicator
Screenshot 2023-10-02 at 12 28 55
also move the scrollview to the body view

sherlockvn: Why need to use SilverList and load more indicator already have.

drminh2807: it is implemented wrong
all components should in one scroll view only, so the parent is CustomeScrollView, the child is Slivers

Issue #719, opened 2023-10-02
in group chat: if user send multiple messages, only show name in the first bubble


hoangdat:
### DoD:
- only show displayname at the first bubble message

PR #723, opened 2023-10-02
TW-719 feat: display only first message display name when multi

imGok:
#719 
image

### DIFF
--- a/lib/pages/chat/events/message.dart
+++ b/lib/pages/chat/events/message.dart
@@ -206,7 +206,7 @@ class Message extends StatelessWidget {
                                       crossAxisAlignment:
                                           CrossAxisAlignment.start,
                                       children: [
-                                        ownMessage || event.room.isDirectChat
+                                        hideDisplayName(ownMessage)
                                             ? const SizedBox(height: 0)
                                             : FutureBuilder<User?>(
                                                 future: event.fetchSenderUser(),
@@ -576,6 +576,9 @@ class Message extends StatelessWidget {
     );
   }
 
+  bool hideDisplayName(bool ownMessage) =>
+      ownMessage || event.room.isDirectChat || !isSameSender(nextEvent, event);
+
   Widget _menuActionsRowBuilder(BuildContext context, bool ownMessage) {
     return ValueListenableBuilder(
       valueListenable: isHover,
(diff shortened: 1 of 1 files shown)


lib/pages/chat/events/message.dart
```diff
@@ -206,7 +207,9 @@ class Message extends StatelessWidget {
                                       crossAxisAlignment:
                                           CrossAxisAlignment.start,
                                       children: [
-                                        ownMessage || event.room.isDirectChat
+                                        ownMessage ||
+                                                event.room.isDirectChat ||
+                                                !showDisplayName
```
hoangdat, line 212: Deserve a method

drminh2807: .

imGok: there is already one I forgot to use it mb

lib/pages/chat/events/message.dart
```diff
@@ -579,6 +576,9 @@ class Message extends StatelessWidget {
     );
   }
 
+  bool displayUsername(bool ownMessage) =>
+      ownMessage || event.room.isDirectChat || !isSameSender(nextEvent, event);
```
hoangdat, line 580: no need to display name in `direct chat`

lib/pages/chat/events/message.dart
```diff
@@ -683,26 +683,26 @@ class Message extends StatelessWidget {
...
+    }.contains(comparedEvent.type);
 
     // Ignoring events that are not messages, stickers, encrypted or redaction.
-    if (!isPreviousEventMessage) {
+    if (!isPreviousOrNextEventMessage) {
       return true;
     }
 
-    return previousEvent.senderId != currentEvent.senderId;
+    return currentEvent.senderId != comparedEvent.senderId;
```
hoangdat, line 705: why we need to change it?

lib/pages/chat/events/message.dart
imGok: methods name was not good, it should be "hideDisplayName"

lib/pages/chat/events/message.dart
imGok: I prefer to refactor this one so we can use it with previousEvent and nextEvent, didn't want to do 2 sames methods with exact same code

lib/pages/chat/events/message.dart
```diff
@@ -680,26 +683,26 @@ class Message extends StatelessWidget {
     );
   }
 
-  // Check if the sender of the current event is the same as the previous event.
-  bool isSameSender(Event? previousEvent, Event currentEvent) {
-    // If the previous event is null, it is assumed that the message is the newest.
-    if (previousEvent == null) {
+  // Check if the sender of the current event is the same as the compared event.
+  bool isSameSender(Event? comparedEvent, Event currentEvent) {
```
sherlockvn, line 687: i think `previousEvent` is better name, appropriate with currentEvent name

drminh2807 (approved):

lib/pages/chat/events/message.dart
imGok: it shouldn't be specific to previous or next because we use this method for both previous and next

hoangdat (approved):

Issue #736, opened 2023-10-03
Edit download video in video bubble


sherlockvn:
### DoD: 
- When user click to menu context of video bubble, the last item should be "Download video" instead of "Download image"

![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/2c81711d-44b5-492c-b9b4-5b61b8eab030)

hoangdat:
`Download`

PR #787, opened 2023-10-09
TW-736: rename download file

imGok:
#736 

image

### DIFF
--- a/assets/l10n/intl_zh_Hant.arb
+++ b/assets/l10n/intl_zh_Hant.arb
@@ -527,11 +527,6 @@
         "type": "text",
         "placeholders": {}
     },
-    "downloadFile": "下載文件",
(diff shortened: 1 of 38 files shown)


assets/l10n/intl_zh_Hant.arb
```diff
@@ -527,8 +527,8 @@
         "type": "text",
         "placeholders": {}
     },
-    "downloadFile": "下載文件",
```
hoangdat, line 530: because it is wrong, IMO, delete it, then people will translate it again in Weblate

imGok: true, updated I only kept english, french and vietnamese

hoangdat (approved):

drminh2807 (approved):

PR #788, opened 2023-10-09
TW-727: close selection mode after action

imGok:
#727 

Mobile

https://github.com/linagora/twake-on-matrix/assets/48354990/7cde9b73-0487-44ab-a2e1-1781770c6992

Web

https://github.com/linagora/twake-on-matrix/assets/48354990/e4408cd6-862e-41b4-a7b6-f6a28052a9a6

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1222,7 +1222,7 @@ class ChatController extends State<Chat>
     }
   }
 
-  void pinEvent() {
+  void pinEventAction() {
     final room = this.room;
     if (room == null) return;
     final pinnedEventIds = room.pinnedEventIds;
@@ -1234,7 +1234,6 @@ class ChatController extends State<Chat>
     } else {
       pinnedEventIds.addAll(selectedEventIds);
     }
-    clearSelectedEvents();
     showFutureLoadingDialog(
       context: context,
       future: () => room.setPinnedEvents(pinnedEventIds),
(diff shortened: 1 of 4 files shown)


nqhhdev:
pls add new demo for web, mobile responsive

imGok:
@nqhhdev added web but imo responsive mobile is pointless there since it's a functional change and not ui

drminh2807 (approved):

lib/pages/chat/chat.dart
```diff
@@ -1545,6 +1544,11 @@ class ChatController extends State<Chat>
     }
   }
 
+  void actionWithClearSelections(Function action) {
+    action();
+    clearSelectedEvents();
+  }
```
nqhhdev, line 1550: Hard to understand for me, Why put func inside func?

drminh2807: understand for Khai, I also don't know better way
it will reuse `clearSelectedEvents`, make sure it called every action performed

hoangdat:
Srr, but the ticket is the problem of `Chat List` not `Chat`

imGok:
It's needed for both right ?

hoangdat:
> It's needed for both right ?

yes.

lib/pages/chat/chat.dart
imGok: I don't want to put clearSelectedEvents after every action I find it redundant so I searched a way to refactor that. For me it's not hard to understand but if you have any ideas tell me

nqhhdev (approved):

PR #793, opened 2023-10-09
[WIP] chore: code analyzer unused checks

drminh2807:


lib/config/app_emojis.dart
```diff
@@ -1,30 +0,0 @@
-abstract class AppEmojis {
-  static const List<String> emojis = [
```
hoangdat, line 2: we should keep it for message reaction?

pubspec.yaml
```diff
@@ -38,8 +36,6 @@ dependencies:
     sdk: flutter
   flutter_map: ^3.1.0
   #  flutter_matrix_html: ^1.1.0
-  flutter_olm: ^1.2.0
-  flutter_openssl_crypto: ^0.1.0
```
hoangdat, line 42: keep it

scripts/code_analyze.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh -ve
 flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs
 dart format --set-exit-if-changed lib/ test/
-flutter analyze
+dart run dart_code_metrics:metrics check-unused-files lib
```
tk-nguyen, line 4: Please note, `dart-code-metrics` is discontinued and now requires a license: https://github.com/dart-code-checker/dart-code-metrics.

tk-nguyen: Not that I know of.

drminh2807: dont know another lib
also this check cannot ignore files
so this PR is pending

Issue #725, opened 2023-10-02
Keyboard is close in chat with scroll


hoangdat:
### Desc:
1. Scroll to the middle of Chat
2. Focus to the keyboard
3. Keyboard is closed

drminh2807:
waiting for official fix https://github.com/flutter/flutter/issues/57609

Issue #594, opened 2023-09-10
[Chat] While scrolling, keyboard blinks


nqhhdev:
### Precondition

- Open keybroad
- Scroll message

https://github.com/linagora/twake-on-matrix/assets/99852347/46cd0bb6-cb57-49a0-ab74-e08c63f55b08

QuangHoang1210:
Noted UX later

PR #801, opened 2023-10-10
TW-725: Fix auto hide keyboard when scrolling and send msg

nqhhdev:
### Issue:
- #725 

### Resolved:

```
enum ScrollViewKeyboardDismissBehavior {
  /// `manual` means there is no automatic dismissal of the on-screen keyboard.
  /// It is up to the client to dismiss the keyboard.
  manual,
  /// `onDrag` means that the [ScrollView] will dismiss an on-screen keyboard
  /// when a drag begins.
  onDrag,
}

```

```
  if (keyboardDismissBehavior == ScrollViewKeyboardDismissBehavior.onDrag) {
       return NotificationListener<ScrollUpdateNotification>(
         child: scrollableResult,
         onNotification: (ScrollUpdateNotification notification) {
           final FocusScopeNode focusScope = FocusScope.of(context);
           if (notification.dragDetails != null && focusScope.hasFocus) {
             focusScope.unfocus();
           }
           return false;
         },
       );
     } else {
       return scrollableResult;
     }
```

If on mobile use `ScrollViewKeyboardDismissBehavior.onDrag,` it will be unfocused and cause blinking. Should only support platforms other than mobile.

https://github.com/linagora/twake-on-matrix/assets/99852347/7e826484-d46a-418f-a39d-620c39947327

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -261,7 +261,6 @@ class ChatController extends State<Chat>
     if (!mounted) {
       return;
     }
-    hideKeyboardChatScreen();
     hideSearchKeyboardIfNeeded();
     setReadMarker();
     if (!scrollController.hasClients) return;
--- a/lib/pages/chat/chat_event_list.dart
+++ b/lib/pages/chat/chat_event_list.dart
@@ -1,5 +1,6 @@
 import 'package:fluffychat/pages/chat/group_chat_empty_view.dart';
 import 'package:fluffychat/pages/chat_draft/draft_chat_empty_view.dart';
+import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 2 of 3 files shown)


hoangdat (approved):

drminh2807:
As a fan of iPhone :) I still want scroll to close keyboard
the right behavior is ScrollViewKeyboardDismissBehavior.interactive, but it not supported yet
so I think we still use onDrag but add flag when scolling to bottom after send message

drminh2807:
waiting for official fix https://github.com/flutter/flutter/issues/57609

drminh2807:
Is this refect this issue https://github.com/linagora/twake-on-matrix/issues/594

hoangdat (approved):

drminh2807 (approved):

sherlockvn (approved):

PR #822, opened 2023-10-16
TW-728: fix sharing screenshot is not working in IOS

sherlockvn:
### Issues:
Because the screenshot image need to copy in file on the iOS platform, but the native code is missing the completer for handle async operation, so it have error when sharing in our app, the telegram have the same problem.

### Demo sharing screenshot:

https://github.com/linagora/twake-on-matrix/assets/43041967/00ab87a0-dbcd-489f-bd4c-e49978c43fb2

### Demo sharing to the same room as before

https://github.com/linagora/twake-on-matrix/assets/43041967/dddb06fe-fa7d-4d64-bf55-4d1ffcfb831d

### Demo remove the default share UI when sharing screenshot in IOS

https://github.com/linagora/twake-on-matrix/assets/43041967/79e555dd-6556-49f9-bcfb-74997e56b350


ios/Podfile
```diff
@@ -32,6 +32,10 @@ target 'Runner' do
   use_modular_headers!
 
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
+  
+  target 'Twake Share' do
+    inherit! :search_paths
+  end
```
nqhhdev, line 38: Why we need it? pls explain it to me

sherlockvn: The reason for including inherit! :search_paths is to make sure that the pods you've added through CocoaPods can find the headers and libraries they need during compilation and linking. It's a crucial part of integrating CocoaPods into your project, as it ensures that the pods can be used correctly within the specified target.

This line helps to avoid issues where the pods might not be able to find necessary files or libraries, which can lead to build errors. By inheriting the search paths, you make sure that the pods have access to the appropriate build settings for the specified target.

drminh2807 (approved):

hoangdat (approved):

Issue #824, opened 2023-10-16
[Web] Use `inter` from google instead of from `asset`


hoangdat:

imGok:
I don't think it's a good idea to add a font link to index html for web because it might have conflicts with widget font (assets). 

However I suggest this library https://pub.dev/packages/google_fonts which will lighten our app size. This has HTTP fetching instead

PR #832, opened 2023-10-17
TW-824: install google fonts to fetch font from http

imGok:
implementation of https://pub.dev/packages/google_fonts

#824


web/style.css
```diff
@@ -7,6 +7,7 @@ body {
   height: 100%;
   margin: 0;
   background-color: #1c1b1f;
+
```
drminh2807, line 10: ?

drminh2807:
Screenshot 2023-10-18 at 08 34 29
Screenshot 2023-10-18 at 08 34 39

Better
but it is from linagora.github.io, it can slower on beta.twake.app
I wonder why we still need Inter font in assets
I read it https://pub.dev/packages/google_fonts#bundling-fonts-when-releasing
but I think it good for mobile or desktop, but on web it should download fonts after first loaded
does flutter support it ?

assets/google_fonts/Inter-Black.ttf
hoangdat, line 1: how about to remove all font files?

imGok: with assets, the package will first look into assets folder then if not present it will fetch. 
without assets it will http fetch on first loaded then cache it (no more http font requests after first loading)

without assets
image
with assets 
image

hoangdat:
> @drminh2807 I think it's possible if we use only googlefonts package for web only and do a condition in themes.dart. But is it a good practice?

let's try :)))

drminh2807:
> @drminh2807 I think it's possible if we use only googlefonts package for web only and do a condition in themes.dart. But is it a good practice?

you still keep font in pubspec, so web still download font files from assets

Screenshot 2023-10-20 at 08 14 27

platform specific asset is not supported yet in flutter
https://github.com/flutter/flutter/issues/8230
So you should reverse back to yesterday
we will improve add assets/google-fonts later

sherlockvn (dismissed):

imGok:
@drminh2807 reversed to normal implemention of google fonts, please check again:)

drminh2807 (dismissed):

lib/pages/bootstrap/bootstrap_dialog.dart
```diff
@@ -140,7 +141,7 @@ class BootstrapDialogState extends State<BootstrapDialog> {
                   minLines: 2,
                   maxLines: 4,
                   readOnly: true,
-                  style: const TextStyle(fontFamily: 'RobotoMono'),
+                  style: GoogleFonts.robotoMono(),
```
hoangdat, line 144: no need to use `roboto`, please choose an inter

PR #835, opened 2023-10-18
hot-fix/change back to manual code signing

sherlockvn:


drminh2807:
Is it duplicate with https://github.com/linagora/twake-on-matrix/pull/833 ?

Issue #849, opened 2023-10-23
[Web] Select text in Message bubble by mouse


hoangdat:
### Desc:
- at the moment, user hard to select text to copy in Chat by mouse
- example: 

https://user-images.githubusercontent.com/6462404/277494965-315b3f7d-222f-406b-8db6-3e4e21dac44b.mov

### DoD:
- [ ] can select the whole text or a part of text by mouse

hoangdat:
- [ ] try to add Ctrl+C to copy selected text
- [ ] try to select text in bubble which have formatted (include tag, code block, ...)

nqhhdev:
#892

QuangHoang1210:
Can select by mouse and copy

PR #855, opened 2023-10-24
WIP TW-849: select text in message bubble web

imGok:
Remove select on tap on web and long press

#849 

issues:
- Selectable texts in flutter web is not working wel and interfers with other gestures like dropzone buttons etc.
- sometimes I cannot select text even if I use SelectableText, SelectionArea, SelectableRegion none of them works (for cleanrichtext especially)
- sometimes I activate the drag & drop feature even If I have nothing selected
- selection features such as double tap to select etc doesn't work in Flutter web, it's not like HTML 
- selection works well with simple text such as display name etc (SelectableText)

refs: https://api.flutter.dev/flutter/material/SelectionArea-class.html

### DIFF
--- a/lib/config/go_routes/go_router.dart
+++ b/lib/config/go_routes/go_router.dart
@@ -39,6 +39,7 @@ import 'package:fluffychat/pages/sign_up/signup.dart';
 import 'package:fluffychat/widgets/log_view.dart';
 import 'package:fluffychat/widgets/matrix.dart';
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 import 'package:matrix/matrix.dart';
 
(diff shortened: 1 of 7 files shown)


sherlockvn:
> Remove select on tap on web and long press
> 
> #849
> 
> issues:
> 
> * Selectable texts in flutter web is not working wel and interfers with other gestures like dropzone buttons etc.
> * sometimes I cannot select text even if I use SelectableText, SelectionArea, SelectableRegion none of them works (for cleanrichtext especially)
> * sometimes I activate the drag & drop feature even If I have nothing selected
> * selection features such as double tap to select etc doesn't work in Flutter web, it's not like HTML
> * selection works well with simple text such as display name etc (SelectableText)
> 
> refs: https://api.flutter.dev/flutter/material/SelectionArea-class.html

I have checked the web again and found several issues:
- When user tap to other places like composer, the selected text is blink.
- The copy paste Ctrl+C and Ctrl+V is not working.

PR #910, opened 2023-11-03
Fix color of conversation in chat list

nqhhdev:
### Issue:

- Style for chat list item is wrong for read/unread

### Resolved

Screenshot_20231103_173533


imGok:
hello we implemented google fonts in `themes.dart` is it still necessary with ur new linagora design PR ?

pubspec.yaml
```diff
@@ -108,7 +108,7 @@ dependencies:
   linagora_design_flutter:
     git:
       url: git@github.com:linagora/linagora-design-flutter.git
-      ref: master
+      ref: apply-google-fonts
```
drminh2807, line 111: do we need PR for that ?

nqhhdev: Pls review my PR: 
- https://github.com/linagora/linagora-design-flutter/pull/30

After that I'll back it to ref master

sherlockvn:
Please check other screens also to avoid side effects

nqhhdev:
> Please check other screens also to avoid side effects

It only app for `DecoratorStyle` in chat list

hoangdat (approved):

PR #914, opened 2023-11-05
Refactor all dialog inapp

nqhhdev:
### Issue: 

- In twake app, we use some routers: `FirstColumnInnerRoutes` and `GoRouter`
- We can't manage for `context` of both routers when displaying loading dialog on web, mobile, ...

### Resolved:

- Don't care to which router `context` belongs to because we will probably use multiple routers to build layers UI.
- Just only manage `Tweak Dialog` to display dialog using `context` of `GoRouter` because `context` of `Gouter` is the main router in twake app

https://github.com/linagora/twake-on-matrix/assets/99852347/70e8cca5-4491-4e6e-a830-c3314f19c795


drminh2807 (approved):

lib/utils/dialog/twake_dialog.dart
```diff
@@ -0,0 +1,114 @@
...
+      useRootNavigator: false,
+    );
+  }
+
+  static Future<bool?> showCupertinoDialogFullScreen({
+    required Widget Function() builder,
+  }) {
+    final twakeContext = TwakeApp.routerKey.currentContext;
+    if (twakeContext == null) {
+      Logs().e(
```
sherlockvn, line 79: the `showCupertinoDialog` still called if twakeContext is null

lib/utils/dialog/twake_dialog.dart
```diff
@@ -0,0 +1,114 @@
...
+    return const AlertDialog(
+      content: Row(
+        children: [
+          Padding(
+            padding: EdgeInsets.only(right: 16.0),
+            child: CircularProgressIndicator.adaptive(),
+          ),
+          Expanded(
+            child: Text(
+              'Loading... Please Wait!',
```
sherlockvn, line 106: I think you should use l10n

lib/utils/dialog/twake_dialog.dart
```diff
@@ -0,0 +1,114 @@
...
+      future: future,
+    );
+  }
+
+  static Future<bool?> showDialogFullScreen({
+    required Widget Function() builder,
+  }) {
+    final twakeContext = TwakeApp.routerKey.currentContext;
+    if (twakeContext == null) {
+      Logs().e(
```
sherlockvn, line 62: same

lib/utils/dialog/twake_dialog.dart
nqhhdev: done

lib/utils/dialog/twake_dialog.dart
nqhhdev: done

lib/utils/dialog/twake_dialog.dart
nqhhdev: done

hoangdat (approved):

sherlockvn (approved):

PR #941, opened 2023-11-09
Hotfix cant use context menu in chat list

nqhhdev:
### Issue:

https://github.com/linagora/twake-on-matrix/assets/99852347/1832461e-fddd-45d8-8ac1-06b0c257744f

- Click on any action → nothing happen
- Try to click many times, if it is ok → action can not be close

### Root case:
- In chat list screen use `context` of `Navigator`.
- When context item is opening and click on  context item, We will use the `context` of that screen to trigger closing the context menu. However, in chat list screen that trigger is using the context of `Gouter`. So the two contexts here are wrong.

### Resolved:

- Create popup items specific to different routes

https://github.com/linagora/twake-on-matrix/assets/99852347/38d3516f-ffb4-4464-a8ad-2db213c5cdc2

### DIFF
--- a/lib/widgets/mixins/popup_menu_widget_mixin.dart
+++ b/lib/widgets/mixins/popup_menu_widget_mixin.dart
@@ -5,7 +5,7 @@ import 'package:flutter_svg/flutter_svg.dart';
 import 'package:linagora_design_flutter/colors/linagora_sys_colors.dart';
 
 mixin PopupMenuWidgetMixin {
-  Widget popupItem(
+  Widget popupItemByTwakeAppRouter(
     BuildContext context,
     String nameAction, {
     IconData? iconAction,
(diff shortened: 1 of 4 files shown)


hoangdat (approved):

lib/widgets/mixins/popup_menu_widget_mixin.dart
```diff
@@ -26,36 +26,90 @@ mixin PopupMenuWidgetMixin {
...
+                color: colorIcon ?? Colors.black,
+              ),
+            if (imagePath != null)
+              SvgPicture.asset(
+                imagePath,
+                width: iconSize ?? 20,
+                height: iconSize ?? 20,
+                fit: BoxFit.fill,
+              ),
+            const SizedBox(width: 12),
```
imGok, line 102:
```suggestion
if (iconAction != null || imagePath != null)
    const SizedBox(width: 12),
```

imGok (approved):

sherlockvn (approved):

PR #945, opened 2023-11-10
Improve size for app adaptive scaffold

nqhhdev:
- Updated every styles file for upscaling
- Removed unused styles 

image

image

image

image

#1065

### DIFF
--- a/lib/config/themes.dart
+++ b/lib/config/themes.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter_screenutil/flutter_screenutil.dart';
 import 'package:go_router/go_router.dart';
 import 'package:google_fonts/google_fonts.dart';
 import 'package:linagora_design_flutter/colors/linagora_ref_colors.dart';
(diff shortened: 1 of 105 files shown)


lib/pages/chat_list/client_chooser_button_style.dart
```diff
@@ -1,6 +1,8 @@
+import 'package:flutter_screenutil/flutter_screenutil.dart';
+
 class ClientChooserButtonStyle {
-  static double get avatarSizeInAppBar => 40.0;
-  static double get avatarFontSizeInAppBar => 17.0;
+  static double get avatarSizeInAppBar => 40.0.r;
```
hoangdat, line 4: why it is `r`?

lib/config/themes.dart
```diff
@@ -90,13 +104,13 @@ abstract class TwakeThemes {
             : Colors.blueGrey.shade900,
         popupMenuTheme: PopupMenuThemeData(
           shape: RoundedRectangleBorder(
-            borderRadius: BorderRadius.circular(AppConfig.borderRadius),
+            borderRadius: BorderRadius.circular(AppConfig.borderRadius.r.r),
```
hoangdat, line 107: why need to use double `.r`?

lib/pages/chat_list/client_chooser_button_style.dart
imGok: Fixed

nqhhdev:
Pls rebase main before doing it, because we updated some thing about style text, height, width of appbar

imGok:
@hoangdat size of widget looks good overall except fontsize, I'll check manually why they are not scaling properly

hoangdat:
in 1280 x 800 resolution, we have this problem

image

hoangdat:
- [ ] avatar also be smaller in 1280 x 800

PR #1005, opened 2023-11-17
[Part II] TW-888: Unit test for contacts manager

nqhhdev:


test/domain/contacts/contacts_manager_test.dart
```diff
@@ -0,0 +1,462 @@
...
+
+      contactsManager.initialSynchronizeContacts(
+        isAvailableSupportPhonebookContacts: false,
+      );
+
+      await mockGetTomContactsInteractor
+          .execute(limit: AppConfig.maxFetchContacts)
+          .forEach(
+            (event) => contactsManager.setContactsNotifier(event),
+          );
```
hoangdat, line 125: why we need to care about it?

test/domain/contacts/contacts_manager_test.dart
```diff
@@ -0,0 +1,462 @@
...
+      ).thenAnswer(
+        (_) async* {
+          yield const Right(ContactsLoading());
+          yield const Right(GetContactsSuccess(contacts: []));
+        },
+      );
+
+      contactsManager.initialSynchronizeContacts(
+        isAvailableSupportPhonebookContacts: false,
+      );
```
hoangdat, line 119: that is the function we want to test, we will verify the outputs of it and also the behavior

PR #1008, opened 2023-11-19
TW-888: Add unit test for `ContactsManager`

nqhhdev:


drminh2807:
It good for Unit Test of Contact Manger
but cant test the Interactor
what do you think if we create Intergration test
just mock the datasource, then can test Repo, Interactor, ContactManger
we don't write much test for now, so write 1 test can wrap many cases maybe better

test/domain/contacts/contacts_manager_test.dart
```diff
@@ -0,0 +1,444 @@
...
+    );
+  });
+
+  group('ContactsManager Unit test - ENV: WEB', () {
+    late ContactsManager contactsManager;
+    late MockGetTomContactsInteractor mockGetTomContactsInteractor;
+    late MockPhonebookContactInteractor mockPhonebookContactInteractor;
+
+    test(
+      'WHEN it is not available get Phonebook contact.\n'
```
hoangdat, line 105: AND tomInteractor will return GetContactsSuccess with empty list

lib/domain/contact_manager/contacts_manager.dart
```diff
@@ -12,18 +11,23 @@ import 'package:flutter/foundation.dart';
...
 
   final ValueNotifier<Either<Failure, Success>> _contactsNotifier =
       ValueNotifier(const Right(ContactsInitial()));
 
   final ValueNotifier<Either<Failure, Success>> _phonebookContactsNotifier =
       ValueNotifier(const Right(GetPhonebookContactsInitial()));
 
+  ContactsManager({
+    required this.getTomContactsInteractor,
+    required this.phonebookContactInteractor,
```
hoangdat, line 28: why not keep it as before?

hoangdat (approved):

hoangdat:
please rebase it

nqhhdev:
> please rebase it

Done

Issue #977, opened 2023-11-15
[Quote] Quote indicator inside bubble not expand well


hoangdat:
![Image](https://github.com/linagora/twake-on-matrix/assets/6462404/15f2af0d-a2a1-44e4-afa6-964409fd33e7)

PR #1021, opened 2023-11-20
TW-977: quote indicator expand

imGok:
html list in reply shows only one item now
https://github.com/linagora/flutter_matrix_html/pull/10

#977 

image
image

# hotfix: preview reply html list
old : 
image

new: 
image

### DIFF
--- a/lib/pages/chat/events/reply_content.dart
+++ b/lib/pages/chat/events/reply_content.dart
@@ -68,44 +68,57 @@ class ReplyContent extends StatelessWidget {
       padding: ReplyContentStyle.replyParentContainerPadding,
       decoration:
           ReplyContentStyle.replyParentContainerDecoration(context, ownMessage),
-      child: Row(
-        mainAxisSize: MainAxisSize.min,
-        children: <Widget>[
-          Container(
-            width: ReplyContentStyle.prefixBarWidth,
-            height: ReplyContentStyle.fontSizeDisplayContent * 2,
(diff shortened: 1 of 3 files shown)


lib/pages/chat/reply_display.dart
```diff
@@ -23,33 +23,36 @@ class ReplyDisplay extends StatelessWidget {
...
-                    controller.editEvent?.getDisplayEvent(controller.timeline!),
-                  ),
-          ),
-          IconButton(
-            padding: EdgeInsets.zero,
-            tooltip: L10n.of(context)!.close,
-            icon: Icon(
-              Icons.close,
-              color: Theme.of(context).colorScheme.primary,
+      child: SingleChildScrollView(
```
nqhhdev, line 26: IMO: If you use `SingleChildScrollView` we can scroll now? Should we not allow scrolling here?

imGok: You're right, as discussed this morning we'll only show one line now. I've updated the html package here : https://github.com/linagora/flutter_matrix_html/pull/10 please check it out

nqhhdev (approved):
I tried it on your PR. Look good to me 👍🏻

drminh2807 (approved):

imGok:
Rebased it with new pubspeclock, rdy to merge

hoangdat (approved):

PR #1055, opened 2023-11-23
hot-fix: improve selectable text in message

sherlockvn:
### improve
- Text now can be easier to be select and don't get unselected ofter more
- Selected text can be keep after user have scroll a lot in chat
### Problems: 
- when new event in chat is coming, the selectable text will be disappeared

### DIFF
--- a/lib/pages/chat/chat_event_list.dart
+++ b/lib/pages/chat/chat_event_list.dart
@@ -1,7 +1,9 @@
 import 'package:fluffychat/pages/chat/group_chat_empty_view.dart';
 import 'package:fluffychat/pages/chat_draft/draft_chat_empty_view.dart';
+import 'package:fluffychat/utils/clipboard.dart';
 import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart' hide Clipboard;
 
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:linagora_design_flutter/linagora_design_flutter.dart';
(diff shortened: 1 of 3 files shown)


nqhhdev:
pls add small demo

imGok (approved):
Tested, work way better than before good job

lib/pages/chat/chat_event_list.dart
```diff
@@ -167,3 +174,45 @@ class ChatEventList extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    if (!PlatformInfos.isWeb) {
+      return child;
+    }
+
+    return CallbackShortcuts(
+      bindings: {
+        SingleActivator(
+          LogicalKeyboardKey.keyC,
+          meta: PlatformInfos.isMacKeyboardPlatform,
```
drminh2807, line 202: should create another class to reuse this check, you used it in some cases

sherlockvn: It's different logic, so it's hard to find anything to reused here

sherlockvn: maybe defined const `SingleActivator` to a class

drminh2807: like that
```
SingleActivator.adaptiveControl(LogicalKeyboardKey.keyC)
```

sherlockvn: but some shortcut have ctrl, command, so if change like you, it's should be like before is better

sherlockvn: not only control,cmd also have shift, and future may be shift+cmd, .... option, ...

sherlockvn: then we should keep it like before

Issue #1032, opened 2023-11-21
[Feedback] Can not click on Language settings


hoangdat:

sherlockvn:
https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization

PR #1070, opened 2023-11-26
TW-1032: Language settings

nqhhdev:
### Ticket:
- #1032

#### Mobile 

https://github.com/linagora/twake-on-matrix/assets/99852347/23f18d2f-9023-4406-99d6-69d6eca0284d

#### Web

https://github.com/linagora/twake-on-matrix/assets/99852347/ef3cd29c-f1fb-4099-83a6-3a799325e44a

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -2927,5 +2927,30 @@
   "contactsCount": "Contacts ({count})",
   "linagoraContactsCount": "Linagora contacts ({count})",
   "fetchingPhonebookContacts": "Fetching contacts from device...({progress}% completed)",
-  "activeNow": "Active now"
+  "activeNow": "Active now",
+  "languageEnglish": "English",
+  "@languageEnglish": {
+    "type": "text",
+    "placeholders_order": [],
+    "placeholders": {}
+  },
+  "languageVietnamese": "Vietnamese",
+  "@languageVietnamese": {
+    "type": "text",
+    "placeholders_order": [],
+    "placeholders": {}
+  },
+  "languageFrench": "French",
+  "@languageFrench": {
+    "type": "text",
+    "placeholders_order": [],
+    "placeholders": {}
+  },
+  "languageRussian": "Russian",
+  "@languageRussian": {
+    "type": "text",
+    "placeholders_order": [],
+    "placeholders": {}
+  },
+  "settingsLanguageDescription": "Set the language you use on Twake Chat"
 }
--- a/lib/data/local/localizations/language_cache_manager.dart
+++ b/lib/data/local/localizations/language_cache_manager.dart
@@ -1,0 +1,43 @@
+import 'dart:ui';
+import 'package:fluffychat/utils/famedlysdk_store.dart';
+import 'package:matrix/matrix.dart';
+
+class LanguageCacheManager {
+  static const keyLanguageCode = 'LANGUAGE_CODE';
+  static const keyCountryCode = 'COUNTRY_CODE';
+
+  final Store pres = Store();
+
+  Future<void> persistLanguage(Locale localeCurrent) async {
+    await Future.wait([
+      pres.setItem(keyLanguageCode, localeCurrent.languageCode),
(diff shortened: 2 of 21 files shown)


lib/di/global/get_it_initializer.dart
```diff
@@ -85,6 +91,7 @@ class GetItInitializer {
     NetworkConnectivityDI().bind();
     getIt.registerSingleton(ResponsiveUtils());
     getIt.registerSingleton(TwakeEventDispatcher());
+    getIt.registerSingleton(LanguageCacheManager());
```
hoangdat, line 94: Why we need it as a singleton?

lib/di/global/get_it_initializer.dart
```diff
@@ -129,6 +141,11 @@ class GetItInitializer {
         getIt.get<MediaAPI>(),
       ),
     );
+    getIt.registerLazySingleton(
```
hoangdat, line 144: Idem

lib/di/global/get_it_initializer.dart
```diff
@@ -209,5 +231,10 @@ class GetItInitializer {
       () => ChatGetPinnedEventsInteractor(),
     );
     getIt.registerSingleton<ContactsManager>(ContactsManager());
+    getIt.registerSingleton<SaveLanguageInteractor>(
```
hoangdat, line 234: Idem

sherlockvn:
Does language really need to cache ? Because as the user, They usually use one, or at most 2 languages, and I think they don't switch language often

hoangdat:
And one question, we can synchronize the language to other clients or not? Fx: you have browser, you change language in browser, and your phone app update too.?

nqhhdev:
> And one question, we can synchronize the language to other clients or not? Fx: you have browser, you change language in browser, and your phone app update too.?

1. If the priority exists locally, it will be used, otherwise it will be used by the system.
2. If you saved to local when changing the language on the system, it will not be applied because we are prioritizing local use

assets/l10n/intl_en.arb
```diff
@@ -2927,5 +2927,30 @@
...
   "fetchingPhonebookContacts": "Fetching contacts from device...({progress}% completed)",
-  "activeNow": "Active now"
+  "activeNow": "Active now",
+  "languageEnglish": "English",
+  "@languageEnglish": {
+    "type": "text",
+    "placeholders_order": [],
+    "placeholders": {}
+  },
+  "languageVietnamese": "Vietnamese",
```
drminh2807, line 2937: you should use this lib to get local name from local codes
https://pub.dev/packages/flutter_localized_locales
we also need support all languages defined in `L10n.supportedLocales`

nqhhdev: Don't need to use lib. Something so simple depends on an overly complicated lib

hoangdat: I prefer to use lib in this case. It is simple now, but it can more complicate in the future.

drminh2807 (approved):

lib/data/local/localizations/language_cache_manager.dart
```diff
@@ -0,0 +1,43 @@
+import 'dart:ui';
+import 'package:fluffychat/utils/famedlysdk_store.dart';
+import 'package:matrix/matrix.dart';
+
+class LanguageCacheManager {
+  static const keyLanguageCode = 'LANGUAGE_CODE';
+  static const keyCountryCode = 'COUNTRY_CODE';
+
+  final Store pres = Store();
```
sherlockvn, line 9: Can you clarify what is this, i think you should change name. and should it be singleton ?

nqhhdev: It belongs to matrix sdk, I reuse it all the time

hoangdat: please use getIt.

nqhhdev: done

sherlockvn (approved):

Issue #815, opened 2023-10-12
Do not show any message in history before you join the room


hoangdat:
### DoD:

drminh2807:
![Image](https://github.com/linagora/twake-on-matrix/assets/12546908/7d048841-a1c5-4ae7-b876-fe50d4c5b88a)

we can fix sdk to check RoomJoint instead RoomCreate
but It can make user cannot read very first message from who create room

PR #1089, opened 2023-12-06
TW-815 Do not show any message in history before you join the room

drminh2807:
# Issue
- #815 

# Spec
https://spec.matrix.org/v1.9/client-server-api/#room-history-visibility

# Demo

https://github.com/linagora/twake-on-matrix/assets/12546908/445338d0-7387-45b1-adb8-09816f514cce


lib/domain/usecase/room/create_new_group_chat_interactor.dart
```diff
@@ -31,6 +31,9 @@ class CreateNewGroupChatInteractor {
       );
 
       if (roomId.isNotEmpty) {
+        await matrixClient
+            .getRoomById(roomId)!
+            .setHistoryVisibility(HistoryVisibility.joined);
```
hoangdat, line 36: IMO, we should change it depend on the kind of the room. For the case of 
- `encrypted`: ok with it?
- `unencrypted`:?
- `public`: ?

drminh2807: fixed, set joined for encrypted room only

lib/domain/usecase/room/create_new_group_chat_interactor.dart
```diff
@@ -31,6 +31,11 @@ class CreateNewGroupChatInteractor {
       );
 
       if (roomId.isNotEmpty) {
+        if (createNewGroupChatRequest.enableEncryption == true) {
+          await matrixClient
+              .getRoomById(roomId)!
```
sherlockvn, line 36: why not `?`

drminh2807: done

lib/domain/usecase/room/create_new_group_chat_interactor.dart
```diff
@@ -31,6 +31,11 @@ class CreateNewGroupChatInteractor {
       );
 
       if (roomId.isNotEmpty) {
+        if (createNewGroupChatRequest.enableEncryption == true) {
+          await matrixClient
+              .getRoomById(roomId)
+              ?.setHistoryVisibility(HistoryVisibility.joined);
+        }
```
hoangdat, line 38: IMO, we can add this state to `initialState` of `matrixClient.createGroupChat` method.

sherlockvn (approved):

lib/domain/usecase/room/create_new_group_chat_interactor.dart
drminh2807: done

hoangdat (approved):

PR #1102, opened 2023-12-12
Use `xcbeautify` instead of `xcpretty`

tk-nguyen:
So errors does not get eaten up.


drminh2807:
xcpretty already exist in `macos-latest`
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

tk-nguyen:
> xcpretty already exist in `macos-latest` https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

Yes, this is replacing `xcpretty` because `xcpretty` hides some errors, making it harder to debug.

drminh2807:
but your PR just install it again, it not actually change use xcbeautify, (even though it works). you should find the correct way to enable it

tk-nguyen:
> but your PR just install it again, it not actually change use xcbeautify, (even though it works). you should find the correct way to enable it

Don't see what you mean? It's using it in the script here: https://github.com/linagora/twake-on-matrix/actions/runs/7176075740/job/19540391658#step:9:360

drminh2807 (dismissed):

hoangdat (approved):

PR #1273, opened 2024-01-04
Support and handle recovery data when logged new account

nqhhdev:
### Demo

https://github.com/linagora/twake-on-matrix/assets/99852347/eaf1111f-c144-4354-86c0-d0086b7ad81d


imGok (approved):

lib/pages/chat_list/chat_list.dart
```diff
@@ -741,6 +742,15 @@ class ChatListController extends State<ChatList>
     );
   }
 
+  void _handleAnotherAccountAdded() {
+    if (widget.adaptiveScaffoldBodyArgs is LoggedInBodyArgs) {
```
hoangdat, line 746: should separate `LoggedInBodyArgs` with `LoggedInOtherAccountBodyArgs`

sherlockvn (approved):

PR #1298, opened 2024-01-08
TW-1269, 1270: Fix video player related errors

sherlockvn:
### Demo IOS:

https://github.com/linagora/twake-on-matrix/assets/43041967/c2fcb6e5-96b0-4169-a71d-c96f22608e20

### Test web again:

https://github.com/linagora/twake-on-matrix/assets/43041967/223f9b47-6e76-4f3a-b8e0-f7d65507a47f

### DIFF
--- a/lib/pages/chat/events/download_video_widget.dart
+++ b/lib/pages/chat/events/download_video_widget.dart
@@ -109,7 +109,9 @@ class _DownloadVideoWidgetState extends State<DownloadVideoWidget>
                   switch (downloadState) {
                     case DownloadVideoState.loading:
                       return InkWell(
-                        onTap: _downloadAction,
+                        onTap: downloadState == DownloadVideoState.loading
+                            ? null
+                            : _downloadAction,
                         child: Center(
                           child: Stack(
                             children: [
--- a/lib/pages/chat/events/download_video_widget.dart
+++ b/lib/pages/chat/events/download_video_widget.dart
@@ -86,16 +86,6 @@ class _DownloadVideoWidgetState extends State<DownloadVideoWidget>
       child: Stack(
         alignment: Alignment.topLeft,
         children: [
-          TwakeIconButton(
-            margin: VideoViewerStyle.backButtonMargin(context),
-            tooltip: L10n.of(context)!.back,
-            icon: Icons.close,
-            onTap: () {
-              cancelToken.cancel();
-              Navigator.of(context).pop();
-            },
-            iconColor: Theme.of(context).colorScheme.surface,
-          ),
           Stack(
             alignment: Alignment.center,
             children: [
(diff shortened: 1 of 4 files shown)


hoangdat:
please refer to the ticket in PR

lib/pages/chat/events/download_video_widget.dart
```diff
@@ -109,7 +109,9 @@ class _DownloadVideoWidgetState extends State<DownloadVideoWidget>
                   switch (downloadState) {
                     case DownloadVideoState.loading:
                       return InkWell(
-                        onTap: _downloadAction,
+                        onTap: downloadState == DownloadVideoState.loading
```
hoangdat, line 102: this is inside `case DownloadVideoState.loading:`. Why we need this check?

lib/pages/chat/events/download_video_widget.dart
```diff
@@ -180,6 +170,16 @@ class _DownloadVideoWidgetState extends State<DownloadVideoWidget>
...
+          TwakeIconButton(
+            margin: VideoViewerStyle.backButtonMargin(context),
+            tooltip: L10n.of(context)!.back,
+            icon: Icons.close,
+            onTap: () {
+              cancelToken.cancel();
+              Navigator.of(context).pop();
+            },
+            iconColor: Theme.of(context).colorScheme.surface,
+          ),
```
hoangdat, line 182: what is the different with your change?

lib/widgets/video_viewer_mobile_theme.dart
```diff
@@ -28,7 +27,7 @@ class VideoViewerMobileTheme extends StatelessWidget {
           TwakeIconButton(
             tooltip: L10n.of(context)!.back,
             icon: Icons.close,
-            onTap: () => context.pop(),
+            onTap: () => Navigator.of(context).pop(),
```
nqhhdev, line 30: Why use navigator and not go_router? What's different here?

lib/pages/chat/events/download_video_widget.dart
sherlockvn: The issue occurs because clicking the download button multiple times during video download causes the app to initiate multiple downloads of the same video.

lib/pages/chat/events/download_video_widget.dart
sherlockvn: I make it to the last item in the stack, so that it can be clickable

lib/widgets/video_viewer_mobile_theme.dart
sherlockvn: use `context.pop` will result the screen back to chat list screen, because the dialog is pushed using Navigator, so...

PR #1317, opened 2024-01-10
Tw 1308: twake chat crashed when reject an invitation

sherlockvn:
### Problem:
- 2 consecutive call to `pop()` in Web will throw an exception
- Don't use `Platform.isAndroid` use `PlatformInfo.isAndroid` instead because the first one need to use `dart:io` or `dart:html` to import, but dart.io only support non-web platform, `dart:html` is for web only

### Demo: 

https://github.com/linagora/twake-on-matrix/assets/43041967/bf21ef79-8ed4-4754-988d-ccdd5c7a4763

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -253,7 +253,7 @@ class ChatController extends State<Chat>
     context.go('/rooms/$roomId');
   }
 
-  void leaveChat() async {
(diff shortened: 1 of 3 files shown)


nqhhdev:
Pls add a demo

hoangdat (approved):

nqhhdev (approved):

Issue #1328, opened 2024-01-12
reduce size of bullet in message bubble


hoangdat:
### Desc

![Image](https://github.com/linagora/twake-on-matrix/assets/6462404/5a7986bc-698f-4174-b423-f62ff8b75e14)

### DoD:
- [ ] same as 

![Image](https://github.com/linagora/twake-on-matrix/assets/6462404/02981c8a-e414-4851-930c-9a371e15602b)

![Image](https://github.com/linagora/twake-on-matrix/assets/6462404/0ecdb14b-c945-449f-baf1-02acaa3eee20)

QuangHoang1210:
Passed

<img width="1312" alt="Screenshot 2024-02-16 at 7 32 54 AM" src="https://github.com/linagora/twake-on-matrix/assets/124866146/83d95de8-9396-4537-8acf-28f7bd34ffc5">

PR #1342, opened 2024-01-15
TW-1328: fix bulletpoints and new lines

imGok:
## Related PRs 
https://github.com/linagora/flutter_matrix_html/pull/13
https://github.com/linagora/matrix-dart-sdk/pull/44

#1328

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -926,7 +926,7 @@ packages:
     description:
       path: "."
       ref: master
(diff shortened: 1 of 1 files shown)


hoangdat:
- with your solution, we will have the problem with copy text, the format is not well

docs/adr/0015-parsing-html.md
hoangdat, line 1: remove this for now

Issue #1358, opened 2024-01-17
[Web] Select mode is turn on when copy text of message


hoangdat:

QuangHoang1210:
Cannot copy by click copy text icon after select a message

https://github.com/linagora/twake-on-matrix/assets/124866146/ae441f1c-7196-4c42-a6f2-94f58adad297

QuangHoang1210:
close as following related issue on https://github.com/linagora/twake-on-matrix/issues/1467

PR #1361, opened 2024-01-17
TW-1358: remove unnecessary select actions

imGok:
#1358 

## WEB

https://github.com/linagora/twake-on-matrix/assets/48354990/ad0de827-7a0d-41a2-a6c9-3089bd91b2e8

## MOBILE

https://github.com/linagora/twake-on-matrix/assets/48354990/7b0532be-b8e4-4c74-a4ed-3ab83435e1d7

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -774,10 +774,10 @@ class ChatController extends State<Chat>
         .any((cl) => selectedEvents.first.senderId == cl!.userID);
   }
 
-  void forwardEventsAction() async {
-    if (selectedEvents.length == 1) {
+  void forwardEventsAction({Event? event}) async {
+    if (selectedEvents.isEmpty && event != null) {
       Matrix.of(context).shareContent =
-          selectedEvents.first.getDisplayEvent(timeline!).content;
+          event.getDisplayEvent(timeline!).content;
       Logs().d(
         "forwardEventsAction():: shareContent: ${Matrix.of(context).shareContent}",
       );
@@ -1358,8 +1358,7 @@ class ChatController extends State<Chat>
         replyAction(replyTo: event);
         break;
       case ChatHorizontalActionMenu.forward:
-        onSelectMessage(event);
-        forwardEventsAction();
+        forwardEventsAction(event: event);
         break;
       case ChatHorizontalActionMenu.more:
         handleContextMenuAction(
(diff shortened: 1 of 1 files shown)


lib/pages/chat/chat_view.dart
```diff
@@ -29,8 +29,9 @@ class ChatView extends StatelessWidget with MessageContentMixin {
           TwakeIconButton(
             icon: Icons.copy_outlined,
             tooltip: L10n.of(context)!.copy,
-            onTap: () => controller
-                .actionWithClearSelections(controller.copySingleEventAction),
+            onTap: () => controller.actionWithClearSelections(
+              () => controller.copySingleEventAction(),
+            ),
```
hoangdat, line 34: what is the different?

lib/pages/chat/chat.dart
```diff
@@ -645,10 +645,12 @@ class ChatController extends State<Chat>
...
 
-  void copySingleEventAction() async {
-    if (selectedEvents.length == 1) {
-      await selectedEvents.first.copy(context, timeline!);
+  void copySingleEventAction({Event? event}) async {
+    if (event != null) {
+      return event.copy(context, timeline!);
     }
+
+    await selectedEvents.first.copy(context, timeline!);
```
sherlockvn, line 653: you remove the safety of the code, please check the length of selectecEvents

imGok: you're right thanks

lib/pages/chat/chat_view.dart
imGok: Forgot to remove it after debugging my bad

lib/pages/chat/chat.dart
```diff
@@ -645,10 +645,19 @@ class ChatController extends State<Chat>
...
+    }
+
+    if (selectedEvents.isNotEmpty) {
       await selectedEvents.first.copy(context, timeline!);
+      return;
     }
+
+    Logs().e(
+      'copySingleEventAction():: event is null and selectedEvents is empty',
+    );
```
sherlockvn, line 660: How about you keed this method unchanged

lib/pages/chat/chat.dart
```diff
@@ -1413,15 +1421,13 @@ class ChatController extends State<Chat>
         onSelectMessage(event);
         break;
       case ChatContextMenuActions.copyMessage:
-        onSelectMessage(event);
-        copySingleEventAction();
+        copySingleEventAction(event: event);
```
sherlockvn, line 1424: in here, you can call directly to `event.copy(...)`

lib/pages/chat/chat.dart
imGok: At the moment we can't copy something without selecting it because we forced the selection before copying. That's why I added a way to copy one event individually

lib/pages/chat/chat.dart
imGok: check again pls

hoangdat (approved):

sherlockvn (approved):

PR #1433, opened 2024-01-31
TW-382: remove auto fill chat group name

sherlockvn:
### demo android:
https://github.com/linagora/twake-on-matrix/assets/43041967/c858c141-c10e-4c8b-ba48-00d0ea29e46e

### DIFF
--- a/lib/pages/new_group/new_group_chat_info.dart
+++ b/lib/pages/new_group/new_group_chat_info.dart
@@ -294,17 +294,8 @@ class NewGroupChatInfoController extends State<NewGroupChatInfo>
   }
 
   void _getDefaultGroupName(Set<PresentationContact> contactList) async {
-    if (contactList.length <= 3) {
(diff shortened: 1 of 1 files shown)


lib/pages/new_group/new_group_chat_info.dart
```diff
@@ -294,17 +294,8 @@ class NewGroupChatInfoController extends State<NewGroupChatInfo>
   }
 
   void _getDefaultGroupName(Set<PresentationContact> contactList) async {
```
imGok, line 296: it makes more sense 

```suggestion
  void _initializeGroupName() {
```

sherlockvn: because i don't need it anymore, so removed it.

nqhhdev (approved):

imGok (approved):

Issue #1240, opened 2023-12-28
Exception throw when you close a chat in mobile


hoangdat:
### Desc:
Reproduce:
1. open a chat 
2. chating 
3. close a chat, you will see this exception

![Image](https://github.com/linagora/twake-on-matrix/assets/6462404/e9a091ff-5adf-4574-92ee-be07b3943753)

### DoD:
- [ ] no exception

PR #1464, opened 2024-02-15
TW-1240: Use try catch for `ValueNotifier`

nqhhdev:
### Ticket 

- #1240

### Resolved
- When I disposed of a `ValueNotifier` but still listen to value from it. So need to check if a `ValueNotifier` has been disposed, you can use a try-catch block to catch the `FlutterError` that is thrown when you try to access a disposed `ValueNotifier`.

### DIFF
--- a/lib/pages/chat/input_bar/focus_suggestion_controller.dart
+++ b/lib/pages/chat/input_bar/focus_suggestion_controller.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:matrix/matrix.dart';
 
 class FocusSuggestionController {
   List<Map<String, String?>> _suggestions = List.empty();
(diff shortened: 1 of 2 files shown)


lib/pages/chat/input_bar/focus_suggestion_controller.dart
```diff
@@ -15,16 +16,28 @@ class FocusSuggestionController {
...
-      currentIndex.value = _suggestions.length - 1;
+    try {
+      currentIndex.value--;
+      if (currentIndex.value < 0) {
+        currentIndex.value = _suggestions.length - 1;
+      }
+    } on FlutterError catch (error) {
+      Logs().e(
+        "FocusSuggestionController()::up(): FlutterError: $error",
+      );
```
sherlockvn, line 27: Why we need add try/catch here

lib/pages/chat/input_bar/focus_suggestion_controller.dart
```diff
@@ -15,16 +16,28 @@ class FocusSuggestionController {
...
-      currentIndex.value = 0;
+    try {
+      currentIndex.value++;
+      if (currentIndex.value >= _suggestions.length) {
+        currentIndex.value = 0;
+      }
+    } on FlutterError catch (error) {
+      Logs().e(
+        "FocusSuggestionController()::down(): FlutterError: $error",
+      );
```
sherlockvn, line 40: why need try/catch

hoangdat (approved):

Issue #1454, opened 2024-02-07
[Android / IOS] The list of search chat message result is becoming empty when user trying to scroll down to load more results


QuangHoang1210:
OS: Android / IOS
Device: Mobile
Twake Chat version:  2.4.17 -- beta.twake.app

### *Reproduce Step:
1, Select Chat tab
2, User click on search bar
3, User input keywords "messsage"
4, User click search icon
5, User try to scroll down to load more message

### **Actual:** 

- Becoming empty -> No result

### **Expected:** 

- Still keep the same list of result , just loading more result if having

https://github.com/linagora/twake-on-matrix/assets/124866146/23ff87de-9dce-4dbe-a28f-d38c55adf45a

hoangdat:
can not reproduce

PR #1470, opened 2024-02-16
TW-1454: fix list search messages empty when load more

sherlockvn:
### Demo mobile:

Issue: #1454

https://github.com/linagora/twake-on-matrix/assets/43041967/54850388-e3b0-4d40-9fb0-d8327f944e5d

### DIFF
--- a/lib/domain/app_state/search/server_search_state.dart
+++ b/lib/domain/app_state/search/server_search_state.dart
@@ -35,3 +35,10 @@ class ServerSearchChatFailed extends Failure {
   @override
   List<Object?> get props => [exception];
 }
+
+class ServerSearchLoading extends Success {
+  const ServerSearchLoading();
+
+  @override
+  List<Object?> get props => [];
+}
--- a/lib/pages/search/server_search_controller.dart
+++ b/lib/pages/search/server_search_controller.dart
@@ -68,25 +68,34 @@ class ServerSearchController with SearchDebouncerMixin {
     _searchCategories = null;
   }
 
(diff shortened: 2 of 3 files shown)


lib/domain/app_state/search/server_search_state.dart
```diff
@@ -35,3 +35,10 @@ class ServerSearchChatFailed extends Failure {
   @override
   List<Object?> get props => [exception];
 }
+
+class ServerSearchLoading extends Success {
```
nqhhdev, line 39: IMO: I don't see you using it in `Interactor`, where to use it?

lib/pages/search/server_search_controller.dart
```diff
@@ -74,7 +74,14 @@ class ServerSearchController with SearchDebouncerMixin {
       (success) {
         if (success is ServerSearchChatSuccess) {
           updateNextBatch(success.nextBatch);
-          if (success.results?.isEmpty == true) {
+          List<Result>? searchResults;
+          if (searchResultsNotifier.value is PresentationServerSideSearch) {
```
nqhhdev, line 78: IMO: Why do I need this condition?

sherlockvn: for cast safety

lib/domain/app_state/search/server_search_state.dart
sherlockvn: okey, change in interactor from ServerSearchInitial => ServerSearchLoading

Issue #1400, opened 2024-01-26
[Bug] Cannot pin messages if I'm not the admin of the chat


tprudentova:
1. Open a chat where you're NOT an admin 
2. Select a message and try to pin it 
3. No permission 

Expectation: anyone in the chat can pin messages 

https://github.com/linagora/twake-on-matrix/assets/97684890/556a0632-59a9-4003-8b71-1f56ecf174b2

shepilov:
Everyone can pin messages
Everyone can can change avatar
Everyone can rename the group

QuangHoang1210:
Alll platform - v 2.4.22 @sherlockvn 

Everyone can pin messages -> FAILED 
Everyone can can change avatar -> FAILED 
Everyone can rename the group -> FAILED
Everyone can invite people -> PASSED

QuangHoang1210:
Passed with new created group

PR #1512, opened 2024-02-27
TW-1400: allow user to change avatar, group name, pin message, invite people in chat

sherlockvn:
### Ticket:
#1400 

### Blocker:
https://github.com/linagora/matrix-dart-sdk/pull/49
### Problem:
Before that, the power level of pin message default to be 100 (means that only admin can pin the message), but when create a room, we have an option to override the default power level of the members in the room. So that, we only need to set the power level of pin/unpin event to equals the power level of user in the room

### Demo:

https://github.com/linagora/twake-on-matrix/assets/43041967/0a19c5c1-d86b-40c7-9bb1-3034ff8f9b69

### DIFF
--- a/lib/pages/new_group/new_group_chat_info.dart
+++ b/lib/pages/new_group/new_group_chat_info.dart
@@ -116,6 +116,7 @@ class NewGroupChatInfoController extends State<NewGroupChatInfo>
       EventTypes.RoomName: getPowerLevelUserInChat(),
       EventTypes.RoomAvatar: getPowerLevelUserInChat(),
       EventTypes.RoomMember: getPowerLevelUserInChat(),
+      EventTypes.RoomTopic: getPowerLevelUserInChat(),
     };
   }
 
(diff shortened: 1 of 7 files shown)


Te-Z (approved):

lib/config/power_level_member_in_chat.dart
```diff
@@ -0,0 +1,5 @@
+class DefaultPowerLevelMemberInChat {
+  static const int user = 0;
+  static const int moderator = 50;
+  static const int admin = 100;
+}
```
hoangdat, line 5: enum DefaultPowerLevelMember

lib/pages/new_group/new_group_chat_info.dart
```diff
@@ -102,10 +103,27 @@ class NewGroupChatInfoController extends State<NewGroupChatInfo>
...
 
+  Map<String, dynamic> _getOverridePowerLevelEventForMember() {
+    return {
+      EventTypes.RoomPinnedEvents: getPowerLevelUserInChat(),
+      EventTypes.RoomName: getPowerLevelUserInChat(),
+      EventTypes.RoomAvatar: getPowerLevelUserInChat(),
+      EventTypes.RoomMember: getPowerLevelUserInChat(),
+      EventTypes.RoomTopic: getPowerLevelUserInChat(),
+    };
+  }
```
hoangdat, line 121: create a class `PowerLevelManager`

hoangdat (approved):

PR #1516, opened 2024-02-28
Add Android release APK build and workflow

tk-nguyen:
Release file name should be `twake-<version>-android.apk`

### DIFF
--- a/scripts/release-android-apk.sh
+++ b/scripts/release-android-apk.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env sh
+
+set -eux
+
+flutter build apk --release
+version=$(git describe --tags --exact-match)
+cp build/app/outputs/apk/release/app-release.apk twake-"$version"-android.apk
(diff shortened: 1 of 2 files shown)


sherlockvn (approved):

hieutbui (approved):

scripts/release-android-apk.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+set -eux
+
```
hoangdat, line 4: `flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs`

please add it before build.

tk-nguyen: This step is after [releasing to store](https://github.com/linagora/twake-on-matrix/blob/main/.github/workflows/release.yaml#L78-L80), [which already has the dependencies step](https://github.com/linagora/twake-on-matrix/blob/main/scripts/release-playstore-beta.sh#L2). Using it again here will make the job take more time.

hoangdat (approved):

Issue #1148, opened 2023-12-15
[Bug] Clicking X while uploading an image selects the message


tprudentova:
1. Go to chat, select Attach image 
2. Choose an image, click Send 
3. Click X while the image is still uploading 
4. The message is selected, upload isn't canceled
Expectation -> Clicking X cancels the upload

https://github.com/linagora/twake-on-matrix/assets/97684890/2e7637bb-d46a-4e35-b212-c5ecf9f99f2c

imGok:
cancel image uploading feature is not implemented yet

sherlockvn:
### Dod:
- [ ] Cancel uploading file if have any
- [ ] Remove FileSending event from local database

PR #1546, opened 2024-03-06
[Pending] TW-1148:clicking x while uploading an image selects the message

Te-Z:
Issue: https://github.com/linagora/twake-on-matrix/issues/1148

Resolution:
- add cancel token to cancel the upload
- remove event after cancellation 

[Capture vidéo du 06-03-2024 18:21:05.webm](https://github.com/linagora/twake-on-matrix/assets/31937920/82ab1bf0-a46a-4d31-812e-ff3695a5fcbc)

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -3008,5 +3008,6 @@
   "unselect": "Unselect",
   "searchContacts": "Search contacts",
   "tapToAllowAccessToYourMicrophone": "You can enable microphone access in the Settings app to make voice in",
-  "showInChat": "Show in chat"
+  "showInChat": "Show in chat",
+  "cannotCancelImageUpload": "Cannot cancel image update"
 }
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -3009,5 +3009,5 @@
   "searchContacts": "Search contacts",
   "tapToAllowAccessToYourMicrophone": "You can enable microphone access in the Settings app to make voice in",
   "showInChat": "Show in chat",
-  "cannotCancelImageUpload": "Cannot cancel image update"
+  "cannotCancelImageUpload": "Upload Cancellation Unsuccessful"
 }
--- a/lib/domain/usecase/send_images_interactor.dart
+++ b/lib/domain/usecase/send_images_interactor.dart
@@ -8,7 +8,7 @@ class SendMediaInteractor {
     required Room room,
     required List<FileAssetEntity> entities,
     String? caption,
-    CancelToken? cancelToken,
+    List<CancelToken> cancelTokens = const [],
   }) async {
     try {
       final txIdMapToFileInfo = await room.sendPlaceholdersForImagePickerFiles(
@@ -27,18 +27,20 @@ class SendMediaInteractor {
       }
 
       for (final txId in txIdMapToFileInfo.keys) {
+        final fileIndex = txIdMapToFileInfo.keys.toList().indexOf(txId);
+
         final fakeSendingFileInfo = txIdMapToFileInfo[txId];
         if (fakeSendingFileInfo == null) {
           continue;
         }
 
-        await room.sendFileEvent(
+        room.sendFileEvent(
           fakeSendingFileInfo.fileInfo,
           msgType: fakeSendingFileInfo.messageType,
           fakeImageEvent: fakeSendingFileInfo.fakeImageEvent,
           shrinkImageMaxDimension: 1600,
           txid: txId,
-          cancelToken: cancelToken,
+          cancelToken: cancelTokens[fileIndex],
         );
       }
       if (messageID != null && msgEventContent != null) {
--- a/lib/domain/usecase/send_images_interactor.dart
+++ b/lib/domain/usecase/send_images_interactor.dart
@@ -8,7 +8,7 @@ class SendMediaInteractor {
     required Room room,
     required List<FileAssetEntity> entities,
     String? caption,
-    List<CancelToken> cancelTokens = const [],
+    void Function(Map<String, CancelToken>)? cancelMapCallback,
   }) async {
     try {
       final txIdMapToFileInfo = await room.sendPlaceholdersForImagePickerFiles(
@@ -26,13 +26,18 @@ class SendMediaInteractor {
         );
       }
 
-      for (final txId in txIdMapToFileInfo.keys) {
-        final fileIndex = txIdMapToFileInfo.keys.toList().indexOf(txId);
+      final cancelTokenMap = <String, CancelToken>{};
 
+      for (final txId in txIdMapToFileInfo.keys) {
         final fakeSendingFileInfo = txIdMapToFileInfo[txId];
         if (fakeSendingFileInfo == null) {
           continue;
         }
+
+        final cancelTokenMapKey =
+            "${txId}_${fakeSendingFileInfo.fileInfo.fileName}";
+
+        cancelTokenMap[cancelTokenMapKey] = CancelToken();
 
         room.sendFileEvent(
           fakeSendingFileInfo.fileInfo,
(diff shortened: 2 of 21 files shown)


lib/pages/chat/events/sending_image_info_widget.dart
```diff
@@ -119,4 +128,9 @@ class SendingImageInfoWidget extends StatelessWidget {
       ),
     );
   }
+
+  void _cancelSending() {
+    mediaCancelToken?.cancel();
+    event.remove();
```
sherlockvn, line 134: the return type should be Future<void>, what if event.remove() failed, i see it have return type Future<bool>, you should check the condition and handle error as well

lib/pages/chat/chat.dart
```diff
@@ -1279,6 +1280,8 @@ class ChatController extends State<Chat>
     }
   }
 
+  final CancelToken mediaCancelToken = CancelToken();
```
sherlockvn, line 1283: i think it may not work well in case of multiple images get cancel, try send multiple images

hoangdat: we should have the better way for the case of multiple uploading

Te-Z:
@hoangdat @sherlockvn multiple uploading now handled


https://github.com/linagora/twake-on-matrix/assets/31937920/2133e49f-d4e4-4458-a76f-2ada9bff9817

Te-Z:
> * create an ADR for your modification

There you go @hoangdat . Please let me know if I have to change something on it since it's the first time a write one.

pubspec.lock
hoangdat, line 1: a lot of unnecessary change for it. please revert

assets/l10n/intl_en.arb
```diff
@@ -3008,5 +3008,6 @@
   "unselect": "Unselect",
   "searchContacts": "Search contacts",
   "tapToAllowAccessToYourMicrophone": "You can enable microphone access in the Settings app to make voice in",
-  "showInChat": "Show in chat"
+  "showInChat": "Show in chat",
+  "cannotCancelImageUpload": "Cannot cancel image update"
```
hoangdat, line 3012: why `update`? 
```suggestion
  "cannotCancelImageUpload": "Upload Cancellation Unsuccessful"
```

lib/domain/usecase/send_images_interactor.dart
```diff
@@ -25,17 +27,20 @@ class SendMediaInteractor {
...
         }
 
-        await room.sendFileEvent(
+        room.sendFileEvent(
           fakeSendingFileInfo.fileInfo,
           msgType: fakeSendingFileInfo.messageType,
           fakeImageEvent: fakeSendingFileInfo.fakeImageEvent,
           shrinkImageMaxDimension: 1600,
           txid: txId,
+          cancelToken: cancelTokens[fileIndex],
```
hoangdat, line 43: what happen if `cancelTokens` is empty?

lib/pages/chat/chat_event_list.dart
```diff
@@ -145,6 +146,9 @@ class ChatEventList extends StatelessWidget {
                   final nextEvent = index < controller.timeline!.events.length
                       ? controller.timeline!.events[currentEventIndex + 1]
                       : null;
+
+                  _cleanCancelTokenMap(event);
+
```
hoangdat, line 151: I still not get your point? why put it here the cleanup?

lib/pages/chat/chat_event_list.dart
```diff
@@ -203,6 +211,41 @@ class ChatEventList extends StatelessWidget {
...
   }
 
+  void _cleanCancelTokenMap(Event event) {
+    if (event.status == EventStatus.sent &&
+        controller.mediaCancelTokenMapNotifier.value[event.body] != null) {
+      controller.mediaCancelTokenMapNotifier.value.remove(event.body);
+    }
+  }
+
+  Future<void> _cancelSending({
```
hoangdat, line 221: it is a stateless widget, why we should keep logic here?

lib/pages/chat/chat.dart
```diff
@@ -1294,20 +1298,37 @@ class ChatController extends State<Chat>
...
+      onCameraPicked: (_) => _onSendTap(imagePickerController),
       captionController: _captionsController,
       focusSuggestionController: _focusSuggestionController,
       typeAheadKey: _chatMediaPickerTypeAheadKey,
     );
   }
 
+  void _onSendTap(ImagePickerGridController imagePickerController) {
+    mediaCancelTokenMapNotifier.value = {};
+    final tempMap = <String, CancelToken>{};
```
sherlockvn, line 1313: In this case, the old cancelToken can be removed if the user sends photos twice in a row

lib/pages/chat/chat.dart
```diff
@@ -1294,20 +1298,37 @@ class ChatController extends State<Chat>
...
+
+    for (final selection in imagePickerController.selectedAssets) {
+      final fileName = selection.asset.title;
+      if (fileName != null) {
+        tempMap[fileName] = CancelToken();
```
sherlockvn, line 1318: the key should be `{eventId}_{fileName}`, because user may send the same image 2 times

lib/pages/chat/chat.dart
```diff
@@ -174,6 +175,9 @@ class ChatController extends State<Chat>
 
   final ValueNotifier<DateTime?> stickyTimestampNotifier = ValueNotifier(null);
 
+  final ValueNotifier<Map<String, CancelToken>> mediaCancelTokenMapNotifier =
+      ValueNotifier({});
```
sherlockvn, line 179: i worried that if user back to chat list, this map becomes to empty again

assets/l10n/intl_en.arb
Te-Z: I was wrong, I wanted to write `upload`

pubspec.lock
Te-Z: Those changes are automatic and due to semantic versioning of some packages inside pubspec.yaml. I can revert but those changes will happen anyway when deploying.
Maybe we should add `pubspec.lock` in `.gitignore` since it is a generated file created each time there is a `flutter pub get` ?

lib/pages/chat/chat_event_list.dart
Te-Z: It is here because:
- it has a strong relationship with the Message widget, so I don't think we should use a callback that needs to propagate back to the controller. I'm afraid it might reduce the code's readability.
- it needs the event used in the Message widget
- it does not need a `setState()` call, so for me no need to replace it with a `StatefulWidget`

WDYT ?

lib/pages/chat/chat.dart
Te-Z: If the user go back to chat list while some uploadings are not done, should those uploads continue or not @hoangdat ?

lib/pages/chat/chat_event_list.dart
Te-Z: clean up moved to `chat.dart`

lib/pages/chat/chat.dart
Te-Z: fixed

lib/pages/chat/chat.dart
Te-Z: Smart idea. I implemented it

lib/pages/chat/chat.dart
sherlockvn: The uploading is still continue, because it's registered under getIt as a singleton.

sherlockvn: In order to to fix this case, i think you need to have something like a download manager, register it under the getIt, do all the stuff releated to download here

pubspec.yaml
```diff
@@ -232,7 +232,7 @@ msix_config:
 
 dependency_overrides:
   # Until all dependencies are compatible. Missing: file_picker_cross, flutter_matrix_html
-  ffi: ^2.0.0
+  ffi: 2.1.0
```
hoangdat, line 235: why no change for local_notification?

pubspec.lock
hoangdat, line 1: why too much change here?

PR #1627, opened 2024-03-25
hot-fix: cancel the stream subscription if not used anymore

sherlockvn:
### Problem:
In Chat screen, we have a lot of `StreamSubcription` but we usually don't cancel it, when back to the previous screen. So we have to cancel it everytime the widget is disposed to free the resources. If we don't cancel the subscription, other notifier inside the subscription which is already dispose() can be invoked again.

### DIFF
--- a/lib/pages/chat/chat_pinned_events/pinned_messages.dart
+++ b/lib/pages/chat/chat_pinned_events/pinned_messages.dart
@@ -1,4 +1,8 @@
+import 'dart:async';
+
+import 'package:dartz/dartz.dart' hide State;
 import 'package:fluffychat/app_state/failure.dart';
+import 'package:fluffychat/app_state/success.dart';
 import 'package:fluffychat/di/global/get_it_initializer.dart';
 import 'package:fluffychat/domain/app_state/room/update_pinned_events_state.dart';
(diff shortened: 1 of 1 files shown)


lib/pages/chat/chat_pinned_events/pinned_messages.dart
```diff
@@ -425,17 +438,7 @@ class PinnedMessagesController extends State<PinnedMessages>
...
-    } on FlutterError catch (exception) {
-      Logs().e(
-        'PinnedMessages::_updateEventsNotifier():: FlutterError $exception',
-      );
-    } catch (exception) {
-      Logs().e(
-        'PinnedMessages::_updateEventsNotifier():: ErrorCode $exception',
-      );
-    }
+    eventsNotifier.value = events;
```
hoangdat, line 441: why remove `try - catch`?

sherlockvn: because, exception only throw when we use it not correctly, its lifecycle error.

sherlockvn: now we ensure that we dont use the valuenotifier after dispose, so no need to catch error

hoangdat (approved):

Te-Z (approved):

PR #1628, opened 2024-03-25
Tw 1586: Part 3 downloading file web

sherlockvn:
### Problem with `file_saver`:
- In the older version, in don't throw error or let other people handle the case when the user cancel the download. Updated version of file_saver let developer catch the error, or cancel download action. Therefore, we need to update the file_saver version.

### Updated Demo web:

https://github.com/linagora/twake-on-matrix/assets/43041967/44eba3c6-3a59-43e2-a294-4753d4a4c42d

### DIFF
--- a/lib/pages/chat/events/message_download_content.dart
+++ b/lib/pages/chat/events/message_download_content.dart
@@ -15,6 +15,7 @@ import 'package:fluffychat/widgets/file_widget/download_file_tile_widget.dart';
 import 'package:fluffychat/widgets/file_widget/file_tile_widget.dart';
 import 'package:fluffychat/widgets/file_widget/message_file_tile_style.dart';
 import 'package:fluffychat/widgets/mixins/handle_download_and_preview_file_mixin.dart';
+import 'package:fluffychat/widgets/twake_app.dart';
 import 'package:flutter/material.dart';
 import 'package:matrix/matrix.dart';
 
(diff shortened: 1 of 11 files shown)


hoangdat:
Hi @sherlockvn, we should do the same with it. WDYT?

https://github.com/linagora/twake-on-matrix/assets/6462404/5c5360cd-0611-4413-bb79-c92798ca6f09

lib/pages/chat/events/message_download_content.dart
```diff
@@ -92,17 +94,37 @@ class _MessageDownloadContentState extends State<MessageDownloadContent>
...
+        event: widget.event,
+        context: TwakeApp.routerKey.currentContext!,
+      );
+    }
+  }
+
   @override
   void dispose() {
-    streamSubscription?.cancel();
+    if (!PlatformInfos.isWeb) {
```
Te-Z, line 125: why not cancel the subscription on web ?

sherlockvn: we cancel after the file is downloaded, because we want to download the file even if user change to another chat

lib/pages/chat/events/message_download_content.dart
```diff
@@ -164,6 +186,25 @@ class _MessageDownloadContentState extends State<MessageDownloadContent>
               style: const MessageFileTileStyle(),
             ),
           );
+        } else if (PlatformInfos.isWeb) {
```
hoangdat, line 189: hard to understand

lib/pages/chat/events/message_download_content.dart
hoangdat, line 1: separate this widget for mobile and web

hoangdat: try to rename `setupListeningForStreamSubcription` -> `_trySetupDownloadingStreamSubcription`

Te-Z (approved):

lib/pages/chat/events/message_download_content.dart
sherlockvn: done

hoangdat:
please rebase and squash

Issue #1605, opened 2024-03-19
[Bug] Video sent by user don't have thumbnail


sherlockvn:
### DESC:
When send video from image picker, the thumbnail is loss when the video is sent successfully, the video thumbnail from receiver still can be display, but from sender is loss.
### DEMO:

https://github.com/linagora/twake-on-matrix/assets/43041967/c6dd4977-6870-4a19-92a4-27d402261ed5

### DOD:
- [ ] Display the thumbnail of video

PR #1630, opened 2024-03-26
TW-1605: Fix display thumbnail when user sent a video

nqhhdev:
### Ticket

- #1605 

### Need to merge:

- https://github.com/linagora/matrix-dart-sdk/pull/53

### Root cause:

- While sending a video, the thumbnail display condition was wrong. This condition says that only when uploading and encrypting will the thumbnail be displayed.

```dart

bool isThumbnailGenerated() {
    return fileSendingStatus == FileSendingStatus.encrypting ||
        fileSendingStatus == FileSendingStatus.uploading;
  }

```

### Solution:

- Delete the false condition, update the new condition 

```dart
matrixFile.bytes != null && matrixFile.bytes!.isNotEmpty
```

### Resolved

https://github.com/linagora/twake-on-matrix/assets/99852347/244df7ae-0102-48af-94a7-8ff8a56ed9ba

### DIFF
--- a/lib/domain/model/extensions/platform_file/platform_file_extension.dart
+++ b/lib/domain/model/extensions/platform_file/platform_file_extension.dart
@@ -2,13 +2,26 @@ import 'package:file_picker/file_picker.dart';
 import 'package:matrix/matrix.dart';
 
 extension PlatformFileListExtension on PlatformFile {
-  MatrixFile toMatrixFile() {
+  MatrixFile toMatrixFile({
+    required String temporaryDirectoryPath,
+  }) {
     return MatrixFile.fromMimeType(
       bytes: bytes,
       name: name,
-      filePath: '',
+      filePath: path ?? '$temporaryDirectoryPath/$name',
       readStream: readStream,
       sizeInBytes: size,
+    );
+  }
+
+  FileInfo toFileInfo({
+    required String temporaryDirectoryPath,
+  }) {
+    return FileInfo(
+      name,
+      path ?? '$temporaryDirectoryPath/$name',
+      size,
+      readStream: readStream,
     );
   }
 }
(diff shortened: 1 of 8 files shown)


lib/presentation/mixins/send_files_mixin.dart
```diff
@@ -41,6 +41,7 @@ mixin SendFilesMixin {
     final result = await FilePicker.platform.pickFiles(
       withReadStream: true,
       allowMultiple: true,
+      withData: true,
```
hoangdat, line 44: why we need it?

hoangdat (commented):
what is the root cause?

lib/presentation/extensions/send_file_extension.dart
```diff
@@ -455,35 +455,22 @@ extension SendFileExtension on Room {
...
-    }
-    var width = fileInfo.width;
-    var height = fileInfo.height;
-    if (width == null || height == null) {
-      final imageDimension = await runBenchmarked(
-        '_calculateImageDimension',
-        () => _calculateImageBytesDimension(fileInfo.imagePlaceholderBytes),
-      );
-      width = imageDimension.width.toInt();
-      height = imageDimension.height.toInt();
```
hoangdat, line 478: try catch

Te-Z (approved):

lib/domain/model/extensions/platform_file/platform_file_extension.dart
```diff
@@ -1,12 +1,13 @@
...
 import 'package:matrix/matrix.dart';
+import 'package:path_provider/path_provider.dart';
 
 extension PlatformFileListExtension on PlatformFile {
   MatrixFile toMatrixFile() {
     return MatrixFile.fromMimeType(
       bytes: bytes,
       name: name,
-      filePath: '',
+      filePath: path ?? '${getTemporaryDirectory()}/$name',
```
sherlockvn, line 10: await

sherlockvn (approved):

hoangdat (approved):

PR #1668, opened 2024-04-01
Hotfix: Can't get file on web

nqhhdev:
### Root cause

- When get file from `file_picker`, convert to `MatrixFile`

```dart
MatrixFile toMatrixFileOnMobile({
    required String temporaryDirectoryPath,
  }) {
    return MatrixFile.fromMimeType(
      bytes: bytes,
      name: name,
      filePath: path ?? '$temporaryDirectoryPath/$name',
      readStream: readStream,
      sizeInBytes: size,
    );
  }

```

1. Failed because On web this is always `null`. You should access `bytes` property instead.

```dart
  /// The absolute path for a cached copy of this file. It can be used to create a
  /// file instance with a descriptor for the given path.
  /// ```
  /// final File myFile = File(platformFile.path);
  /// ```
  /// On web this is always `null`. You should access `bytes` property instead.
  /// Read more about it [here](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ)
  String? _path;

  String? get path {
    if (kIsWeb) {
      /// https://github.com/miguelpruivo/flutter_file_picker/issues/751
      throw '''
      On web `path` is always `null`,
      You should access `bytes` property instead,
      Read more about it [here](https://github.com/miguelpruivo/flutter_file_picker/wiki/FAQ)
      ''';
    }
    return _path;
  }
```

2. And using `getTemporaryDirectory` failed on the web

### Resolved

https://github.com/linagora/twake-on-matrix/assets/99852347/5de90a43-21c5-4adf-8362-d739abb051e5

https://github.com/linagora/twake-on-matrix/assets/99852347/70ad6f42-5644-4882-a632-6912b1cce6c9

### DIFF
--- a/lib/domain/model/extensions/platform_file/platform_file_extension.dart
+++ b/lib/domain/model/extensions/platform_file/platform_file_extension.dart
@@ -2,13 +2,23 @@ import 'package:file_picker/file_picker.dart';
 import 'package:matrix/matrix.dart';
 
 extension PlatformFileListExtension on PlatformFile {
-  MatrixFile toMatrixFile({
+  MatrixFile toMatrixFileOnMobile({
     required String temporaryDirectoryPath,
   }) {
     return MatrixFile.fromMimeType(
       bytes: bytes,
       name: name,
       filePath: path ?? '$temporaryDirectoryPath/$name',
+      readStream: readStream,
+      sizeInBytes: size,
+    );
+  }
+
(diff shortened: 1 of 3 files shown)


sherlockvn:
should test in mobile draft chat + send files also

sherlockvn (approved):

hoangdat (approved):

PR #1669, opened 2024-04-01
TW-1666: Filename does not resize and time overlapping.

hieutbui:
## Ticket
- #1666 

### DOD:
- [x] the filename text should be resized also
- [x] the time is not overlap with text

## Root cause
* We are using `RichText` for `filename` but not providing a `textScaler` value for it. Therefore, by default it will always be `1.0` and `noScaling`.
* Inconsistencies in padding for timeline and widget for timeline:
   * We pass `MessageTime` to message widget as `endOfBubbleWidget` to create the necessary buffering for the timeline. This `MessageTime` is wrapped inside a `TextSpan` &rarr; It will resize as the system size changes.
   * We use `MessageTime` without `TextSpan` for rendering timeline &rarr; It will not resize as the system size changes.

## Solution
- Use `Text.rich` widget instead cause by default, it uses `MediaQueryData.textScaleFactor`.
- Wrap widget for timeline inside a `TextSpan`.
- Remove fake `MessageTime` inside message body.
- Re-calculate message bubble width.

## Not handled
* Change system's font size to large and very large
* Still big gap:
    * Message contains preview link &rarr; Cause width of preview and this message will has special HTML tag (`<blockquote>`)
    * Message has reply &rarr; Cause width of reply content and this message will has special HTML tag (`<blockquote>`)

## Relevant
- https://github.com/linagora/flutter_matrix_html/pull/17

## Resolved
- Web:

https://github.com/linagora/twake-on-matrix/assets/80142234/952385ad-7d5c-455c-a4e5-1406c30c104c

- Android:

https://github.com/linagora/twake-on-matrix/assets/80142234/02240cdf-9ab7-497a-ba4a-77c1ef5bd6a8

- HTML special tag that cannot calculate width:

https://github.com/linagora/twake-on-matrix/assets/80142234/502e5be4-33e3-40f2-a9a9-a709ce59c8ee

### DIFF
--- a/lib/pages/chat/events/message/message_content_builder.dart
+++ b/lib/pages/chat/events/message/message_content_builder.dart
@@ -1,12 +1,15 @@
 import 'package:fluffychat/pages/chat/events/message/display_name_widget.dart';
 import 'package:fluffychat/pages/chat/events/message/message_style.dart';
 import 'package:fluffychat/pages/chat/events/message_time.dart';
+import 'package:fluffychat/pages/chat/events/message_time_style.dart';
 import 'package:fluffychat/utils/date_time_extension.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/event_extension.dart';
 import 'package:flutter/material.dart';
 import 'package:fluffychat/pages/chat/events/message/reply_content_widget.dart';
 import 'package:fluffychat/pages/chat/events/message_content.dart';
-import 'package:matrix/matrix.dart';
+import 'package:intl/intl.dart' hide TextDirection;
+import 'package:linagora_design_flutter/colors/linagora_ref_colors.dart';
+import 'package:matrix/matrix.dart' hide Visibility;
 import 'package:fluffychat/utils/string_extension.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/matrix_locals.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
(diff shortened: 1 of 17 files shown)


sherlockvn:
### small size text
image
### Normal text
Screenshot 2024-04-03 at 09 38 18

hoangdat:
- [ ] time MUST BE fixed size, but message content should not overlap in time

hoangdat:
- [ ] verify why in some case, time and message have a very big gap 
image

sherlockvn:
image

sherlockvn:
Screenshot 2024-04-08 at 10 58 29

hieutbui:
> image

fixed!

hieutbui:
Pending:
- [ ] Calculate for tag name inside message

lib/pages/chat/events/message/message_content_builder.dart
```diff
@@ -197,23 +247,150 @@ class MessageContentBuilder extends StatelessWidget {
...
+          totalMessageWidth = lastLineWithTimeWidth;
+        }
+      }
+    }
+
+    final Map<double, bool> result = {
+      totalMessageWidth: isNeedAddNewLine,
+    };
+    return result;
+  }
```
hieutbui, line 395: Will separate later

hoangdat (approved):

sherlockvn (approved):

PR #1683, opened 2024-04-04
hot-fix: PopScope make the app throw at runtime in android

sherlockvn:


hoangdat:
- [ ] side effect in chrome

Te-Z (approved):

Issue #1675, opened 2024-04-03
status of user is not correct


sherlockvn:
### Description:
- Status of user is unstable.

### Demo:

https://github.com/linagora/twake-on-matrix/assets/43041967/302b9b79-d454-4bb9-b4c2-d3884960b3ef

### DOD:
- [ ] make the status of file correct

PR #1684, opened 2024-04-04
TW-1675: status of user is not correct

Te-Z:
### Issue: 
https://github.com/linagora/twake-on-matrix/issues/1675

### Root cause:

The root cause was in matrix-dart-sdk. Here is the copy paste of the explanation I made there:
 

> This was the previous code:
> https://github.com/linagora/matrix-dart-sdk/blob/e0b65295255b852532878a948662920d6963cc6f/lib/src/client.dart#L1793-L1799
> 
> The UI previously listened to `onPresenceChanged` which is a `CachedStreamController` . This instance had a new value added in the `for each` loop for every items in `sync.presence`list. This means if there's  different values in `sync.presence` , the UI will be updated 6 times.
> 
> To handle that I changed for this code:
> 
> https://github.com/linagora/matrix-dart-sdk/blob/b0c34117c60360d3f1241af9653ddaaa88aaead4/lib/src/client.dart#L1809-L1829
> 
> Here `lowestLastActivePresence` is updated for each items in `sync.presence` list if it is `null` or if the current item's timestamp is before the one in `lowestLastActivePresence`. Then when the loop is over and we are sure to have the right value, we can update onLatestPresenceChange with the right value and this way update the UI.

### Warning:

Merge https://github.com/linagora/matrix-dart-sdk/pull/54 before this one

### Demo:

[Capture vidéo du 04-04-2024 14:20:22.webm](https://github.com/linagora/twake-on-matrix/assets/31937920/2a85ea3b-e895-43e2-b637-9304a35811c7)

https://github.com/linagora/twake-on-matrix/assets/31937920/2ef61a81-6278-4771-aca5-3b938559ec4f

### DIFF
--- a/docs/adr/0021-listen-to-presence-status.md
+++ b/docs/adr/0021-listen-to-presence-status.md
@@ -29,39 +29,36 @@ This was because the UI was listening to `onPresenceChanged`'s stream which is u
 
 To avoid this problem we created a new stream which purpose is to get the status of presence the closest of current time: `onlatestPresenceChanged` . That's the one who should be listened by the UI.
 
-Here `lowestLastActivePresence` is updated for each items in sync.presence list if it is `null` or if the current item's timestamp is before the one in `lowestLastActivePresence`. Then when the loop is over and we are sure to have the right value, we can update `onLatestPresenceChange` with the right value and this way update the UI.
+Here `lastActivePresence` is updated for each items in `sync.presence` list if it is `null` or if the current item's timestamp is after the one in `lastActivePresence`. Then when the loop is over and we are sure to have the right value, we can update `onLatestPresenceChange` with the right value and this way update the UI.
 
 ```dart
   /// Callback will be called on presence updates.
   final CachedStreamController<CachedPresence> onPresenceChanged =
-      CachedStreamController();
+    CachedStreamController();
 
   /// Callback will be called on presence update and return latest value.
   final CachedStreamController<CachedPresence> onlatestPresenceChanged =
-      CachedStreamController();
+    CachedStreamController();
 
-    CachedPresence? lowestLastActivePresence;
+  CachedPresence? lastActivePresence;
 
+  for (final newPresence in sync.presence ?? []) {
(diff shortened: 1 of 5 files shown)


hoangdat:
- [ ] write the ADR for this problem

docs/adr/0021-listen-to-presence-status.md
```diff
@@ -0,0 +1,67 @@
...
+      onPresence.add(newPresence);
+      onPresenceChanged.add(cachedPresence);
+  }
+```
+
+## Decision
+
+To avoid this problem we created a new stream which purpose is to get the status of presence the closest of current time: `onlatestPresenceChanged` . That's the one who should be listened by the UI.
+
+Here `lowestLastActivePresence` is updated for each items in sync.presence list if it is `null` or if the current item's timestamp is before the one in `lowestLastActivePresence`. Then when the loop is over and we are sure to have the right value, we can update `onLatestPresenceChange` with the right value and this way update the UI.
```
hoangdat, line 32: before or after?

lib/pages/chat/chat_view.dart
```diff
@@ -170,7 +170,8 @@ class ChatView extends StatelessWidget with MessageContentMixin {
                             sendController: controller.sendController,
                             connectivityResultStream: controller
                                 .networkConnectionService
-                                .getStreamInstance(),
+                                .connectivity
+                                .onConnectivityChanged,
```
hoangdat, line 174: what is the different? IMO, it works well now.

docs/adr/0021-listen-to-presence-status.md
Te-Z: after. ADR updated and unit test added on https://github.com/linagora/matrix-dart-sdk/pull/54

lib/pages/chat/chat_view.dart
Te-Z: no real difference, just more straightforward since the package gives us directly this stream

Issue #1650, opened 2024-03-29
[Out of Sprint] Memory footprint


nqhhdev:
### Decs

![Image](https://github.com/linagora/twake-on-matrix/assets/99852347/4dabc91d-8fb6-4b39-a155-8c349cdc3d6f)

### DoD:
- use Flutter Dev tool to inspect memory leak of the app (profile)
- handle found leak in separated PR
- [x] disable log when app run in production (release)

hoangdat:
### handle sync
- `onUpdate` stream in each room still visible whenever user open other chat?
- how to manange `onUpdate` better?

Te-Z:
Still happening

![Image](https://github.com/linagora/twake-on-matrix/assets/31937920/b9fe2f2c-5036-4cc5-b964-3b3cc1482217)

Te-Z:
This ticket will take a long time so first let's find a direction to apply on the next sprint

Te-Z:
might be related to https://github.com/linagora/twake-on-matrix/issues/1693

hoangdat:
- [ ] will do memory profiling in every release 
- [ ] release manager should do it and post the result to release ticket

PR #1741, opened 2024-05-06
TW-1650: part II: memory footprint in web: remove unnecessary animation and Image.memory in web

sherlockvn:
### DESC:
#1650 

### Problem:
- Currently, we use `Image.memory` to load image data to view, which required two steps, its need to downloaded to memory first, then pass it to the widget, which make app freezes a lot while scrolling.
-  Moreover, `MxcImage` use combination of animation and setState when image is downloaded, which is not efficient.
-`Image.network` have multiple optimization, and build-in method (`loadingBuilder` and `frameBuilder`) which make the rendering in web more efficient. And We don't have to convert to bytes anymore, use the download link, and the browser will help us caching the image.

- For the case of encrypted chat, we still have to use `Image.memory` because of we have to decrypte bytes data after downloading done

- For the case of `Avatar`, use cacheWidth to reduce the memory usage when decoding image

### Reference:
[Reference of frameBuilder in Image.network](https://github.com/flutter/flutter/issues/102828)

### Compare performance and memory usage between beta.twake.app and current 
https://github.com/linagora/twake-on-matrix/assets/43041967/7469752f-37c0-4a19-a40a-38306fe92440

### Demo:

https://github.com/linagora/twake-on-matrix/assets/43041967/4723ab44-17c9-4c54-9bbf-3ee82d8d1a3a

https://github.com/linagora/twake-on-matrix/assets/43041967/841f8bf9-ff33-4ac9-be5a-c2579e99ac41

https://github.com/linagora/twake-on-matrix/assets/43041967/71f19992-0feb-4e5d-b399-f466a4aa3a41

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -16,6 +16,7 @@ import 'package:fluffychat/widgets/mixins/popup_menu_widget_style.dart';
 import 'package:fluffychat/widgets/mixins/twake_context_menu_mixin.dart';
 import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:fluffychat/utils/extension/global_key_extension.dart';
+import 'package:inview_notifier_list/inview_notifier_list.dart';
 import 'package:universal_html/html.dart' as html;
 
 import 'package:adaptive_dialog/adaptive_dialog.dart';
@@ -1930,7 +1931,7 @@ class ChatController extends State<Chat>
     pinnedMessageScrollController.dispose();
     onUpdateEventStreamSubcription?.cancel();
     keyboardVisibilitySubscription?.cancel();
-
+    InViewNotifierListCustom.of(context)?.dispose();
     replyEventNotifier.dispose();
     super.dispose();
   }
--- a/lib/pages/chat/events/image_builder_web.dart
+++ b/lib/pages/chat/events/image_builder_web.dart
@@ -1,0 +1,118 @@
+import 'package:fluffychat/pages/chat/events/image_bubble.dart';
+import 'package:fluffychat/pages/chat/events/message_content_style.dart';
+import 'package:fluffychat/pages/image_viewer/image_viewer.dart';
+import 'package:fluffychat/presentation/enum/chat/media_viewer_popup_result_enum.dart';
+import 'package:fluffychat/utils/interactive_viewer_gallery.dart';
+import 'package:fluffychat/utils/platform_infos.dart';
+import 'package:fluffychat/widgets/hero_page_route.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_blurhash/flutter_blurhash.dart';
+import 'package:matrix/matrix.dart';
+
+class ImageBuilderWeb extends StatelessWidget {
+  final Event event;
+
+  final bool isThumbnail;
+
+  final double width;
+
+  final double height;
+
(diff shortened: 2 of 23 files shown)


lib/pages/chat/events/image_builder_web.dart
```diff
@@ -0,0 +1,148 @@
...
+            );
+          },
+        ),
+      );
+      if (result == MediaViewerPopupResultEnum.closeRightColumnFlag) {
+        closeRightColumn?.call();
+      }
+    } else if (onTapSelectMode != null) {
+      onTapSelectMode!();
+      return;
```
Te-Z, line 85: since the method has a void type, no need to add return IMO

sherlockvn: done

lib/pages/chat/events/image_builder_web.dart
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      return;
+    }
+  }
+}
+
+class UnencryptedImageWidget extends StatelessWidget {
```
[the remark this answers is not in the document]
sherlockvn: done

lib/pages/chat/events/image_builder_web.dart
```diff
@@ -0,0 +1,148 @@
...
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return Hero(
+      tag: event.eventId,
+      child: Material(
+        child: InkWell(
+          mouseCursor: SystemMouseCursors.click,
+          borderRadius: BorderRadius.circular(12.0),
```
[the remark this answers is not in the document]
sherlockvn: done

lib/pages/chat/events/images_builder/image_placeholder.dart
```diff
@@ -0,0 +1,54 @@
...
+        child: CircularProgressIndicator.adaptive(),
+      );
+    }
+    final String blurHashString =
+        event.blurHash ?? AppConfig.defaultImageBlurHash;
+    final ratio = event.infoMap['w'] is int && event.infoMap['h'] is int
+        ? event.infoMap['w'] / event.infoMap['h']
+        : 1.0;
+    var width = 32;
+    var height = 32;
```
nqhhdev, line 34: Move it to style file

hoangdat (approved):

sherlockvn:
- Hot fix: timestamp is not updated when scroll in chat
-> move `InViewState` dispose() in chat controller, because its like a static instance, so we can't dispose it in the message dispose() method

https://github.com/linagora/twake-on-matrix/assets/43041967/59c61fd0-3c6f-47be-8fe8-809c1a472c65

lib/pages/chat/chat.dart
```diff
@@ -1930,7 +1931,7 @@ class ChatController extends State<Chat>
     pinnedMessageScrollController.dispose();
     onUpdateEventStreamSubcription?.cancel();
     keyboardVisibilitySubscription?.cancel();
-
+    InViewNotifierListCustom.of(context)?.dispose();
```
hoangdat, line 1934: it still dispose `inview` well?

sherlockvn: yes, here is where its reference.
Screenshot 2024-05-13 at 17 53 25

hoangdat (approved):

Issue #1602, opened 2024-03-18
[BUG] When typing and shift + ctrl to get a line break, the internal mouse cursor disappears


nqhhdev:
Reproduce step:

- Typing something
- Use `shift` + `enter` many time to get a line break
- Can't see cursor in composer

https://github.com/linagora/twake-on-matrix/assets/99852347/9d137251-29ee-41ac-bb8f-fc17b2418ef2

sherlockvn:
- [ ] This is happend when user shift + enter many times, so that it gets many lines, then the cursor is deep down at the bottom of composer

PR #1853, opened 2024-06-05
TW-1602: when typing shift ctrl, the cursor is dissappear

sherlockvn:
Issue: https://github.com/linagora/twake-on-matrix/issues/1602

### Demo:
- Safari:

https://github.com/linagora/twake-on-matrix/assets/43041967/7b51f497-59ca-4427-88a5-8a3207163e27

- Chrome:

https://github.com/linagora/twake-on-matrix/assets/43041967/5f2b909b-44fe-4d46-8c59-07f6e160e891

- Firefox:

https://github.com/linagora/twake-on-matrix/assets/43041967/b4889a9b-83ac-4579-bf6b-30d8dff4f2e1

### IOS:

https://github.com/linagora/twake-on-matrix/assets/43041967/07d4df01-e235-4387-9d14-6ea81e781ec7

### DIFF
--- a/lib/pages/chat/input_bar/input_bar_shortcut.dart
+++ b/lib/pages/chat/input_bar/input_bar_shortcut.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:fluffychat/pages/chat/input_bar/focus_suggestion_controller.dart';
 import 'package:fluffychat/presentation/extensions/text_editting_controller_extension.dart';
 import 'package:fluffychat/utils/one_time_debouncer.dart';
@@ -16,13 +18,16 @@ class InputBarShortcuts extends StatelessWidget {
 
   final FocusSuggestionController? focusSuggestionController;
 
+  final ScrollController scrollController;
+
   InputBarShortcuts({
     super.key,
-    required this.child,
+    required this.scrollController,
     this.room,
     this.controller,
     this.onEnter,
     this.focusSuggestionController,
+    required this.child,
   });
 
   final _debouncer = OneTimeDebouncer(milliseconds: 50);
@@ -37,6 +42,10 @@ class InputBarShortcuts extends StatelessWidget {
         ): () {
           _debouncer.run(() {
             controller?.addNewLine();
+            Timer(const Duration(milliseconds: 50), () {
+              scrollController
+                  .jumpTo(scrollController.position.maxScrollExtent);
+            });
           });
         },
         const SingleActivator(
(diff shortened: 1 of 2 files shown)


lib/pages/chat/input_bar/input_bar_shortcut.dart
```diff
@@ -37,6 +42,10 @@ class InputBarShortcuts extends StatelessWidget {
         ): () {
           _debouncer.run(() {
             controller?.addNewLine();
+            Timer(const Duration(milliseconds: 50), () {
+              scrollController
+                  .jumpTo(scrollController.position.maxScrollExtent);
+            });
```
hoangdat, line 48: how about if user typing very quick?

lib/pages/chat/input_bar/input_bar_shortcut.dart
```diff
@@ -16,13 +16,16 @@ class InputBarShortcuts extends StatelessWidget {
...
+
   InputBarShortcuts({
     super.key,
-    required this.child,
+    required this.scrollController,
     this.room,
     this.controller,
     this.onEnter,
     this.focusSuggestionController,
+    required this.child,
```
hieutbui, line 28: IMO, all `required` properties should be right after key then the optional properties

lib/pages/chat/input_bar/input_bar_shortcut.dart
```diff
@@ -37,6 +40,10 @@ class InputBarShortcuts extends StatelessWidget {
         ): () {
           _debouncer.run(() {
             controller?.addNewLine();
+            _debouncer.run(() {
+              scrollController
```
Te-Z, line 44: why two debouncers ?

lib/pages/chat/input_bar/input_bar_shortcut.dart
sherlockvn: every flutter widget, put the child/children as the last element in parameter lists

lib/pages/chat/input_bar/input_bar_shortcut.dart
sherlockvn: sr, i try with debouncer, but not work, so i use Timer. But the general idea is that we need an interval so that the `maxScrollExtent` can be updated correctly when make a new line everytime

lib/pages/chat/input_bar/input_bar_shortcut.dart
sherlockvn: When user typing very fast (in my case, i long press it). it has no delay when scroll to bottom
https://github.com/linagora/twake-on-matrix/assets/43041967/0f58033b-e087-4216-a028-1ff4c1885bae

Te-Z (approved):

nqhhdev:
I tested all browsers:

- [x] Safari
- [x] Firefox
- [x] Arc
- [x] Chrome

Works well for me

nqhhdev (approved):
LGMT

Issue #1899, opened 2024-06-27
[Bug] [Web] When cancel sending file in web, chat still being created


sherlockvn:
### DESC:
- When select file from device in web, if user cancel the sending file in sendingFileDialog, The chat is still created

https://github.com/linagora/twake-on-matrix/assets/43041967/0cfaa6e1-dce8-45b8-a1fe-32f51a9f98c1

PR #1903, opened 2024-07-01
TW-1899: Chat created after cancel sending file inside draft chat on web

hieutbui:
## Ticket
- #1899 

## Resolved

https://github.com/linagora/twake-on-matrix/assets/80142234/9171397e-aae8-43dd-8e19-af65d2ee13ee

### DIFF
--- a/lib/pages/chat_draft/draft_chat.dart
+++ b/lib/pages/chat_draft/draft_chat.dart
@@ -13,6 +13,7 @@ import 'package:fluffychat/pages/chat/chat.dart';
 import 'package:fluffychat/pages/chat/input_bar/focus_suggestion_controller.dart';
 import 'package:fluffychat/pages/chat_draft/draft_chat_view.dart';
 import 'package:fluffychat/presentation/enum/chat/right_column_type_enum.dart';
+import 'package:fluffychat/presentation/enum/chat/send_media_with_caption_status_enum.dart';
 import 'package:fluffychat/presentation/mixins/common_media_picker_mixin.dart';
 import 'package:fluffychat/presentation/mixins/media_picker_mixin.dart';
 import 'package:fluffychat/presentation/mixins/send_files_mixin.dart';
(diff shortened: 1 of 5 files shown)


lib/pages/chat_draft/draft_chat.dart
```diff
@@ -356,26 +346,62 @@ class DraftChatController extends State<DraftChat>
...
-          sendFileOnWebInteractor.execute(
-            room: newRoom,
-            files: matrixFilesList,
+
+    if (dialogStatus is SendMediaWithCaptionStatus) {
+      switch (dialogStatus) {
+        case SendMediaWithCaptionStatus.cancel:
+          break;
+        case SendMediaWithCaptionStatus.done:
+        case SendMediaWithCaptionStatus.error:
```
Te-Z, line 385: in the case of `SendMediaWithCaptionStatus.error`  why would it behave the same as the case of `SendMediaWithCaptionStatus.done` ?

hieutbui: This is because the `error` case stands for `room` is null

sherlockvn: i think in the case of error, no need to create room, instead of log it only

hieutbui: After selected images, the `SendFileDialog` appear. If click on `Send`, it check if the current room is null then pop back with value `error`. In this problem, user is inside a draft chat so the `room` will be null. Therefore, `SendFileDialog` always return `error`. So case `error` also need to create a room or file will never be sent.

sherlockvn: okay, you're right

sherlockvn: so i think the best we should do is change `SendMediaWithCaptionStatus.error` -> `SendMediaWithCaptionStatus.emptyRoom`. Because its not error in the case of draft chat, but will be error in case of send Files in chat

nqhhdev: IMO: Separate func to handler result of dialog.
- Easy to understanding 
- Easy to handle result

hieutbui: Updated!

sherlockvn (approved):

sherlockvn:
Tested and worked well from my side.

Te-Z (approved):
Tested on:
- android, 
- chrome  
- firefox
- gnome web
- opera
- edge
Looks good to me

Te-Z:
tested on iOS

Issue #1900, opened 2024-06-27
[Urgent] iOS is forced to logout many times


Te-Z:

Te-Z:
@nqhhdev @sherlockvn can one of you take this issue urgently please ? Ping Dat if

sherlockvn:
Can't reproduce in twake version 2.5.9

Te-Z:
@hoangdat do you have more informations about this one ?

sherlockvn:
i can reproduce now, rarely happend and i'm trying to put log for debugging it

sherlockvn:
i can only reproduce in the specific group `Chat dev team`

PR #1921, opened 2024-07-04
Tw 1900: IOS is forced to log out many times

sherlockvn:
### Issue:
#1900

### How to reproduce:
- in your iphone, lock the iphone and make sure the Twake Chat is terminated. Wait for a notification from encrypted room to come, when the notification appear in the screen, wait for 5-10 second, then unlock the iphone, click the notification, the user will be redirect to /twakeWelcome page.

- Logs:
image

### Root cause:
- When the device are locked, Twake chat can't access to the keychain storage, then because of the twake chat is terminated, the app need to run from main() again, in main method, it need to create client again. When creating a new client, the old logic need to write data to the Keychain -> this operation will throw an exception at runtime, and the client is null -> user is log out

### Solution:
add a try/catch in the write operation to keychain, so that if its failed, it still return the old client. Make the access to keychain storage to lowest (which is KeychainAccessibility.first_unlock_this_device), this will make sure we don't hvae this error anymore.

### Demo:

https://github.com/linagora/twake-on-matrix/assets/43041967/f761cf93-dae2-4a22-ac93-40ed970067ba

### DIFF
--- a/lib/domain/keychain_sharing/keychain_sharing_manager.dart
+++ b/lib/domain/keychain_sharing/keychain_sharing_manager.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/domain/keychain_sharing/keychain_sharing_restore_token.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:matrix/matrix.dart';
 
 class KeychainSharingManager {
   static FlutterSecureStorage get _secureStorage => const FlutterSecureStorage(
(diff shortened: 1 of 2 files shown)


hoangdat:
Can you reproduce issue 100%?

sherlockvn:
- Reproduceable video
https://github.com/linagora/twake-on-matrix/assets/43041967/358e2168-67fd-42fb-80f2-0008a7c95a64

lib/domain/keychain_sharing/keychain_sharing_manager.dart
```diff
@@ -3,12 +3,15 @@ import 'dart:convert';
...
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:matrix/matrix.dart';
 
 class KeychainSharingManager {
   static FlutterSecureStorage get _secureStorage => const FlutterSecureStorage(
         iOptions: IOSOptions(
           groupId: AppConfig.iOSKeychainSharingId,
           accountName: AppConfig.iOSKeychainSharingAccount,
+          synchronizable: true,
+          accessibility: KeychainAccessibility.first_unlock_this_device,
```
nqhhdev, line 14: Could you write an ADR for this case.

nqhhdev (approved):
I was built it and test on my iphone. LGMT

hieutbui (approved):

Issue #1920, opened 2024-07-04
[Bug] Responsive is broken on firefox, edge, gnome web and opera


Te-Z:
On web (firefox, opera, edge, maybe other browsers also) responsive is broken for chat details. Only works on chrome and chromium.

Steps to reproduce:
- open app on web browser
- go in chat room
- open chat details panel
- reduce window's width as much as possible

Bug: 
- Chat panel and chat details is too small and can't display correctly

Expected:
- Should display the same way as mobile

[Capture vidéo du 03-07-2024 23_08_44 (1).webm](https://github.com/linagora/twake-on-matrix/assets/31937920/9de9313d-0c61-4602-8458-3d7c8ddd0278)

Te-Z:
Tested on 2.5.9
What about previous versions ?

PR #1938, opened 2024-07-09
TW-1920: hot-fix: wrong responsive when size of screen is small

sherlockvn:
## Ticket
**Related issue**
#1920

## Root cause
When the width of the screen is less than `bodyRadioWidth` (which is 472), the `getChatBodyRatio` will be large then 1, which will make the `getMinDesktopWidth` be negative. The negative number in breakpoint will make the responsive wrong when the width is less than 472.
## Solution
The solution is when the width of screen is less than `bodyRadioWidth`, we will make the breakpoint for desktop to be constant like before. So that only when the screen is in tablet or desktop will apply the ratio calculated by new formular.
## Impact description

## Test recommendations
Change the width of the screen in web to see any problem about responsive.
## Pre-merge
No.
## Resolved
- Web:

https://github.com/linagora/twake-on-matrix/assets/43041967/c1a86fdb-55e5-468a-9bbc-315fc4044501

- Android:
- IOS:


KhaledNjim:
Tested on firefox and chrome, works well on my side

Te-Z (approved):
tested on chrome, edge and opera
LGTM

KhaledNjim (approved):

hieutbui (approved):

Issue #1766, opened 2024-05-15
Can not decrypt message in chrome in v2.5.2


hoangdat:
### Desc:
- Web (Chrome and Safari at least), we can not execute all the backup process after sign in 

### DoD:
- Should work as normal

tprudentova:
<img width="1280" alt="Screenshot 2024-06-19 at 12 03 34" src="https://github.com/linagora/twake-on-matrix/assets/97684890/f93d7e72-8ac4-4a17-bf50-25feeef4c7f7">

Safari as well

sherlockvn:
- [ ] Spend 2 points for testing and reproduce it

sherlockvn:
The recoveryWords API in both `matrix.linagora.com` and `matrix.lin-saas.com` have 404 status

![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/30ad31a6-0321-4466-b854-57c87ba24e0d)



![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/647cf626-6981-4785-82f2-5baabbcb94f5)



![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/f234342d-c124-4c85-8576-a2391a6ceeef)

sherlockvn:
i need to have the recoveryKey so that, i can continue debug about the message can't be decrypted. As the error message suggests, i guess that the recoverykey is deleted in database

tjiho:
On firefox there is the same issue. 
When I login, I have a 500 error on https://tom.linagora.com/_twake/recoveryWords with `{"error":"duplicate key value violates unique constraint \"recoverywords_pkey\""}`

AudranBert:
> On firefox there is the same issue. When I login, I have a 500 error on https://tom.linagora.com/_twake/recoveryWords with `{"error":"duplicate key value violates unique constraint \"recoverywords_pkey\""}`

Same issue here on firefox

guimard:
Hi, the real issue is "why Chrome wants to replace recoveryWords which exists ? Does it means that it failed to get it previously or is it a wanted replacement ?

Anyway @rezk2ll is going to accept such replacement

rezk2ll:
> Hi, the real issue is "why Chrome wants to replace recoveryWords which exists ? Does it means that it failed to get it previously or is it a wanted replacement ?
> 
> Anyway @rezk2ll is going to accept such replacement but the logical way should be to accept such replacement with a "PUT" request, not a "POST"

[...]

PR #1946, opened 2024-07-11
TW-1766: fix 500,404 error in POST request when login

sherlockvn:
## Ticket
**Related issue**
#1766 
## Root cause
We create the bootstrap dialog at the wrong time, so the state of bootstrap dialog always be `newSSSSKey`, that means, it will call to POST request to upload new key no matter what even when you have got the recovery key success before in tom server.
## Solution
The solution is direct. We move the creating bootstrap dialog after we have loaded successfully the roomData and accountData from homeserver. This will make sure that the bootstrap state is correct and will not call to upload recoveryKey everytime we log in.
## Impact description
Can decrypt message with new account now.
## Test recommendations
1. Create a new account (make sure you follow every steps)
2. Create an encrypted chat
3. Send some messages in encrypted chat, then wait for the messages is sent.
4. Log out and log in again
5. You will see that the messages in encrypted chat has been decrypted.
## Pre-merge
no.
## Resolved
- Web:

https://github.com/linagora/twake-on-matrix/assets/43041967/9bde5adc-7433-431d-b508-c0776258fa99

- Android:
- IOS:

https://github.com/linagora/twake-on-matrix/assets/43041967/76f8ea95-a6fb-44d6-b0e6-b60e3e51a2e5

### DIFF
--- a/lib/pages/bootstrap/bootstrap_dialog.dart
+++ b/lib/pages/bootstrap/bootstrap_dialog.dart
@@ -15,8 +15,6 @@ import 'package:matrix/encryption/utils/bootstrap.dart';
 import 'package:matrix/matrix.dart';
 import 'package:share_plus/share_plus.dart';
 
-import '../key_verification/key_verification_dialog.dart';
-
 class BootstrapDialog extends StatefulWidget {
   final bool wipe;
   final Client client;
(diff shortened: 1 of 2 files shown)


Te-Z:
Tested on android, chrome, edge, firefox. LGTM

hoangdat:
For this ticket, please note the cases we use to test
- [ ] Test with brand new twake account, sign in to app, test in encrypted group, then logout and login again
- [ ] Test with matrix account which you have a good backup by self recovery key
- ... 

Please add more cases, we should maintain these test case to verify problem of backup.

sherlockvn:
case 1:

https://github.com/user-attachments/assets/c89ebee7-55b5-4a20-97a3-f0da4ea93fa1

sherlockvn:
case 2: 
https://github.com/user-attachments/assets/f6b42da9-f6fb-4c12-a8ce-258e948aa1ad

sherlockvn:
case 3: import backup key from mobile

https://github.com/user-attachments/assets/bff3396b-569a-4abd-a28d-61efabcb98fb

lib/pages/bootstrap/tom_bootstrap_dialog.dart
```diff
@@ -57,19 +56,17 @@ class TomBootstrapDialogState extends State<TomBootstrapDialog>
...
-  void _createBootstrap() async {
+  void _initData() async {
     WidgetsBinding.instance.addPostFrameCallback((_) async {
       await _loadingData();
     });
   }
 
   Future<void> setupAdditionalDioCacheOption(String userId) async {
-    Logs().d('TomBootstrapDialog::setupAdditionalDioCacheOption: $userId');
+    Logs().i('TomBootstrapDialog::setupAdditionalDioCacheOption: $userId');
```
Te-Z, line 69: IMO those logs should get back to `Logs().d`

lib/pages/bootstrap/bootstrap_dialog.dart
```diff
@@ -333,25 +331,26 @@ class BootstrapDialogState extends State<BootstrapDialog> {
...
-                              );
-                              if (req.error != null) return;
-                              await KeyVerificationDialog(request: req.result!)
-                                  .show(context);
-                              Navigator.of(context, rootNavigator: false).pop();
-                            },
-                    ),
+                    // TODO: TW-1766: temporary disable right now, because not supported yet
+                    // const SizedBox(height: 16),
+                    // ElevatedButton.icon(
```
Te-Z, line 336: can you explain why you remove this part ?

Te-Z (approved):
Tested again on android and chrome. Works fine

lib/pages/bootstrap/tom_bootstrap_dialog.dart
sherlockvn: don't worry, its only appear when you debug

lib/pages/bootstrap/tom_bootstrap_dialog.dart
```diff
@@ -364,35 +377,35 @@ class TomBootstrapDialogState extends State<TomBootstrapDialog>
...
-                  () =>
-                      _uploadRecoveryKeyState = UploadRecoveryKeyState.initial,
-                );
-              }
-            },
-            (success) => setState(
+    await _deleteRecoveryWordsInteractor.execute().then((either) {
+      _createBootstrap();
+      either.fold(
+        (failure) {
```
nqhhdev, line 383: Pls test case delete key failed also

nqhhdev: This demo

signal-2024-07-17-003020_002

nqhhdev (approved):
LGTM

Issue #1887, opened 2024-06-18
Can't receive notifications


Te-Z:
People complain from not being able to receive notifications.
We need to investigate this

Te-Z:
How it works right now: 
sender -> sygnal -> firebase -> receiver

Te-Z:
On iOS work on the version on testflight but do not work on the version from the store.

Te-Z:
![image](https://github.com/user-attachments/assets/3d1376aa-0151-4cbb-851d-665cc8205aa6)

Te-Z:
Hypothesis: problems with https://pub.dev/packages/flutter_local_notifications (v. 17.1.2 used in twake chat 2.5.9)

Te-Z:
sometimes no notifications at all on Chromium

[Capture vidéo du 13-07-2024 08:07:32.webm](https://github.com/user-attachments/assets/fe7b91a4-30d4-4fc6-9fbb-c66b54783783)

Te-Z:
Opera

[Capture vidéo du 13-07-2024 08:11:59.webm](https://github.com/user-attachments/assets/3ff4caff-6bae-472b-a933-33b55b8ae8f2)

nqhhdev:
What happens on mobile and web?

IMO: 
- We need to focus on mobile first (P1) → Find root cause and solution 
- Note for each browser on web (P2) → Create other tickets

Te-Z:
@nqhhdev no problem on mobile with v2.5.9 (see table above)

Te-Z:
@Bobpodvalnyi I think we should display something explicit like Whatsapp when the notification permission is not granted. On some browsers the notification permission native dialog is not displayed on click. This would avoid some complaints.  WDYT ?

![notification permission](https://github.com/user-attachments/assets/ffec136a-693d-4110-afa9-c1dcb54fa250)

Te-Z:
@Bobpodvalnyi you can see here how the app behaves on different browsers. On each browser it's really different, that's why I think we need an explicit indication:
https://github.com/linagora/twake-on-matrix/pull/1954#issue-2412940079

Te-Z:
Suggestion:
- how to update counter live on tab (web) -> without enabling notification
-

tprudentova:
https://github.com/user-attachments/assets/410de636-3e1b-4a8e-9f20-6ae2da24bb33

The problem with notifications reappeared on Android this week :(

guimard:
[...]

PR #1954, opened 2024-07-17
TW-1887: improve notification permission request on web

Te-Z:
## Ticket
[**Related issue**](https://github.com/linagora/twake-on-matrix/issues/1887)

## Root cause
On web the notification permission was asked when the app starts. But for some browsers it should be on an action of the user. 
Also sound can't be played on some browsers causing exception before displaying the notification.

## Solution
For sound a try catch has been added to avoid to be stuck.
The notification permission is asked when clicked on chat list item but still also when starting the app for browsers who can handle it.

But the notification display conditions changes from one browser to an other.

## Test recommendations
- [ ] Test on all browsers
- [ ] With the browser on foreground without being on the room where you will receive message
- [ ] With the browser on background

## Pre-merge
no

## Resolved
- Web:
Opera (only sounds, notification display when the browser is closing.. will try to handle this later)

[ici.webm](https://github.com/user-attachments/assets/af14fefa-6850-4e70-acdb-c8fa79b96327)

Chromium

[Capture vidéo du 17-07-2024 08:52:16.webm](https://github.com/user-attachments/assets/71e25707-9d5f-4af5-a96e-34a59811ddbd)

Chrome

[Capture vidéo du 17-07-2024 08:50:16.webm](https://github.com/user-attachments/assets/de9fceb2-2f4a-4a9b-8245-2a9ca7d13c01)

Edge

[Capture vidéo du 17-07-2024 08:39:49.webm](https://github.com/user-attachments/assets/855fdad6-35c9-49b1-ac37-f82469ab0176)

Firefox

[Capture vidéo du 17-07-2024 08:37:15.webm](https://github.com/user-attachments/assets/ec5421eb-a4bd-4b3f-9a4f-a4ba743a27c5)

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -17,6 +17,7 @@ import 'package:fluffychat/utils/room_status_extension.dart';
 import 'package:fluffychat/widgets/context_menu/context_menu_action.dart';
 import 'package:fluffychat/widgets/mixins/popup_menu_widget_style.dart';
 import 'package:fluffychat/widgets/mixins/twake_context_menu_mixin.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:fluffychat/utils/extension/global_key_extension.dart';
 import 'package:inview_notifier_list/inview_notifier_list.dart';
@@ -1971,6 +1972,16 @@ class ChatController extends State<Chat>
     }
   }
 
+  void requestNotificationPermission() {
+    if (kIsWeb) {
+      try {
+        html.Notification.requestPermission();
+      } catch (e) {
+        Logs().e("Error requesting notification permission: $e");
+      }
+    }
+  }
+
   @override
   void dispose() {
     unregisterPasteShortcutListeners();
@@ -2015,7 +2026,10 @@ class ChatController extends State<Chat>
   Widget build(BuildContext context) {
     return MouseRegion(
       onHover: (_) => _resetLocationPath(),
-      child: ChatView(this, key: widget.key),
+      child: GestureDetector(
+        onTap: requestNotificationPermission,
+        child: ChatView(this, key: widget.key),
+      ),
     );
   }
 }
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -17,7 +17,6 @@ import 'package:fluffychat/utils/room_status_extension.dart';
 import 'package:fluffychat/widgets/context_menu/context_menu_action.dart';
 import 'package:fluffychat/widgets/mixins/popup_menu_widget_style.dart';
 import 'package:fluffychat/widgets/mixins/twake_context_menu_mixin.dart';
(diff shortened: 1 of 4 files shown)


nqhhdev:
- Pls check case: When open link `/rooms/roomId` => User can click on chat screen then request permission as well.

lib/pages/chat/chat.dart
```diff
@@ -2015,7 +2026,10 @@ class ChatController extends State<Chat>
   Widget build(BuildContext context) {
     return MouseRegion(
       onHover: (_) => _resetLocationPath(),
-      child: ChatView(this, key: widget.key),
+      child: GestureDetector(
+        onTap: requestNotificationPermission,
```
nqhhdev, line 2030: Why not use `onEnter ` of `MouseRegion`.

Te-Z: if we use onEnter it won't support click. Also with gesture detector we will be able to ask notification if the user enters a room but for example just waits for a message

KhaledNjim (approved):
Tested on edge, firefox and chrome worked well in my case

nqhhdev:
I tested on Firefox, can't show pop-up request notification permission

https://github.com/user-attachments/assets/87443266-a6af-4822-9b0a-ea28a344bf7b

Te-Z:
> I tested on Firefox, can't show pop-up request notification permission
> 
>  Screen.Recording.2024-07-19.at.13.31.26.mov

Yes this is because notifications are already allowed on your test.
But I can confirm it works

image

hoangdat:
It is an improvement for notification in web only, not a fix for `receive notificaiton`. IMO, please update the title to more precise to prevent confuse when we need to trace back.
Moreover, we can not sure the problem of notification in mobile will not comeback. @Te-Z

lib/pages/chat_list/chat_list_item.dart
```diff
@@ -42,6 +44,14 @@ class ChatListItem extends StatelessWidget with ChatListItemMixin {
   });
 
   void clickAction(BuildContext context) async {
+    if (kIsWeb) {
```
sherlockvn, line 47: should use `PlatformInfos.isWeb`

lib/pages/chat/chat.dart
```diff
@@ -1971,6 +1972,16 @@ class ChatController extends State<Chat>
     }
   }
 
+  void requestNotificationPermission() {
+    if (kIsWeb) {
```
sherlockvn, line 1976: should use `PlatformInfos.isWeb`

nqhhdev:
Hi @Te-Z pls check all comments

lib/pages/chat/chat.dart
```diff
@@ -1971,6 +1972,16 @@ class ChatController extends State<Chat>
...
+      try {
```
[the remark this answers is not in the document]
Te-Z: No it does not throw any exception or return anything

Te-Z:
done @nqhhdev @sherlockvn

sherlockvn (approved):

Issue #1923, opened 2024-07-05
[Design] Corrections to do after design review


Te-Z:
@Bobpodvalnyi has made a review of the application.  This is one part of the needed corrections

Here's the link of his comments with the expected design for each point: https://www.notion.so/linagora/CHAT-UX-IU-changement-in-Chat-9267fd4b2a5348dcb2db13a7578e98a8

Please ask us if you have any question.

DoD: 

- [ ] 1 -> some searches screens still have the text
- [ ] 2 -> remove the "no results" and the related section if a result has been found for other section
- [ ] 3 -> change quick action colors
- [ ] 5 -> change the design of log out dialog
- [ ] 6 -> add padding at the top of search panel

PR #1961, opened 2024-07-22
TW-1923: Improve search screen

hieutbui:
## Ticket
- #1923 

## Root cause
- Currently, when search message is empty, search controller doesn't check for keyword is Matrix ID

## Solution
- When search message is empty, check if keyword is Matrix ID then hide empty widget.

## Test recommendations
- Search keywords that is a valid Matrix ID 
- Search keywords that is a valid Matrix ID but not complete (Ex: @test:domain.co)
- Send message with a Matrix ID inside a room then search for this Matrix ID

## Resolved
- Web:

https://github.com/user-attachments/assets/5a4374c7-98c8-4295-9d6a-0f6c324d6c61

- Android:
- IOS:

### DIFF
--- a/lib/pages/search/search.dart
+++ b/lib/pages/search/search.dart
@@ -58,7 +58,7 @@ class SearchController extends State<Search> {
 
   String get searchWord => textEditingController.text;
 
-  bool get isSearchWordUserId =>
+  bool get isSearchMatrixUserId =>
       searchWord.isValidMatrixId && searchWord.startsWith('@');
 
   String getBodyText(Event event, String searchWord) {
--- a/lib/pages/search/search.dart
+++ b/lib/pages/search/search.dart
@@ -58,6 +58,9 @@ class SearchController extends State<Search> {
 
   String get searchWord => textEditingController.text;
 
+  bool get isSearchWordUserId =>
+      searchWord.isValidMatrixId && searchWord.startsWith('@');
+
   String getBodyText(Event event, String searchWord) {
     final senderName = event.senderFromMemoryOrFallback.calcDisplayname(
       i18n: MatrixLocals(L10n.of(context)!),
--- a/lib/pages/search/search_view.dart
+++ b/lib/pages/search/search_view.dart
@@ -73,8 +73,10 @@ class SearchView extends StatelessWidget {
                 searchController.serverSearchController.searchResultsNotifier,
             builder: ((context, searchResults, child) {
               if (searchResults is PresentationServerSideEmptySearch) {
+                final keyword = searchController.textEditingController.text;
                 if (searchController.searchContactAndRecentChatController!
-                    .recentAndContactsNotifier.value.isNotEmpty) {
+                        .recentAndContactsNotifier.value.isNotEmpty ||
+                    (keyword.isValidMatrixId && keyword.startsWith('@'))) {
                   return child!;
                 }
                 return _SearchHeader(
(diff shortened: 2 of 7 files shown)


sherlockvn:
Screenshot 2024-07-25 at 11 02 37

lib/pages/search/search_view.dart
```diff
@@ -73,8 +73,10 @@ class SearchView extends StatelessWidget {
                 searchController.serverSearchController.searchResultsNotifier,
             builder: ((context, searchResults, child) {
               if (searchResults is PresentationServerSideEmptySearch) {
+                final keyword = searchController.textEditingController.text;
                 if (searchController.searchContactAndRecentChatController!
-                    .recentAndContactsNotifier.value.isNotEmpty) {
+                        .recentAndContactsNotifier.value.isNotEmpty ||
+                    (keyword.isValidMatrixId && keyword.startsWith('@'))) {
```
sherlockvn, line 79: better to make a function

lib/pages/search/server_search_view.dart
```diff
@@ -28,11 +28,13 @@ class ServerSearchMessagesList extends StatelessWidget {
             searchController.serverSearchController.searchResultsNotifier,
         builder: (context, serverSearchNotifier, child) {
           if (serverSearchNotifier is PresentationServerSideEmptySearch) {
+            final keyword = searchController.textEditingController.text;
             if (searchController.searchContactAndRecentChatController!
-                .recentAndContactsNotifier.value.isNotEmpty) {
-              return const SizedBox.shrink();
+                    .recentAndContactsNotifier.value.isEmpty &&
+                !(keyword.isValidMatrixId && keyword.startsWith('@'))) {
+              return child!;
```
sherlockvn, line 35: same

sherlockvn:
not in the DOD, but i think this case, we should remove the `Messages` section too
image

KhaledNjim:
> not in the DOD, but i think this case, we should remove the `Messages` section too image

Updated

image

nqhhdev:
I tested it on mobile, web works well

nqhhdev:
I have one concern about padding in chat search on web. WDYT @hoangdat @KhaledNjim 

- Currently 

Screenshot 2024-11-25 at 10 23 57

- After updated

Screenshot 2024-11-25 at 10 22 32

hoangdat:
> I have one concern about padding in chat search on web. WDYT @hoangdat @KhaledNjim

please base on the new style for this.

nqhhdev:
Reverted style
Screenshot 2024-12-04 at 23 48 15

KhaledNjim:
According to the notion page, the explanation text when no results are found should be removed, its still present in some screens
 - When searching for a contact while creating a new chat
 

https://github.com/user-attachments/assets/bea07524-85fb-494a-bf2b-4305521c1878


  - When searching in contacts screen
  
image

nqhhdev:
Hi @hoangdat  @KhaledNjim, I updated the search result in contact screen, add members into group and create a  new chat.
Pls help me review and test it again. Thanks all

https://github.com/user-attachments/assets/80effcdf-1b92-4ce6-88b1-9c266a85d244

lib/pages/search/search.dart
```diff
@@ -58,6 +58,9 @@ class SearchController extends State<Search> {
 
   String get searchWord => textEditingController.text;
 
+  bool get isSearchWordUserId =>
```
hoangdat, line 61: Can you find a better name for this? Still not understand what is it? What is use for?

Issue #1901, opened 2024-06-28
[CRASH APP] Crash app when open and play video on Mobile


nqhhdev:
### Env:
- iOS version: 17.5.1
- Twake version: 2.5.9
- Homeserver: `matrix.org`

### Desc:
- Send video from web and received on mobile.
- Crash app when open or playing video on chat.
- After playing video and staying at chat. After a while, the app also crashed

### Note:
- video sent from matrix.org
- received on matrix.linagora.com

https://github.com/linagora/twake-on-matrix/assets/99852347/73a5fca5-19c2-482a-8b02-aec6cda88910

sherlockvn:
![Image](https://github.com/linagora/twake-on-matrix/assets/43041967/42c45e22-9e5c-48e4-b53a-5a278f2355b6)

sherlockvn:
This happend only during Huy test uploading video and images file. need to verify again

Te-Z:
Still occurs on iOS (tested by Bob on iPhone 11, app version: 2.6.0)

PR #1979, opened 2024-07-25
TW-1901: crash app when open play video mobile

sherlockvn:
## Ticket
**Related issue**
#1901 
## Root cause
Sometimes the thumbnail of video is not an image (this is due to testing or old video or maybe come from another platform), so that when display to screen, it will throw an exception during decode the image.
image

## Solution
Now when create a thumbnail, we will add the extension which indicate its an image, and in the receiver side, if the thumbnail is not the image, we will throw an exception and handle the exception.
## Impact description

## Test recommendations
Try to send video and image in both normal chat and encrypted chat
## Pre-merge

## Resolved
- Web:

https://github.com/user-attachments/assets/4bd42166-e03b-48d7-ac92-a90f0a24b870

- Android:

https://github.com/user-attachments/assets/d4427093-aab3-49d4-a367-aaae51554a59

- IOS:

https://github.com/user-attachments/assets/e204565f-4f22-4382-a240-364960769f3c

### DIFF
--- a/lib/presentation/extensions/send_file_extension.dart
+++ b/lib/presentation/extensions/send_file_extension.dart
@@ -575,7 +575,7 @@ extension SendFileExtension on Room {
         originalFile.filePath,
         targetPath,
         quality: AppConfig.thumbnailQuality,
-        format: CompressFormat.jpeg,
+        format: AppConfig.imageCompressFormmat,
       );
       if (result == null) return null;
       final size = await result.length();
@@ -591,7 +591,7 @@ extension SendFileExtension on Room {
       }
       uploadStreamController?.add(const Right(GenerateThumbnailSuccess()));
       return ImageFileInfo(
-        result.name,
+        '${result.name}.${AppConfig.imageCompressFormmat.name}',
         result.path,
         size,
         width: width,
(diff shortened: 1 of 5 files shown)


lib/widgets/mxc_image.dart
```diff
@@ -182,26 +182,33 @@ class _MxcImageState extends State<MxcImage> {
...
-        }
+        if (!mounted) return;
+        _imageData = matrixFile.bytes;
+        return;
+      } catch (e) {
+        Logs().e('MxcImage::Error while downloading image: $e');
+      } finally {
+        setState(() {
+          isLoadDone = true;
+        });
```
nqhhdev, line 209: IMO: If the case has error, still go on it. This func only used for success case, pls check it again

lib/presentation/extensions/send_file_extension.dart
```diff
@@ -570,12 +570,12 @@ extension SendFileExtension on Room {
   }) async {
     try {
       uploadStreamController?.add(const Right(GeneratingThumbnailState()));
-
+      const compressFormmat = CompressFormat.jpeg;
```
nqhhdev, line 573: IMO: Move it on `AppConfig`

nqhhdev (approved):

Te-Z (approved):

PR #1982, opened 2024-07-25
TW-1964: handle full matrix id on presentationContactNotifier fail

Te-Z:
## Ticket
**Related issue**
https://github.com/linagora/twake-on-matrix/issues/1964

## Root cause
**If this is a bug, please provide a brief description of the root cause of the issue**
The external search was not handled on presentationContactNotifier failure

## Solution
**Outline the implemented solution, detailing the changes made and how they address the issue**
handle this case

## Resolved
**Attach screenshots or videos demonstrating the changes**
- Web:
Chrome:

[Capture vidéo du 26-07-2024 14:42:10.webm](https://github.com/user-attachments/assets/216654e4-56d1-4010-b77f-cc822e371a2c)

[Capture vidéo du 26-07-2024 14:36:51.webm](https://github.com/user-attachments/assets/5231c6ac-0181-4e5d-bf32-49ee6ee7f168)

- Android:

https://github.com/user-attachments/assets/9ce18539-b5f2-48c0-80cc-25a31ce145e5

- IOS:

Issue #1963, opened 2024-07-22
[Bug] Cannot search by partial or full username, or Matrix ID


tprudentova:
2.6.0, staging, tested on Android and IOS; Safari, Firefox

1. Tap on the search bar in the Chats or Contacts tab
2. Try entering a username (email) or Matrix ID
3. No search results

Expectation: I can search by username

https://github.com/user-attachments/assets/fc34b9ac-fd62-4e10-a22d-f1803871cde4

https://github.com/user-attachments/assets/59993b6c-c6e1-4daa-9134-0ca9c4aa7605

sherlockvn:
Right now, it searches only contacts, but contacts don't have in stg

tprudentova:
Tested on 2.6.1, still cannot search by username through contacts with whom I had a chat


https://github.com/user-attachments/assets/1ea474fe-0436-498d-a411-cb4f9842b30a

hoangdat:
hi @nqhhdev any update on this ticket?

nqhhdev:
> hi @nqhhdev any update on this ticket?

In progress

nqhhdev:
#1983 Fixed for this case. Pls check it @hoangdat

PR #1983, opened 2024-07-25
TW-1963: update search keyword matching

Te-Z:
## Ticket
**Related issue**
https://github.com/linagora/twake-on-matrix/issues/1963

## Need merge first

- #1982 

## Root cause
Search keywork did not include matrix id. Also the `!` without checking first if the thing is null or not throwed some errors which blocked the search.

## Solution
Add matrix id matchers
Check if value is null, if it is return a boolean as default value.

## Resolved
- Web:
- Chrome

[Capture vidéo du 25-07-2024 21:24:31.webm](https://github.com/user-attachments/assets/02b55b38-df68-49b9-8ff5-c0232ae7874c)

- Android:

https://github.com/user-attachments/assets/a96bd246-4005-4543-94fb-2e06200e61d0

- IOS:

### DIFF
--- a/lib/domain/model/room/room_list_extension.dart
+++ b/lib/domain/model/room/room_list_extension.dart
@@ -13,11 +13,19 @@ extension RoomListExtension on List<Room> {
     )
         .map((room) => room.toRecentChatSearchModel(matrixLocalizations))
         .where(
-          (model) =>
-              model.displayName != null &&
-              model.displayName!.toLowerCase().contains(
-                    keyword.toLowerCase(),
-                  ),
+          (model) {
+            final matchedMatrixId = model.directChatMatrixID
+                    ?.toLowerCase()
+                    .contains(keyword.toLowerCase()) ??
+                false;
+
+            final matchedName = model.displayName
+                    ?.toLowerCase()
+                    .contains(keyword.toLowerCase()) ??
+                false;
+
+            return matchedName || matchedMatrixId;
+          },
         )
         .take(limit ?? length)
         .toList();
--- a/lib/domain/model/room/room_list_extension.dart
+++ b/lib/domain/model/room/room_list_extension.dart
@@ -3,6 +3,22 @@ import 'package:fluffychat/domain/model/search/recent_chat_model.dart';
 import 'package:matrix/matrix.dart';
 
 extension RoomListExtension on List<Room> {
+  bool _matchedMatrixId(RecentChatSearchModel model, String keyword) {
+    return model.directChatMatrixID
+            ?.toLowerCase()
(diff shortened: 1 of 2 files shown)


nqhhdev:
We have searched in contact tab, create group, create direct chat. Pls think carefully about this, we have to consider all the cases for different environments. Be careful and do it thoroughly or we will get unpredictable consequences.

- How about contact tab (if phone book, recent Chat and ToM contact available)
- How about create group (if phone book, recent Chat and ToM contact available)
- How about create direct chat (if phone book and ToM contact available)
- How about env support Twake and not support Twake

Te-Z:
Hi @nqhhdev I didn't find any problem. Could you provide more details about what's wrong ?

nqhhdev:
> Hi @nqhhdev I didn't found any problem. Could you provide more details about what's wrong ?

Pls answer my question and add some demo for this case

lib/domain/model/room/room_list_extension.dart
```diff
@@ -13,11 +13,19 @@ extension RoomListExtension on List<Room> {
...
+                    .contains(keyword.toLowerCase()) ??
+                false;
+
+            final matchedName = model.displayName
+                    ?.toLowerCase()
+                    .contains(keyword.toLowerCase()) ??
+                false;
+
+            return matchedName || matchedMatrixId;
+          },
```
hoangdat, line 28: separate method

lib/pages/search/search_contacts_and_chats_controller.dart
```diff
@@ -96,13 +96,7 @@ class SearchContactsAndChatsController with SearchDebouncerMixin, SearchMixin {
...
       }
 
-      final matchedName =
-          contact.displayName!.toLowerCase().contains(keyword.toLowerCase());
-
-      final matchedEmail =
-          contact.email!.toLowerCase().contains(keyword.toLowerCase());
-
-      return matchedName || matchedEmail;
+      return _doesMatchKeyword(contact, keyword);
```
hoangdat, line 99: separate method for where operation

KhaledNjim (approved):
Works well for `matrix.linagora.com` and `matrix.stg.lin-saas.com`

lib/domain/model/room/room_list_extension.dart
nqhhdev: Done

lib/pages/search/search_contacts_and_chats_controller.dart
nqhhdev: Done

hoangdat (approved):

Issue #1980, opened 2024-07-25
[Web] HEIC, HEIF thumbnail can't display in web


sherlockvn:

nqhhdev:
### DoD:

- [ ] Convert `HEIC`, `HEIF` to `JPEG`
- [ ] Android `WEBP` to `JPEG`

hoangdat:
what is the progress of this issue? @sherlockvn

sherlockvn:
its done in the ticket https://github.com/linagora/twake-on-matrix/pull/1988

tprudentova:
The current behaviour is different for web (Safari) and mob (Android) (ver 2.6.5) 

For web: select a HEIC file -> send -> it is sent as HEIC file with no thumbnail 

For mobile: select a HEIC file -> send -> it is sent as a JPEG file but will no thumbnail 

Expectation: the file is converted to JPEG and is sent with a thumbnail (as if it were a basic JPEG image)


https://github.com/user-attachments/assets/16321258-0b18-4b00-87ad-fc269c3023b3

https://github.com/user-attachments/assets/7fa76ae3-624f-49d2-a27e-83327529345b

PR #1988, opened 2024-08-05
TW-1980: Convert HEIC files to jpg file

sherlockvn:
## Ticket
**Related issue**
#1980 
## Root cause
HEIC, HEIF image format can't display in web platform
## Solution
We have to convert HEIC, HEIF format to common image format like JPG when send images from iOS, so that it can be displayed in other platform. 

- `https://pub.dev/packages/image_editor_plus` has an output is raw data not file, so i don't use it, it may freeze the app
## Impact description

## Test recommendations
Try sending HEIC, HEIF files from iOS device, then see that its displayed well in other platforms or not
## Pre-merge

## Resolved
- Web:

https://github.com/user-attachments/assets/1998e708-dbe0-4c91-ab27-9a613ff81ba7

https://github.com/user-attachments/assets/e42c32d0-075f-47a3-b828-54290686d20e

- Android:

https://github.com/user-attachments/assets/58973c9d-81ec-485d-8f97-e2038abb514a

- IOS:

### DIFF
--- a/lib/presentation/extensions/send_file_extension.dart
+++ b/lib/presentation/extensions/send_file_extension.dart
@@ -20,6 +20,7 @@ import 'package:fluffychat/utils/extension/mime_type_extension.dart';
 import 'package:fluffychat/utils/manager/storage_directory_manager.dart';
 import 'package:fluffychat/utils/manager/upload_manager/upload_state.dart';
 import 'package:flutter/widgets.dart';
+import 'package:heif_converter/heif_converter.dart';
 import 'package:image/image.dart' as img;
 import 'package:blurhash_dart/blurhash_dart.dart';
 import 'package:flutter/foundation.dart';
@@ -49,7 +50,6 @@ extension SendFileExtension on Room {
     CancelToken? cancelToken,
     DateTime? sentDate,
   }) async {
-    FileInfo tempfileInfo = fileInfo;
     // Check media config of the server before sending the file. Stop if the
     // Media config is unreachable or the file is bigger than the given maxsize.
     try {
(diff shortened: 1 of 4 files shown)


lib/presentation/extensions/send_file_extension.dart
```diff
@@ -418,6 +427,35 @@ extension SendFileExtension on Room {
...
+  Future<ImageFileInfo> convertHeicToJpgImage(ImageFileInfo fileInfo) async {
+    final convertedFilePath =
+        StorageDirectoryManager.instance.convertFileExtension(
+      fileInfo.filePath,
+      'jpg',
+    );
+    final newPath = await HeifConverter.convert(
+      fileInfo.filePath,
+      output: convertedFilePath,
+    );
```
nqhhdev, line 439: IMO: Should use `try catch` in this case. Because maybe while converting an error may occur.

lib/presentation/extensions/send_file_extension.dart
```diff
@@ -418,6 +427,35 @@ extension SendFileExtension on Room {
...
+    Logs().d('sendFileEvent::Heic converted to jpg', newPath);
+    if (newPath != null) {
+      final newConvertedFile = File(convertedFilePath);
+      fileInfo = ImageFileInfo(
+        newConvertedFile.path.split("/").last,
+        newConvertedFile.path,
+        await newConvertedFile.length(),
```
nqhhdev, line 446: What happens if this func fails?

lib/presentation/extensions/send_file_extension.dart
sherlockvn: already, try/catch outside
image

lib/presentation/extensions/send_file_extension.dart
sherlockvn: if the function fails it just continue the execution

sherlockvn: try/catch already

nqhhdev (approved):
I tested it on:
- Moblie (Android & iOS)
- Web browsers
- env: support `ToM` and `matrix.org`

Te-Z (approved):

Issue #2114, opened 2024-10-29
Display phone number as a clickable in message bubble


hoangdat:
### Desc
- when message contain phone number, we should support some actions

### DoD:
- [ ] can copy number
- [ ] can call directly

hoangdat:
- library: https://pub.dev/packages/linkfy_text
- `flutter_matrix_html`: use `linkfy_text` in this lib ?????

PR #2120, opened 2024-11-04
TW-2114: Display phone number as a clickable in message bubble

nqhhdev:
## Ticket
- #2114 

## Need merge

- https://github.com/linagora/flutter_matrix_html/pull/19
- https://github.com/nqhhdev/linkfy_text/pull/1

- Web:
- Android:
- IOS:

### DIFF
--- a/lib/pages/chat/chat_context_menu_actions.dart
+++ b/lib/pages/chat/chat_context_menu_actions.dart
@@ -8,7 +8,8 @@ enum ChatContextMenuActions {
   pinChat,
   forward,
   downloadFile,
-  jumpToMessage;
+  jumpToMessage,
+  copyNumber;
 
   String getTitle(
     BuildContext context, {
@@ -30,6 +31,8 @@ enum ChatContextMenuActions {
         return L10n.of(context)!.download;
       case ChatContextMenuActions.jumpToMessage:
         return L10n.of(context)!.jumpToMessage;
+      case ChatContextMenuActions.copyNumber:
+        return L10n.of(context)!.copyNumber;
     }
   }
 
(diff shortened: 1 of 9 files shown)


lib/utils/url_launcher.dart
```diff
@@ -73,7 +73,7 @@ class UrlLauncher with GoToDraftChatMixin {
           return;
         }
       }
-      launchUrlString(url!);
+      launchUrlString('https://$url');
```
tddang-linagora, line 76: If `url` already have a scheme, will this work? E.g. "tel:123456" or "mailto:abc@example.com"

nqhhdev: Don't worry about this case. In this case I was verified `url` with scheme if it was valid to return `url` if not only display text. You can see `RegExp` to verify `url` in here https://github.com/nqhhdev/linkfy_text/blob/9f4b9debbe4da621897ea040f24e66da3c9ba980/lib/src/utils/contants.dart#L1718

lib/pages/chat/chat_context_menu_actions.dart
```diff
@@ -8,7 +8,8 @@ enum ChatContextMenuActions {
   pinChat,
   forward,
   downloadFile,
-  jumpToMessage;
+  jumpToMessage,
+  copyNumber;
```
hoangdat, line 12: Should create `PhoneNumberContextMenu`

lib/utils/url_launcher.dart
hoangdat: still not clear to me @nqhhdev

lib/presentation/mixins/linkify_mixin.dart
```diff
@@ -0,0 +1,190 @@
...
+    );
+    final selectedActionIndex = await showTwakeContextMenu(
+      context: context,
+      offset: offset,
+      listActions: listActions,
+      onClose: _handleStateContextMenu,
+    );
+    if (selectedActionIndex != null && selectedActionIndex is int) {
+      _handleClickOnContextMenuItem(
+        action: phoneNumberContextMenuOnWeb[selectedActionIndex],
```
hoangdat, line 60: why not `PhoneNumberContextMenuActions.copy`?

hoangdat (approved):

Issue #2124, opened 2024-11-08
[BUG] Wrong url when using logout


nqhhdev:
https://github.com/user-attachments/assets/b041d531-b904-40c3-9472-0bccf573811d

PR #2126, opened 2024-11-08
TW-2124: Fixed wrong url when using logout

nqhhdev:
## Ticket
- #2124 

## Root cause
- Side effect #2092 

## Resolved

- Web

https://github.com/user-attachments/assets/a7aa7fed-75fe-40c2-a3d1-d021eff98541

- Mobile

https://github.com/user-attachments/assets/1f962c59-cbd4-4a53-bd8b-2eb8c146f247

### DIFF
--- a/lib/utils/string_extension.dart
+++ b/lib/utils/string_extension.dart
@@ -382,7 +382,7 @@ extension StringCasingExtension on String {
     return fragmentIndex != -1 ? substring(0, fragmentIndex) : this;
   }
 
-  String generateAuthPath({
+  String generateLoginAuthPath({
     String? homeserverParams,
     bool isDevMode = false,
   }) {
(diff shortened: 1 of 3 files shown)


lib/utils/string_extension.dart
```diff
@@ -382,7 +382,7 @@ extension StringCasingExtension on String {
     return fragmentIndex != -1 ? substring(0, fragmentIndex) : this;
   }
 
-  String generateAuthPath({
+  String generateLoginAuthPath({
     String? homeserverParams,
```
hoangdat, line 386: Should be non-null parameter

lib/utils/string_extension.dart
```diff
@@ -382,14 +382,22 @@ extension StringCasingExtension on String {
...
     bool isDevMode = false,
+  }) {
+    final newHomeserverParams = homeserverParams?.trim() ?? '';
+    final path = isDevMode ? 'web/auth.html' : 'auth.html';
+    return '$this$path$newHomeserverParams';
+  }
+
+  String generateLogoutAuthPath({
```
hoangdat, line 394: write unit test for these function

nqhhdev: Done

hoangdat (approved):

Issue #2068, opened 2024-09-25
[Bug] After accepting an invitation I need to reload chat to see messages


tprudentova:
2.6.5, staging, Android, IOS 

1. Send an invitation from a staging account to a linagora account. Send some messages
2. Open the invitation on linagora account (mobile app), accept it 
3. I see a unclickable greeting. I don't see any messages
4. If I go to the chat list and come back to the chat, the messages appear

<img width="1280" alt="Screenshot 2024-09-25 at 16 52 04" src="https://github.com/user-attachments/assets/a9fdc723-d827-4b99-99f8-66126a29349e">

https://github.com/user-attachments/assets/c6d93314-288f-4466-a551-33dcf3b9b2ae

PR #2133, opened 2024-11-13
TW-2068: Fix can't see message after accept invitation

nqhhdev:
## Ticket
 - #2068 

## Resolved

- Error case:

https://github.com/user-attachments/assets/01d5baa8-4a55-459e-b3b7-fc5084937d5d

- Mobile:

https://github.com/user-attachments/assets/f2f2632a-52d0-47a9-9f4b-4b45c3956bd2

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1588,7 +1588,7 @@ class ChatController extends State<Chat>
         await room?.join();
         await waitForRoom;
       },
-    );
+    ).then((_) => _tryLoadTimeline());
   }
 
   void onRejectInvitation(BuildContext context) async {


lib/pages/chat/chat.dart
```diff
@@ -1588,7 +1588,7 @@ class ChatController extends State<Chat>
         await room?.join();
         await waitForRoom;
       },
-    );
+    ).then((_) => _tryLoadTimeline());
```
hoangdat, line 1591: Please show the demo for both case, error and success.

nqhhdev: IMO: Currently, for the error case, still using old UI from `showFutureLoadingDialog` so If we need to update UI for this I'll create a new ticket and discuss with Design team to improve it.

KhaledNjim (approved):
Works well in my case

hoangdat (approved):

PR #2210, opened 2025-01-15
TW-2206 added patrol test for sending text message

hoangdat:


integration_test/scenarios/send_text_message_scenario.dart
```diff
@@ -0,0 +1,24 @@
...
+    required this.loginScenario,
+  });
+
+  @override
+  Future<void> execute() async {
+    await loginScenario.execute();
+    await $.tap($(ChatListItem));
+    await $.waitUntilVisible($(ChatView));
+    await $.enterText($(InputBar), "test message");
+    await $.tap($(ChatInputRowSendBtn));
```
hoangdat, line 22: we should have a way to verify it work? For example:
- verify sending status
- verify sent status

PR #2282, opened 2025-04-03
[TEMP] add box and bump version but forget bump db version

hoangdat:


hoangdat:
- MUST BE UPGRADE VERSION FOR DB

PR #2289, opened 2025-04-11
Sync contact error handle link

nqhhdev:
## Root cause

- Has many request to FED API and then phonebook contact not display.

## Resolved

https://github.com/user-attachments/assets/95253644-78bc-4b1d-a83f-6ac5c12f42e8

### DIFF
--- a/lib/domain/usecase/contacts/federation_look_up_phonebook_contact_interactor.dart
+++ b/lib/domain/usecase/contacts/federation_look_up_phonebook_contact_interactor.dart
@@ -149,7 +149,6 @@ class FederationLookUpPhonebookContactInteractor {
               exception: 'Hash details is empty',
             ),
           );
-          return;
         }
 
         hashDetails = res;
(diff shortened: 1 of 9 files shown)


lib/domain/usecase/contacts/federation_look_up_phonebook_contact_interactor.dart
```diff
@@ -149,7 +149,6 @@ class FederationLookUpPhonebookContactInteractor {
               exception: 'Hash details is empty',
             ),
           );
-          return;
```
hoangdat, line 152: No `HashDetail` we still keep lookup working?

lib/domain/usecase/contacts/federation_look_up_phonebook_contact_interactor.dart
```diff
@@ -162,7 +161,6 @@ class FederationLookUpPhonebookContactInteractor {
             exception: e,
           ),
         );
-        return;
```
hoangdat, line 165: idem

hoangdat (approved):

PR #2327, opened 2025-05-07
TW-2323 Disable/Enable invitation base on well-known properties

hoangdat:
### Without `enable_invitations` in well-known
https://github.com/user-attachments/assets/bad32cb5-f38f-436c-ac52-192d7fe7d421

### With `enable_invitations` in well-known
https://github.com/user-attachments/assets/37e5b829-f859-4e58-a70d-a8e8eee6fd5f


nqhhdev (approved):
LGTM

PR #2331, opened 2025-05-13
Fix crash when video display in chat with wrong thumbnail information

hoangdat:
### Root cause
- thumbnail information wrong, but app put the video into preview image for this video
- media uri API endpoint deprecated from spec v1.11 

https://github.com/user-attachments/assets/791abd46-c451-489f-80c2-271dcb588e7f


hoangdat:
image

nqhhdev (approved):

Issue #2320, opened 2025-04-28
Emoji picker for message composer


hoangdat:
### DoD

PR #2332, opened 2025-05-13
TW-2320: emoji picker for message compose

nqhhdev:
## Ticket
 - #2320 

## Resolved
- Web:

https://github.com/user-attachments/assets/17547f5c-6aae-4f83-a3df-1189dfb7f52d


hoangdat:
- [ ] we should display emoji picker for composer in tablet also

hoangdat:
- [ ] input focus must also be stay in composer while we open the picker and select the emoji

lib/config/app_config.dart
```diff
@@ -33,6 +33,10 @@ abstract class AppConfig {
 
   static const double messagePadding = 16.0;
 
+  static const double defaultMaxWidthReactionPicker = 400;
```
hoangdat, line 36: why we have 2 the same constant with different value?

nqhhdev: My fault, this is size for bottom sheet when open reactions view. I'll change the name

hoangdat (approved):

PR #2337, opened 2025-05-15
🐛 fix SSO logout for web

rezk2ll:
## Ticket
image

Cannot log out on the web when using LemonLDAP as SSO

## Root cause
LemonLDAP does not accept the `url` parameter if the host is not configured in the vhosts or is not an HTTP URL.

## Solution
removes the `url` parameter from SSO logout

### DIFF
--- a/lib/presentation/mixins/connect_page_mixin.dart
+++ b/lib/presentation/mixins/connect_page_mixin.dart
@@ -83,14 +83,11 @@ mixin ConnectPageMixin {
     return '${AppConfig.registrationUrl}?$redirectPublicPlatformOnWeb=$redirectUrlEncode&app=${AppConfig.appParameter}';
   }
 
-  String? _getLogoutUrl(
-    BuildContext context, {
-    required String redirectUrl,
-  }) {
+  String? _getLogoutUrl(BuildContext context) {
     final authUrl = Matrix.of(context).authUrl;
     if (authUrl == null) return null;
-    final redirectUrlEncode = base64Url.encode(utf8.encode(redirectUrl));
-    return '$authUrl?logout=1&url=$redirectUrlEncode';
+
+    return '$authUrl?logout=1';
   }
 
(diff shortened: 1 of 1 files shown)


hoangdat:
- so we dont need url anymore? 
- we removed url in mobile logout, why not config to keep it in web logout?

rezk2ll:
> * so we dont need url anymore?
> * we removed url in mobile logout, why not config to keep it in web logout?

We don't need the URL, yes, and currently, it's blocking logout on platforms like linagora.com, etc. Web logout now fails because of it. a simple call to the `?logout=1` will do the job.

hoangdat (approved):

hoangdat:
> Web logout now fails 

But as you said, `https url` is accepted. So why we not config it in lemonldap to accept our url in web? 
Moreover, when we remove it, we can redirect?

rezk2ll:
> > Web logout now fails 
> 
> But as you said, `https url` is accepted. So why we not config it in lemonldap to accept our url in web? 
> Moreover, when we remove it, we can redirect?

lemon will only accept urls that are configured in the vhosts and are protected by it. it will display a countdown screen for redirection etc. and redirecting back to a url in web context is not useful because the portal handles the rest and we will need to display a login form right away

dab246:
IMO, This only applies to `Web`, on `Mobile` still need `redirectUrl`, I tested and it does not automatically close the browser after logout. See video below. Should be fixed.

[demo.webm](https://github.com/user-attachments/assets/5bed57e5-d633-4783-8a09-36a0eb9de462)

rezk2ll:
> IMO, This only applies to `Web`, on `Mobile` still need `redirectUrl`, I tested and it does not automatically close the browser after logout. See video below. Should be fixed.
> 
>  demo.webm

For mobile, we agreed to use synapse logout only, as there is no need to log out from SSO when on mobile

PR #2417, opened 2025-06-23
set up patrol for IOS

thuyenh1081:


hoangdat:
look good to me, but should have demo for this PR

thuyenh1081:
https://github.com/user-attachments/assets/09329c9f-b652-4d00-b987-70e6aa7b79fb

upload video for running tc under second commit

thuyenh1081:
add video for running test case: sending message between group members

https://github.com/user-attachments/assets/c467895f-a6e4-4d34-865a-fc87191467d7

integration_test/base/core_robot.dart
```diff
@@ -6,4 +9,393 @@ abstract class CoreRobot {
...
+    )) {
+      await $.native.grantPermissionWhenInUse();
+    }
+  }
+
+  Future<void> waitForEitherVisible({
+  required PatrolIntegrationTester $,
+  required PatrolFinder first,
+  required PatrolFinder second,
+  required Duration timeout ,
```
hoangdat: indent

integration_test/base/core_robot.dart
hoangdat, line 1: a lot of indent in this file

integration_test/tests/message_test.dart
```diff
@@ -0,0 +1,38 @@
...
+        //// receiver read message
+        //
+        //// check sender see message is read
+        //...
+        // send message by API
+        await Future.delayed(const Duration(seconds: 10)); 
+        await MessageScenario($).sendAMessageByAPI(groupID, messageOfReceiver);
+        // check message is shown on UI
+        await Future.delayed(const Duration(seconds: 30)); 
+        await MessageScenario($).verifyMessageIsShown(messageOfReceiver);
```
hoangdat: still not really understand for me?
```
a user send a message via API
then who verify?
```
Can you find the better way to describe the test case, and make all people understand what you are aiming?

integration_test/tests/chat_list_test.dart
```diff
@@ -0,0 +1,32 @@
...
+    },
+  );
+
+  TestBase().runPatrolTest(
+    description: 'Open a chat group',
+    test: ($) async {
+      await TestBase().loginAndRun($, (_) async {
+      await ChatListRobot($).openChatGroupByIndex(5);
+      //verify chat group is opened
+      expect(await ChatGroupDetailRobot($).isVisible(),isTrue);
```
hoangdat: indent

integration_test/tests/message_test.dart
hoangdat: For example:
```
user A send message by API
current user verify this message was received 
bla bla bla
```

integration_test/base/core_robot.dart
```diff
@@ -6,4 +9,393 @@ abstract class CoreRobot {
...
+
+  sixthRequest.write(jsonEncode({
+    'msgtype': 'm.text',
+    'body': message,
+  }),);
+
+  await sixthRequest.close();
+}
+
+Future<void> closeClient(HttpClient client) async {
```
hoangdat: what is it? why we need it?

integration_test/robots/chat_group_detail_robot.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:linkfy_text/linkfy_text.dart';
+import 'package:patrol/patrol.dart';
+
+import '../base/core_robot.dart';
+
+class ChatGroupDetailRobot extends CoreRobot {
+  ChatGroupDetailRobot(super.$);
+
+  Future<bool> isVisible() async {
+    final chatListSelector = $(ChatEventList);
```
hoangdat: still not really understand this method, why `ChatGroupDetailRobot` have the `isVisible` method but check the visible of `ChatEventList`? does make sense?

integration_test/base/core_robot.dart
Crash--: Don't we have IDE extension and common guidelines for that kind of stuff? We should not waste time on that kind of stuff, it should be done when saving the file.

integration_test/robots/chat_group_detail_robot.dart
thuyenh1081: This method is used to check the display of Group Chat Detail screen. But now, I realize that we can use         expect($(ChatEventList).exists, isTrue); to check the display of Group Chat Detail screen. So, this method will be deleted

integration_test/base/core_robot.dart
thuyenh1081: we initialize a http request to do some methods ( put, post, get ...). After using HTTP, we should close it before we start with other request

integration_test/tests/message_test.dart
thuyenh1081: I will update comments so that readers can understand it more easily

hoangdat:
- [ ] docs to mapping testcases with patrol test

Issue #2060, opened 2024-09-23
[BUG] [WEB] Unable & very difficult to select text in message bubble with mouse


dab246:
### Env: 

- Version: `2.6.5`
- Server: `stg.lin-saas.com`
- Platform: `Chrome, Edge(Macbook)`

### Steps to reproduce

1. Open a group or direct chat
2. Scroll to middle message with large text
3. Use mouse to drag select text in message bubble.

## Expectation

Easily select text by highlighting to copy

## Evidence

https://github.com/user-attachments/assets/98b6ce8d-d6ac-4bfc-bf48-5552c618ebbb

tprudentova:
Connected with #1935

hoangdat:
- [ ] technical debt: whole screen is rebuilding while some new events coming (sync function)

PR #2435, opened 2025-06-30
TW-2060 Fix unable to select text in message bubble with mouse on web

dab246:
## Ticket

#2060 

## Root cause

There are many causes for this error,  typically:

- `SelectionArea` wraps a scrollable widget (such as `ListView`, `SingleChildScrollView`), but some child widgets inside it interfere with pointer events (using `Stack, GestureDetector, InkWell, MouseRegion`, etc.), or the layout becomes unstable during drag selection. As a result, Flutter miscalculates the drag target, leading to the error `"drag target size is larger than scrollable size"` and preventing proper text selection.
- Some exceptions are thrown when manipulating messages.

## Solution

- Replace`OverflowView.flexible`  with using responsive check
- Remove redundant widgets when not needed

## Resolved

https://github.com/user-attachments/assets/9021b7c0-13ad-4d79-9303-48e6f8552797

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -238,8 +238,6 @@ class ChatController extends State<Chat>
 
   final FocusNode rawKeyboardListenerFocusNode = FocusNode();
 
-  String selectionText = "";
-
   Timer? typingCoolDown;
   Timer? typingTimeout;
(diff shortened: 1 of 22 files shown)


nqhhdev (approved):
LGTM

PR #2463, opened 2025-07-21
Enable forced synchronization of contacts when accessing the contacts tab

nqhhdev:
## Ticket
**Related issue**

- https://www.notion.so/linagora/Verify-contact-synchronizing-failed-23762718bad1805d86b9da435f54c029

## Root cause

- Missing Federation API callback condition when force updating in contact tab

## Solution

- Add condition when force updating in contact tab

## Resolved

https://github.com/user-attachments/assets/2db49a8a-a8be-4cb3-9f1d-bb61af9e8910


hoangdat:
it make the contact manager work as normal, but still not trigger every time open the contact.

Fx:
1. open app the first time, open contact page
2. after finished synchronizing, kill the app
3. open the app again, open contact page -> nothing

PR #2487, opened 2025-07-31
Simplify web setup related to libolm

zatteo:
During a setup of Twake Chat for web, I ran into the section about libolm.

It seems to be working when downloading olm.js and olm.wasm from here https://gitlab.matrix.org/matrix-org/olm/-/packages/2190 instead of building it with Docker. 

Why do you think ? It seems a lot simpler for me to setup, but I lack the necessary background and maybe it will not work everywhere/everytime ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,19 +73,8 @@ flutter doctor -v
 
 If you only plan to work on the `web` target we recommend installing Google Chrome as it is the default supported target *(Flutter being developped by Google)*.
 
-It is also **required** to have a web ready version of libolm available in the `assets/js/package` folder. You can build a version using:
+It is also **required** to have a web ready version of libolm available in the `assets/js/package` folder. You can download it [here](https://gitlab.matrix.org/matrix-org/olm/-/packages/2190) and copy olm.js and olm.wasm in `assets/js/package`.
 
-```bash
-docker run -v ./assets/js/package:/package nixos/nix:2.22.1
-
-# within the docker
-nix build -v --extra-experimental-features flakes --extra-experimental-features nix-command gitlab:matrix-org/olm/3.2.16?host=gitlab.matrix.org\#javascript
-cp /result/javascript/* /package/. -v
-exit
-
-# back on your host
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,19 +73,8 @@ flutter doctor -v
 
 If you only plan to work on the `web` target we recommend installing Google Chrome as it is the default supported target *(Flutter being developped by Google)*.
 
-It is also **required** to have a web ready version of libolm available in the `assets/js/package` folder. You can build a version using:
+It is also **required** to have a web ready version of libolm available in the `assets/js/package` folder. You can download it [here](https://gitlab.matrix.org/matrix-org/olm/-/packages/2190) and copy olm.js and olm.wasm in `assets/js/package`.
```
hoangdat, line 76: IMO, we should keep both guideline for user.

zatteo: I can add it so we have both. But is there any good reason to the Docker way? Even if I have no background to this library, I do not see the purpose of building it locally if its a web library?

zatteo: Why close @hoangdat ? Keeping both can be a safe solution, but it was easier and it worked on my local setup by just downloading it from the link. I can create a new PR with both solution.

jbrnzl:
Should we close this or it is still relevant @zatteo ?

zatteo:
> Should we close this or it is still relevant @zatteo ?

It helped me during the setup I did 8 month ago. But no idea if libolm is still required.

pm-McFly:
as per the prepare web script: https://github.com/linagora/twake-on-matrix/blob/main/scripts/prepare-web.sh#L8

libolm is not needed anymore, ref to lib olm should be removed

jbrnzl:
> as per the prepare web script: https://github.com/linagora/twake-on-matrix/blob/main/scripts/prepare-web.sh#L8
> 
> libolm is not needed anymore, ref to lib olm should be removed

I created a follow-up issue here #3144. I think we can close this PR for now.

PR #2500, opened 2025-08-18
Apply dynamic cozy bridge version

tddang-linagora:
## Pre-merge
- https://github.com/linagora/emoji_mart/pull/8

### DIFF
--- a/lib/config/app_config.dart
+++ b/lib/config/app_config.dart
@@ -296,7 +296,8 @@ abstract class AppConfig {
     if (json['support_url'] is String) {
       supportUrl = json['support_url'];
     }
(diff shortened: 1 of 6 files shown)


lib/config/app_config.dart
```diff
@@ -294,5 +295,8 @@ abstract class AppConfig {
     if (json['support_url'] is String) {
       supportUrl = json['support_url'];
     }
+    if (json['cozy_external_bridge_version'] is String) {
+      cozyExternalBridgeVersion = json['cozy_external_bridge_version'];
```
hoangdat, line 301: should add fallback in case config is empty or null

hoangdat (approved):

Issue #2522, opened 2025-09-05
Cannot share image from gallery


tddang-linagora:
### Step to reproduce
1. Make sure Twake Chat app is terminated/killed
2. Open you phone gallery 
3. Open a photo → Share
4. Share by Twake Chat 
5. The process stops in the middle → can’t share

### Video demo
https://github.com/user-attachments/assets/55c22e5f-1c69-4e64-b78a-2f5c88d83d65

### Enviroment
Version 2.11.0

PR #2523, opened 2025-09-05
TW-2522 Fix cannot receive shared media when app terminated

tddang-linagora:
## Ticket
- #2522 

## Root cause
https://github.com/linagora/twake-on-matrix/blob/c8d25d8e482e09cb537aae23ed4d8e2b567ca70f/lib/pages/chat_list/receive_sharing_intent_mixin.dart#L58-L60

When the app is terminated and gets waken up by other apps, the sharing logic stops at these lines

## Solution
Get the initial sharing data AFTER user's already logged in.

## Resolved
- Android:

[sharing-android.webm](https://github.com/user-attachments/assets/97ca951d-6633-4ba9-8f87-aea460fd74b3)

- IOS:

https://github.com/user-attachments/assets/095a061b-b241-4866-a2fb-a9c1bab00e14

### DIFF
--- a/lib/pages/chat_list/receive_sharing_intent_mixin.dart
+++ b/lib/pages/chat_list/receive_sharing_intent_mixin.dart
@@ -122,13 +122,6 @@ mixin ReceiveSharingIntentMixin<T extends StatefulWidget> on State<T> {
       onError: print,
     );
 
-    // For sharing images coming from outside the app while the app is closed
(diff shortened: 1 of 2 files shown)


nqhhdev:
IMO: Should add demo for file: Image, .pdf, .docs, ... etc

hoangdat (approved):

nqhhdev (approved):

Issue #2525, opened 2025-09-08
Jumping to image doesn’t work


tddang-linagora:
### Reproduce steps
1. Open a chat 
2. Go to media shared within chat 
3. Open some image 
4. Click Show in chat 
5. Jumping to that message isn’t performed

### Demo

https://github.com/user-attachments/assets/9b23e2d8-562a-4223-8a45-384646f8d110

https://github.com/user-attachments/assets/288841b2-3906-4a37-8cdf-b85bdee3573f

### Env
mobile, app ver 2.11.1, matrix.stg.lin-saas.com

PR #2526, opened 2025-09-08
TW-2525 Fix jump to image

tddang-linagora:
## Ticket
- #2525 

## Resolved
- Android:

[jump-android.webm](https://github.com/user-attachments/assets/69aa55e0-9bb3-4232-9b2d-07dcab7d1309)

- IOS:

https://github.com/user-attachments/assets/3e78d860-2cce-4623-8027-1ba63ad5e226

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -97,6 +97,7 @@ class Chat extends StatefulWidget {
   final List<MatrixFile?>? shareFiles;
   final String? roomName;
   final void Function(RightColumnType)? onChangeRightColumnType;
+  final String? initialEventId;
 
   const Chat({
     super.key,
@@ -104,6 +105,7 @@ class Chat extends StatefulWidget {
     this.shareFiles,
     this.roomName,
     this.onChangeRightColumnType,
+    this.initialEventId,
   });
 
   @override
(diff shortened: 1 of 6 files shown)


lib/pages/image_viewer/media_viewer_app_bar.dart
```diff
@@ -30,19 +30,8 @@ class MediaViewerAppBarController extends State<MediaViewerAppBar>
...
-    showAppbarPreview = widget.showAppbarPreviewNotifier;
-  }
-
-  @override
-  void dispose() {
-    super.dispose();
-    showAppbarPreview?.dispose();
-  }
+  ValueNotifier<bool>? get showAppbarPreview =>
+      widget.showAppbarPreviewNotifier;
```
hoangdat, line 35: why we remove this?

lib/pages/chat/chat.dart
```diff
@@ -97,13 +97,15 @@ class Chat extends StatefulWidget {
   final List<MatrixFile?>? shareFiles;
   final String? roomName;
   final void Function(RightColumnType)? onChangeRightColumnType;
+  final String? initialEventId;
```
hoangdat: IMO, no need to have this, we can do it same with open chat from notification

lib/pages/image_viewer/media_viewer_app_bar.dart
tddang-linagora: this `ValueNotifier` was passed down from parent. The parent already disposes it, but this widget also try to dispose it again, making the Flutter framework throw exception.

lib/pages/chat/chat.dart
tddang-linagora: Removed

Issue #2530, opened 2025-09-10
Mobile: Create new voice recorder UI in input bar


nqhhdev:
- [ ]  Create new button voice recorder
- [ ]  Update the input bar: Record time, cancel button
- [ ]  Hover to Mic icon to enable send button and starting record
- [ ]  If user exit hover auto send void message
- [ ]  If user click on Mic icon display tooltip message

PR #2535, opened 2025-09-12
TW-2530: Mobile create new voice recorder UI in input bar

nqhhdev:
## Ticket
**Related issue**
- #2530 

## Pre-merge
- https://github.com/linagora/social_media_recorder/pull/1

## Resolved

https://github.com/user-attachments/assets/5d3268d0-c701-4e53-b3b5-edd7a661b049


hoangdat:
- [ ] sometime release record button, but audio still not send 
- [ ] fake event of audio

hoangdat:
- [ ] when input bar in focus, please keep it as it is if recording audio. Try to prevent keyboard disappear

lib/pages/chat/chat.dart
```diff
@@ -659,60 +658,60 @@ class ChatController extends State<Chat>
...
+        'duration': time.inMilliseconds,
+      },
+      'org.matrix.msc3245.voice': {},
+      'org.matrix.msc1767.audio': {
+        'duration': time.inMilliseconds,
+        'waveform': waveform,
+      },
+    };
+
+    final fakeImageEvent = await room?.sendFakeImagePickerFileEvent(
```
hoangdat: should use the method for other file type

lib/presentation/extensions/send_file_extension.dart
```diff
@@ -548,6 +548,7 @@ extension SendFileExtension on Room {
                     'info': {
                       ...fileInfo.metadata,
                     },
+                    if (extraContent != null) ...extraContent,
```
hoangdat, line 551: - `extraContent` in `content`? 
- what is the different between `extraContent` here and `if (extraContent != null) 'extra_content': extraContent` at L565?

lib/pages/chat/chat.dart
nqhhdev: IMO: Not better, should rename this func support for sending fake file info. WDYT

hoangdat:
- [ ] still typing in keyboard during lock recording. we should prevent it (same with signal)

lib/presentation/extensions/send_file_extension.dart
nqhhdev: Refer Matrix SDK 

https://github.com/famedly/matrix-dart-sdk/blob/28b6f9d475f38bfc37609971d6f64e39daa5f898/lib/src/room.dart#L720C1-L721C1

lib/pages/chat/chat.dart
```diff
@@ -659,60 +661,60 @@ class ChatController extends State<Chat>
...
-    if (result == null) return;
-    final audioFile = File(result.path);
-    // ignore: unused_local_variable
-    final file = MatrixAudioFile(
-      bytes: audioFile.readAsBytesSync(),
-      name: audioFile.path,
+
+    final txid = client.generateUniqueTransactionId();
+
+    room?.sendingFilePlaceholders[txid] = audioFile;
```
hoangdat, line 682: try to remove it to prevent memory leak when we finish sending file

lib/pages/chat_draft/draft_chat_input_row.dart
```diff
@@ -39,46 +49,149 @@ class DraftChatInputRow extends StatelessWidget {
...
-                    ? _buildMobileInputRow(context)
-                    : _buildWebInputRow(context),
+                ],
               ),
-              ChatInputRowSendBtn(
-                inputText: inputText,
-                onTap: onInputBarSubmitted,
-                sendingNotifier: isSendingNotifier,
+            ),
+            if (PlatformInfos.isMobile)
```
hoangdat, line 101: how about separate it in method?

lib/pages/chat/chat_input_row.dart
```diff
@@ -27,53 +29,161 @@ class ChatInputRow extends StatelessWidget {
...
+                        controller.stopRecording.call();
+                      },
+                      sendRequestFunction: (soundFile, time, waveFrom) {
+                        Logs().d(
+                          'ChatInputRowMobile:: sendRequestFunction $soundFile',
+                        );
+                        controller.stopRecording.call();
+
+                        final file = MatrixAudioFile(
+                          bytes: soundFile.readAsBytesSync(),
```
Crash--: When I read that, I'm a bit afraid about the memory usage for long audio because we'll store all the binary in memory right? Is there a way to make a stream or something? 

edit: I don't see the use of `file.bytes` elsewhere. Is it really used?

hoangdat: agree. we can use `File path` or `bytes as stream`.

tddang-linagora:
> `late SoundRecordNotifier soundRecordNotifier;` This notifier in https://github.com/linagora/social_media_recorder/blob/39eac5a9e12b53275d2f897f27585af66a093101/lib/screen/social_media_recorder.dart#L185 was not disposed, causing memory leak where `SocialMediaRecorder` is used. Consider create another PR in social_media_recorder to dispose of this notifier.

- https://github.com/linagora/social_media_recorder/pull/2

lib/utils/matrix_sdk_extensions/matrix_file_extension.dart
```diff
@@ -165,3 +165,31 @@ extension MatrixFileExtension on MatrixFile {
 
   bool isFileHaveError(double maxSize) => size > maxSize;
 }
+
+class MatrixAudioFileCustom extends MatrixFile {
```
hoangdat:
```suggestion
class TwakeAudioFile extends MatrixFile {
```

lib/pages/chat/chat.dart
```diff
@@ -659,60 +662,62 @@ class ChatController extends State<Chat>
...
+  Future<void> sendVoiceMessageAction({
+    required TwakeAudioFile audioFile,
+    required Duration time,
+    required List<int> waveform,
+  }) async {
+    final fileInfo = FileInfo(
+      audioFile.name,
+      audioFile.filePath ?? '',
+      audioFile.size,
+      readStream: audioFile.readStream,
```
hoangdat: remove it also

lib/pages/chat/chat.dart
```diff
@@ -659,60 +662,61 @@ class ChatController extends State<Chat>
...
-      barrierDismissible: false,
-      builder: (c) => const RecordingDialog(),
+  Future<void> sendVoiceMessageAction({
+    required TwakeAudioFile audioFile,
+    required Duration time,
+    required List<int> waveform,
+  }) async {
+    final fileInfo = FileInfo(
+      audioFile.name,
+      audioFile.filePath ?? '',
```
hoangdat, line 676: please check filePath here and show toast in this case: `Audio message failed to send.`

lib/pages/chat/chat.dart
```diff
@@ -659,60 +662,61 @@ class ChatController extends State<Chat>
...
-    //       content: Text(
-    //         (e as Object).toLocalizedString(context),
-    //       ),
-    //     ),
-    //   );
-    //   return null;
-    // });
+
+    if (fakeImageEvent == null) {
+      Logs().e('Failed to create fake image event for voice message');
```
hoangdat:
```suggestion
      Logs().e('Failed to create fake event for voice message');
```

hoangdat: also toast here

lib/pages/chat/chat.dart
```diff
@@ -659,60 +662,61 @@ class ChatController extends State<Chat>
...
+      txid: txid,
+      msgType: MessageTypes.Audio,
+      fakeImageEvent: fakeImageEvent,
+      inReplyTo: replyEventNotifier.value,
+      extraContent: extraContent,
+    )
+        .then((_) {
+      room?.sendingFilePlaceholders.remove(txid);
+    }).catchError((e) {
+      Logs().e('Failed to send voice message', e);
```
hoangdat, line 723: also toast here

lib/pages/chat/recording_dialog.dart
hoangdat, line 1: remove this class

pubspec.yaml
```diff
@@ -70,6 +70,11 @@ dependencies:
       url: git@github.com:linagora/emoji_mart.git
       ref: master
 
+  social_media_recorder:
+    git:
+      url: git@github.com:linagora/social_media_recorder.git
+      ref: hot-fix-call-back-after-stop-recording
```
hoangdat: merged: https://github.com/linagora/social_media_recorder/commit/2394c84d327bab78fd9cb2839cf0779c75d0ff9a please update

lib/pages/chat_draft/draft_chat_input_row.dart
```diff
@@ -39,45 +49,154 @@ class DraftChatInputRow extends StatelessWidget {
...
+                Logs().d(
+                  'ChatInputRowMobile:: sendRequestFunction $soundFile',
+                );
+                stopRecording?.call();
+
+                final file = MatrixAudioFile(
+                  bytes: soundFile.readAsBytesSync(),
+                  name: soundFile.path,
+                  filePath: soundFile.path,
+                  readStream: soundFile.openRead(),
```
hoangdat: I dont think we should use all these options

Issue #2541, opened 2025-09-19
Disable editing ability in Chat interface


tddang-linagora:

jbrnzl:
Closed by #2542

PR #2542, opened 2025-09-19
TW-2541 Apply capabilities settings to profile page

tddang-linagora:
## Ticket
- #2541 

#### Enable

https://github.com/user-attachments/assets/6262c141-b543-4d0d-991f-c4a971f830e5

#### Disable

https://github.com/user-attachments/assets/a88996af-ab02-4cb5-b2a7-0b9822c908fb

#### Enable

enable-mobile

#### Disable

disable-mobile

### DIFF
--- a/lib/di/global/get_it_initializer.dart
+++ b/lib/di/global/get_it_initializer.dart
@@ -1,6 +1,7 @@
 import 'dart:collection';
 
 import 'package:fluffychat/config/app_grid_config/app_config_loader.dart';
+import 'package:fluffychat/data/datasource/capabilities/server_capabilities_datasource.dart';
 import 'package:fluffychat/data/datasource/contact/address_book_datasource.dart';
 import 'package:fluffychat/data/datasource/contact/hive_third_party_contact_datasource.dart';
 import 'package:fluffychat/data/datasource/contact/phonebook_datasource.dart';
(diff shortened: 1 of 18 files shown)


lib/di/global/get_it_initializer.dart
```diff
@@ -500,6 +513,8 @@ class GetItInitializer {
     getIt.registerFactory<ReportContentInteractor>(
       () => ReportContentInteractor(),
     );
+
+    getIt.registerLazySingleton(() => const GetServerCapabilitiesInteractor());
```
nqhhdev: Can you explain why you need to use this?

lib/pages/settings_dashboard/settings_profile/settings_profile_capability_check.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  final Widget? child;
+  final Widget Function(Capabilities? capabilities, Widget? child) builder;
+
+  @override
+  Widget build(BuildContext context) {
+    return StreamBuilder(
+      stream: getIt.get<GetServerCapabilitiesInteractor>().execute(),
+      builder: (context, snapshot) {
+        final data = snapshot.data?.fold(id, id);
```
nqhhdev: I don't understand (id, id), could you explain or rename for easy understanding

lib/di/global/get_it_initializer.dart
hoangdat: yeah, maybe `factory` still ok

lib/pages/settings_dashboard/settings_profile/settings_profile_capability_check.dart
tddang-linagora: `.fold((left) => left, (right) => right))` can be replace with `.fold(id, id)`, meaning return the exact value, no modification.

lib/di/global/get_it_initializer.dart
tddang-linagora: Updated

lib/pages/settings_dashboard/settings_profile/settings_profile_capability_check.dart
tddang-linagora: Updated

PR #2557, opened 2025-09-25
Add app.twake.chat information in homeserver summary

hoangdat:


lib/pages/settings_dashboard/settings_profile/settings_profile_redirection_edit_button.dart
```diff
@@ -0,0 +1,61 @@
...
+        .loginHomeserverSummary?.appTwakeInformation?.commonSettingsInformation;
+    final redirectUrl = userId == null
+        ? null
+        : commonSettingsInformation?.completedApplicationUrl(userId);
+
+    if (capabilities?.canEditAvatar == false ||
+        capabilities?.canEditDisplayName == false ||
+        redirectUrl == null) {
+      return const SizedBox();
+    }
```
hoangdat:
```suggestion
    if (capabilities?.canEditAvatar == true ||
        capabilities?.canEditDisplayName == true ||
        redirectUrl == null) {
      return const SizedBox();
    }
```

lib/pages/settings_dashboard/settings_profile/settings_profile_redirection_edit_button.dart
```diff
@@ -0,0 +1,61 @@
...
+
+    return Padding(
```
hoangdat, line 37: can you add ripple for it too

tddang-linagora (approved):

Issue #2572, opened 2025-10-01
Delete account button


tddang-linagora:
Current behaviour: only exists in IOS, opens an email composer 

Expected behaviour: exists in all mobile apps, clicking on it redirects the user to common settings app

PR #2573, opened 2025-10-01
TW-2572 Redirect to common settings when delete account

tddang-linagora:
## Ticket
- #2572 

## Resolved
- Android:

[untitled.webm](https://github.com/user-attachments/assets/b09fac20-964f-4fd2-b7f5-f83e0e9079a1)

- IOS:

https://github.com/user-attachments/assets/84d2649e-0f83-4604-a745-d7e41a2d82c3

### DIFF
--- a/lib/pages/settings_dashboard/settings/settings.dart
+++ b/lib/pages/settings_dashboard/settings/settings.dart
@@ -2,6 +2,8 @@ import 'dart:async';
 
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/di/global/get_it_initializer.dart';
+import 'package:fluffychat/domain/model/extensions/common_settings/common_settings_extensions.dart';
+import 'package:fluffychat/domain/model/extensions/homeserver_summary_extensions.dart';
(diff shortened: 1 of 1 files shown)


lib/pages/settings_dashboard/settings/settings.dart
```diff
@@ -60,7 +62,7 @@ class SettingsController extends State<Settings> with ConnectPageMixin {
     SettingEnum.help,
     SettingEnum.about,
     SettingEnum.logout,
-    if (PlatformInfos.isIOS) SettingEnum.deleteAccount,
+    if (PlatformInfos.isMobile) SettingEnum.deleteAccount,
```
hoangdat, line 65: IMO, we forget to remove the old action of `deleteAccount`

lib/pages/settings_dashboard/settings/settings.dart
hoangdat, line 314: should remove this

lib/pages/settings_dashboard/settings/settings.dart
tddang-linagora: We keep it for when common settings is not found.

Issue #2583, opened 2025-10-09
Add ability to swipe media sent inside a chat


tddang-linagora:
I open a media file sent inside a chat. I can swipe left and right to view the rest of the media files sent inside a chat. We can make a very simple view for that 

### DoD

- try to create the viewer for `Media` tab in Chat details
- create a viewer can handle swipe image
- in case of no more image, try to handle load more logic too
- try to limit 3-5 images in viewer first

PR #2584, opened 2025-10-09
TW-2583 Add ability to swipe media when preview

tddang-linagora:
## Ticket
- #2583 

## Resolved

https://github.com/user-attachments/assets/79d051ad-e0dc-459e-9c67-7259a34e5bb2

### DIFF
--- a/lib/pages/media_viewer/media_viewer.dart
+++ b/lib/pages/media_viewer/media_viewer.dart
@@ -115,8 +115,7 @@ class MediaViewerController extends State<MediaViewer> {
       events.map(
         (event) async {
           try {
-            if (event.type != EventTypes.Encrypted ||
-                !event.content.containsKey('can_request_session')) {
+            if (event.type != EventTypes.Encrypted) {
               return event;
             }
 
@@ -219,6 +218,8 @@ class MediaViewerController extends State<MediaViewer> {
   }
 
   void pageChangedListener() {
+    if (currentPage.value == -1) return;
+
     if (mediaEvents[currentPage.value].messageType == MessageTypes.Video) {
       showAppBarAndPreview.value = true;
     }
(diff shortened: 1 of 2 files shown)


lib/widgets/matrix.dart
```diff
@@ -1083,11 +1083,19 @@ class MatrixState extends State<Matrix>
...
     client.syncPresence = foreground ? null : PresenceType.unavailable;
-    client.sync(setPresence: client.syncPresence);
+    trySyncClient();
     client.requestHistoryOnLimitedTimeline = !foreground;
     backgroundPush?.clearAllNotifications();
   }
 
+  Future<void> trySyncClient() async {
+    try {
+      await client.sync(setPresence: client.syncPresence);
```
hoangdat, line 1093: why we need it?

lib/pages/media_viewer/media_viewer.dart
```diff
@@ -0,0 +1,263 @@
...
+class MediaViewer extends StatefulWidget {
+  const MediaViewer({super.key, required this.event});
+
+  final Event event;
+
+  @override
+  State<MediaViewer> createState() => MediaViewerController();
+}
+
+class MediaViewerController extends State<MediaViewer> {
```
hoangdat, line 17: please check the memory usage and memory leak for this case

lib/widgets/matrix.dart
tddang-linagora: Exception is throwing, but we don't await it, so we can't catch it.

lib/pages/media_viewer/media_viewer.dart
tddang-linagora: Screenshot 2025-10-14 at 10 30 32 AM

No memory leak detected.

Issue #2593, opened 2025-10-15
Don’t send event to the chat when user changes their avatar


tddang-linagora:
2.12.0, tested on IOS, [[linagora.com](http://linagora.com/)](http://linagora.com) 

1. User changes their avatar 
2. An event is sent to a group chat “User changed their avatar” 

Expectation: there’s no event

![Image](https://github.com/user-attachments/assets/bafe51bc-8c25-4c09-b837-7e39e7779a81)

PR #2594, opened 2025-10-15
Tw 2593 Hide change avatar event

tddang-linagora:
## Ticket
- #2593

### DIFF
--- a/lib/domain/model/room/room_extension.dart
+++ b/lib/domain/model/room/room_extension.dart
@@ -202,9 +202,7 @@ extension RoomExtension on Room {
 
       if (lastEventAvailableInPreview == null ||
           lastEventAvailableInPreview.shouldHideRedactedEvent() ||
-          lastEventAvailableInPreview.shouldHideBannedEvent() ||
-          lastEventAvailableInPreview.shouldHideChangedAvatarEvent() ||
-          lastEventAvailableInPreview.shouldHideChangedDisplayNameEvent()) {
+          lastEventAvailableInPreview.shouldHideBannedEvent()) {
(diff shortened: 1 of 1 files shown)


lib/domain/model/room/room_extension.dart
```diff
@@ -205,6 +214,10 @@ extension RoomExtension on Room {
         for (final messageEvent in messageEvents) {
           if (messageEvent.shouldHideRedactedEvent()) continue;
           if (messageEvent.shouldHideBannedEvent()) continue;
+          if (messageEvent.roomMemberChangeType ==
+              RoomMemberChangeType.avatar) {
+            continue;
+          }
```
nqhhdev: separate small func

lib/domain/model/room/room_extension.dart
```diff
@@ -176,6 +176,15 @@ extension RoomExtension on Room {
...
               }
+              if (a.roomMemberChangeType == RoomMemberChangeType.avatar ||
+                  b.roomMemberChangeType == RoomMemberChangeType.avatar) {
+                if (a.roomMemberChangeType == RoomMemberChangeType.avatar) {
+                  return b;
+                }
+                if (b.roomMemberChangeType == RoomMemberChangeType.avatar) {
+                  return a;
+                }
+              }
```
nqhhdev: separate small func

hoangdat: please check with change display name too

lib/domain/model/room/room_extension.dart
hoangdat: with display name too

PR #2597, opened 2025-10-15
Pin unpin chat

thuyenh1081:
## Ticket
this ticket contains code for 2 script: "Search for messages inside a chat" , "Pin/unpin a chat"

## Resolved

https://github.com/user-attachments/assets/8f3700de-32ca-45c8-bbb3-78c8c246e0aa

https://github.com/user-attachments/assets/5f37d660-40aa-4c29-97e0-60eb47c5310a


integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,19 @@
...
+class ChatDetailScenario extends BaseScenario {
+  ChatDetailScenario(super.$);
+
+  Future<void> makeASearch(String searchText) async {
+    await ChatGroupDetailRobot($).getSearchIcon().tap();
+    await $.waitUntilVisible($(AppBar).$(TextField));
+    await $(AppBar).$(TextField).enterText(searchText);
+    // await SearchRobot($).enterSearchText(searchText);
+    await ChatGroupDetailRobot($).waitForEitherVisible($: $, first: $(TwakeListItem), second: $("No Results"), timeout: const Duration(seconds: 10));
+    await Future.delayed(const Duration(seconds: 2));
```
Crash--, line 16: You don't want to add the click on a search result and see if it display / scroll to the right message? (We know that today, it doesn't work, maybe that's why you didn't add it here?)

thuyenh1081: Thank for your suggestion @Crash--  .In the next PRs, I’ll try simulating the user’s scroll actions to figure out the item they want

hoangdat: please create another test case for searching a message in the past, then open it, then verify the highlight of the message, and the content. please do it in other PR @thuyenh1081

Issue #2076, opened 2024-10-09
[Bug] Glitch when jumping to a notification


tprudentova:
2.6.5, linagora.com, tested on Android 

1. Receive a notification  
2. Jump to it 
3. The necessary chat is opened, but it's glitching like crazy 

NB: I only got this bug once and cannot reproduce it for now. The message in the notification included a string `mention only` (in backticks) 

https://github.com/user-attachments/assets/a30cd0c9-071d-4d1a-8bf0-d2430a246a46

PR #2624, opened 2025-10-27
TW #2076 Prevent forever loop when not found event in timeline

hoangdat:


lib/pages/chat/chat.dart
```diff
@@ -1212,8 +1225,9 @@ class ChatController extends State<Chat>
   }
 
   int _getEventIndex(String eventId) {
-    final foundEvent =
-        timeline!.events.firstWhereOrNull((event) => event.eventId == eventId);
+    final foundEvent = timeline?.events
+        .where((event) => event.isVisibleInGui)
+        .firstWhereOrNull((event) => event.eventId == eventId);
```
tddang-linagora: Revert this part

tddang-linagora (approved):

Crash--:
I'm sorry @hoangdat but I don't understand this fix. How an event can not be found? 

I mean, today it seems we can encounter this bug at least by 2 different paths: 
- The first one by clicking on a push notification 
- The second one by clicking on a chat room from the chat list 

I can imagine that we can send a push notif and when the user click on the push, the message has been deleted or something like that, but I don't think this is the case regarding the bug reports we have. 

So I don't understand how the timeline can be empty in this case and how an event can not be found? Is it something related to the pagination? Me, as a user, the latest message I read on a channel is number 10, but there is a lot of discussion and now we are on number 10000 then we're lost somewhere? But it seems that we can also have that in Direct Message (1-1 chat) so I don't think this is related...

Issue #2635, opened 2025-10-29
Wrong typing status


tddang-linagora:
Sometimes (quite randomly) users are displayed as “typing” even though they stopped typing quite a while ago

### DoD

- [x] receiver: adding cooldown time for sync typing from other people
- [x] sender: send typing in every on change 10s

PR #2637, opened 2025-10-30
TW-2635 Add typing timer to receiver

tddang-linagora:
## Ticket
- #2635


tddang-linagora:
>Do we have more typing event if user still keep typing in more than 30s?

Yes the event is still sent after 30s

https://github.com/linagora/twake-on-matrix/blob/bb02a31342a1c29ed428d35e202ae9c106f3fc81/lib/pages/chat/chat.dart#L1484-L1498

hoangdat (approved):

hoangdat:
please rebase

PR #2645, opened 2025-11-03
Un pin and pin

thuyenh1081:
## Ticket
this PR include code for script 'Pin/unpin a chat'

## Resolved

https://github.com/user-attachments/assets/9a84ee49-8d6c-40df-b9c8-83b11962c930

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -4,4 +4,3 @@ android.enableJetifier=true
 android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
-org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home


integration_test/scenarios/chat_scenario.dart
```diff
@@ -481,4 +482,38 @@ class ChatScenario extends BaseScenario {
...
         'Chat list is not scrollable',);
   }
+
+  bool isPinAChat(TwakeListItemRobot takeListItem) {
+    final pin = takeListItem.getPinIcon();
+    return pin.visible;
+  }
+  
+  Future<void> pinAChat(String title) async {
+    final twakeListItem = ChatListRobot($).getChatGroupByTitle(title);
```
hoangdat, line 492: how about if this group chat not visible in chat list? need to scroll to find or sth else?

thuyenh1081: In that method, before making an action on twakeListItem, I will make sure it is visible on the screen by using below code: If it is existed but is hidden, it will be scroll automatically
await $.tester.ensureVisible(twakeListItem.root);

integration_test/scenarios/chat_scenario.dart
```diff
@@ -481,4 +482,38 @@ class ChatScenario extends BaseScenario {
...
+    if(!isPinAChat(twakeListItem))
```
hoangdat, line 495: but this maybe not have the method `ensureVisible`. 

Please test the case of multiple items in chat list, and attach demo

thuyenh1081: I have updated the attached video to help you check this case. In the video, you can see group named "groupTest" was not shown in the screen. And then it is scrolled to be clicked

Insalien: Demo looks good, is the PR good to merge?

hoangdat: still have one point need to cover: maybe the test item was not rendered in the render tree. So maybe can not find this item in test case

thuyenh1081: @hoangdat I have tried with chat list that containing 106 items and I see that:
- after finish loading, flutter can inspect all 106 items ( the last one named "nhóm 1" in the below video). 
- after open a chat, we call method  await $.pumpAndSettle(); This help to wait for finishing the loading
- I tried to run my script to tried to click the last one and the script pass
- I also updated pinAchat() to make it more resonable in the last commit

https://github.com/user-attachments/assets/24e08c66-a768-428f-b2f8-51ceb1ec700d
https://github.com/user-attachments/assets/b00f88b5-50b0-4845-b2ac-01dcdcfeabcf

Insalien: > still have one point need to cover: maybe the test item was not rendered in the render tree. So maybe can not find this item in test case

@hoangdat is there any reasonable timeout that we need to wait for the item to be rendered before making the test failed?

android/gradle.properties
```diff
@@ -3,4 +3,5 @@ android.useAndroidX=true
 android.enableJetifier=true
 android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
-android.nonFinalResIds=false
\ No newline at end of file
+android.nonFinalResIds=false
+org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home
```
hoangdat: IMO, we dont need it in CICD

thuyenh1081: my mistake in push :( I have remove it, thank you

hoangdat (approved):

PR #2648, opened 2025-11-03
Cache last events for faster load speed

tddang-linagora:


hoangdat:
> last event for the edited message (when opening app again)


https://github.com/user-attachments/assets/92ba28c9-48ca-4722-a88c-2f9d340913dd

lib/pages/chat_list/chat_list_sort_rooms.dart
```diff
@@ -0,0 +1,135 @@
...
+    super.didUpdateWidget(oldWidget);
+    _lastEventByRoomId = Map.fromEntries(
+      widget.rooms.map(
+        (room) => MapEntry(
+          room.id,
+          _lastEventByRoomId.putIfAbsent(room.id, () => null),
+        ),
+      ),
+    );
+    _roomsDiff = oldWidget.rooms.whereNot(widget.rooms.contains).toList();
```
hoangdat: why need use as global variable?

How about?
```suggestion
    final _removedRooms = oldWidget.rooms.whereNot(widget.rooms.contains).toList();
```

lib/widgets/matrix.dart
```diff
@@ -318,6 +319,7 @@ class MatrixState extends State<Matrix>
   @override
   void initState() {
     super.initState();
+    initSettings();
```
hoangdat, line 322: no need await it?

lib/widgets/matrix.dart
```diff
@@ -1100,57 +1102,67 @@ class MatrixState extends State<Matrix>
...
+          .getItemBool(
+            SettingKeys.enableRightAndLeftMessageAlignmentOnWeb,
+            AppConfig.enableRightAndLeftMessageAlignmentOnWeb,
+          )
+          .then(
+            (value) =>
+                AppConfig.enableRightAndLeftMessageAlignmentOnWeb = value,
+          ),
+    ]);
+    initSettingsCompleter.complete();
```
hoangdat: how about if this code cannot be execute?

lib/pages/chat_list/chat_list_view_builder.dart
```diff
@@ -18,31 +19,65 @@ class ChatListViewBuilder extends StatelessWidget {
...
+                },
+              );
+            },
+            childCount: sortedRooms.length,
+            findChildIndexCallback: (key) {
+              return sortedRooms
+                  .indexWhere((room) => room.id == (key as ValueKey).value);
+            },
+          ),
+          // delegate: SliverChildListDelegate(
```
hoangdat: can we remove it?

lib/widgets/matrix.dart
tddang-linagora: cannot await inside `initState`

lib/pages/chat_list/chat_list_sort_rooms.dart
tddang-linagora: Updated

lib/widgets/matrix.dart
tddang-linagora: try-catch added

lib/pages/chat_list/chat_list_view_builder.dart
tddang-linagora: Removed

PR #2658, opened 2025-11-08
View profile

thuyenh1081:
## Ticket
this ticket contains code for script 'View profile of a members in a group'

## Video
the video for happy case
https://github.com/user-attachments/assets/28f14746-4ac6-4a50-80a9-450b224b8d13

the video for faild case due to unmatch name between GroupInfo and Profile Info screen:
https://github.com/user-attachments/assets/e148ff18-3e99-4c85-9fdc-ead40a2292c0

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end scenario and test to view and verify member profiles across group chats, including search and per-member checks.
* **New Features**
  * Added test-side user-role metadata to support role-based assertions.
* **Tests / Automation**
  * Added a profile-screen test robot, streamlined member-list handling, and simplified input-field helpers for paste/interact flows.
* **Stability**
  * Improved login test resilience with retry handling and timing adjustments; removed minor unused/test clutter.


integration_test/robots/group_information_robot.dart
```diff
@@ -110,13 +110,10 @@ class GroupInformationRobot extends CoreRobot {
...
-    final matches = $(TwakeListItem).evaluate();
-    for (final element in matches) {
-      final finder = $(element.widget.runtimeType);
-      groupList.add(TwakeListItemRobot($, finder));
-    }
-    return groupList;
+    final count = $(TwakeListItem).evaluate().length;
+    return List.generate(count, (i) => TwakeListItemRobot($, $(TwakeListItem).at(i)));
+  
+
```
chibenwa: Do we need those 2 extra lines?

integration_test/robots/profile_information_robot.dart
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  PatrolFinder getAvatar() {
+    return $(ProfileInfoHeader).$(Avatar);
+  }
+
+  PatrolFinder getDisplayName() {
+    final count = $(ProfileInfoHeader).$(Text).evaluate().length;
+    if(count>2){
+      return $(ProfileInfoHeader).$(Text).at(count-2);
```
chibenwa:
```suggestion
    if (count>2) {
      return $(ProfileInfoHeader).$(Text).at(count-2);
```

We do not have a checkstyle equivalent for this project?

Cc @hoangdat

chibenwa: The `count - 2` hack relies on something inherently fragile: coupling (the fact of 2 distinct part of the code shares a secret on data organisation or each other inner working).

Given so, random changes done in software part A might impact in an impredictable way software part B

Can't we find a better way to locate the Display Name?

integration_test/robots/profile_information_robot.dart
```diff
@@ -0,0 +1,68 @@
...
+  PatrolFinder getOnlineStatus() {
+    return $(ProfileInfoHeader).$(Text).at(1);
+  }
+
+  PatrolFinder getMatrixAddress() {
+    return $(ProfileInfoContactRows).$(Text).at(1);
+  }
+
+  PatrolFinder getMatrixAddressCopyIcon() {
+    const copyAddressData = IconData(0xE190, fontFamily: 'MaterialIcons');
```
chibenwa: Where are the `0xXXX` constants coming from?

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,95 @@
...
+import '../robots/twake_list_item_robot.dart';
+
+class ChatDetailScenario extends BaseScenario {
+  ChatDetailScenario(super.$);
+
+  Future<void> makeASearch(String searchText) async {
+    await ChatGroupDetailRobot($).getSearchIcon().tap();
+    await $.waitUntilVisible($(AppBar).$(TextField));
+    await $(AppBar).$(TextField).enterText(searchText);
+    // await SearchRobot($).enterSearchText(searchText);
```
chibenwa: Commented code is needed?

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,95 @@
...
+
+  UserLevel parseUserLevel(String? s) {
+    final v = (s ?? '').trim().toLowerCase();
+    final head = v.split(RegExp(r'\s+|\(')).first;
+
+    return UserLevel.values.firstWhere(
+      (e) => e.name.toLowerCase() == head,
+      orElse: () => UserLevel.member,
+    );
+  }
```
chibenwa: I tend to say with my teams: no complex logic in tests, let's focus on descriptive tests.

My thought is that having this piece of code here, we are not really following this principle.

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,95 @@
...
+      final matrixAddress = row.$(Text).last.text ?? "";
+      const currentAccount  = String.fromEnvironment('CurrentAccount');
+      if(matrixAddress == currentAccount)
+      {
+        await verifyProfileInfoOfAMember(s, displayName, matrixAddress, isCurrentUser: true, level: parseUserLevel(owner));
+      }
+      else
+      {
+        await verifyProfileInfoOfAMember(s, displayName, matrixAddress, isCurrentUser: false, level: parseUserLevel(owner));
+      }
```
chibenwa:
```suggestion
      await verifyProfileInfoOfAMember(s, displayName, matrixAddress, isCurrentUser: matrixAddress == currentAccount, level: parseUserLevel(owner));
```

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,95 @@
...
+
+  Future<void> verifyProfileInfo(SoftAssertHelper s, String displayName, String matrixAdress, {bool isCurrentUser = true,
+    UserLevel level = UserLevel.member,}) async {
+    
+    s.softAssertEquals(ProfileInformationRobot($).getAvatar().exists, true, "Avatar is not shown");
+
+    // final expectedName = matrixAdress.substring(matrixAdress.indexOf("@")+1, matrixAdress.indexOf(":"));
+    final actualName = ProfileInformationRobot($).getDisplayName().text;
+    s.softAssertEquals( actualName == displayName
+    , true, "displayName is not correct, expected name is $displayName while actual name is $actualName",);
```
chibenwa:
```suggestion
    s.softAssertEquals( actualName == displayName, 
      true, "displayName is not correct, expected name is $displayName while actual name is $actualName",);
```

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,95 @@
...
+        
+    try{
+      s.softAssertEquals(ProfileInformationRobot($).getOnlineStatus().exists, true, "online status is not shown");
+    }
+    catch(_){}
```
chibenwa: Empty catch blocks are allways fishy...

integration_test/tests/chat/chat_group_test.dart
```diff
@@ -179,4 +180,25 @@ void main() {
...
+
+      //open group Info
+      await ChatScenario($).openGroupChatInfo();
+
+      //Verify profile Info of all member
+      await ChatDetailScenario($).verifyProfileInfoOfAllMember(s);
+
+      s.verifyAll();
+    },
+  );
```
chibenwa:
```suggestion
  TestBase().runPatrolTest(
    description: 'View profile of a members in a group',
    test: ($) async {
      final s = SoftAssertHelper();
      await HomeRobot($).gotoChatListScreen();

      const groupTest = String.fromEnvironment('GroupTest');
      await ChatScenario($).openChatGroupByTitle(groupTest);

      await ChatScenario($).openGroupChatInfo();

      await ChatDetailScenario($).verifyProfileInfoOfAllMember(s);

      s.verifyAll();
    });
```

 -> 1/ Method names are exactly the comments, modulo camel case, the comments are thus redundant and can be stripped away. If that's unclear to you please take time to read `Clean Code: A Handbook of Agile Software Craftsmanship` 08/2008 from Robert Martin, it should come at a revelation...

 -> 2/ I hate lone closing brackets on lines and are in war with them :-)))

integration_test/tests/chat/chat_group_test.dart
```diff
@@ -179,4 +180,25 @@ void main() {
 
     },
   );
+
+  TestBase().runPatrolTest(
```
chibenwa: We wrote one scenario: that's great. 

Can we think of alternative scenarii that can complement this one on the same topic?

integration_test/robots/profile_information_robot.dart
thuyenh1081: When I inspect that icon, I see a value that represents the type of icon — at least, that’s what I think. I use it to avoid depending on the language.
In the attached image, you can see it’s U+E190, but when putting that value into the code, I have to convert it to 0xE190 (a hexadecimal number) so that Dart can work with it
Screenshot 2025-11-10 at 10 15 16

integration_test/scenarios/chat_detail_scenario.dart
thuyenh1081: will be removed, I forgot to remove the old way

integration_test/tests/chat/chat_group_test.dart
thuyenh1081: Yes, in this case the method name is clear and matches my comments, so I’ll remove the comment.
But in cases where multiple lines are required to perform one behavior, I would like to add a comment to make it easier for the reader to understand

integration_test/tests/chat/chat_group_test.dart
thuyenh1081: Yes, I’ll add more scenarios.
For now this is just a basic one to help me finish automating the main actions on this screen for this feature. Once that’s done, I’ll extend it with additional complementary scenarios.

integration_test/scenarios/chat_detail_scenario.dart
thuyenh1081: Yeah, it does look ugly, but in reality there are some users whose online status isn’t displayed at all — the element simply doesn’t exist
So if the online status isn’t shown, I don’t want the test to fail at that point, that’s why I use a catch here.
I haven’t thought of a better way to handle it yet :(
Screenshot 2025-11-10 at 10 44 16

integration_test/scenarios/chat_detail_scenario.dart
thuyenh1081: Yes, I’ll change it to something simpler

integration_test/robots/profile_information_robot.dart
thuyenh1081: Yes, I would like to explain for this and previous your comment:
- The profile header has several Text elements. The last one is for the online status, and the one above it is the display name. Sometimes we don’t have the online status element, so the display name becomes the last one. I don’t see any attribute that can distinguish between these two, so I have to use this approach, even though it does look a bit ugly.
Screenshot 2025-11-10 at 10 44 16
Screenshot 2025-11-10 at 10 13 35

integration_test/robots/group_information_robot.dart
thuyenh1081: I will remove it, I didn't notice it

integration_test/robots/login_robot.dart
```diff
@@ -228,6 +233,8 @@ class LoginRobot extends CoreRobot {
         getSignInBtn(),
```
hoangdat, line 233: What happens if we click on signIn button success? How we can break the while loop?

integration_test/robots/profile_information_robot.dart
```diff
@@ -0,0 +1,70 @@
...
+
+  PatrolFinder getAvatar() {
+    return $(ProfileInfoHeader).$(Avatar);
+  }
+
+  PatrolFinder getDisplayName() {
+    // Display name is second-to-last when online status exists, otherwise first
+    final count = $(ProfileInfoHeader).$(Text).evaluate().length;
+    if (count > 2) {
+      return $(ProfileInfoHeader).$(Text).at(count - 2);
```
hoangdat: quite magic for this. how about to use another way to detect this element? put Id?

integration_test/robots/profile_information_robot.dart
```diff
@@ -0,0 +1,70 @@
...
+        .$(Icon)
+        .containing(find.byIcon(_copyAddressIcon));
+  }
+
+  PatrolFinder getSentMessageBtn() {
+    return $(InkWell).containing(find.byIcon(_sentMsgIcon));
+  }
+
+  PatrolFinder getRemoveFromBtn() {
+    return $(InkWell).containing(find.byIcon(_removeAccountIcon));
```
hoangdat, line 49: we can add `id` for this to find element instead of finding by iconData.

Cc @tddang-linagora

integration_test/scenarios/chat_detail_scenario.dart
```diff
@@ -0,0 +1,142 @@
...
+    await ChatGroupDetailRobot($).waitForEitherVisible(
+      $: $,
+      first: $(TwakeListItem),
+      second: $("No Results"),
+      timeout: const Duration(seconds: 10),
+    );
+    await Future.delayed(const Duration(seconds: 2));
+  }
+
+  UserLevel parseUserLevel(String? userlevel) {
```
hoangdat, line 33: Quite magic, and it depends much on language.

thuyenh1081: - I’m trying to find another way that doesn’t depend on the app’s display text, but currently there’s no good option. You can see the image below. So, I must use this way. However, user just pass the value the same ones that declared here: 
Screenshot 2025-12-25 at 16 30 34
- If we can add an attribute to identify the member type, that would be great.
- If you don’t mind, I’ll add a TODO to change the way to delare type of member, when such an attribute is available, so it won’t depend on the app language.
Screenshot 2025-12-25 at 14 34 35

integration_test/robots/profile_information_robot.dart
thuyenh1081: If there were an ID or any attribute to help me identify the display name on the Profile Info page, I would use it. Currently, I haven’t found any other way
Screenshot 2025-12-25 at 16 17 38
Screenshot 2025-12-25 at 16 09 33

integration_test/robots/login_robot.dart
thuyenh1081: The while loop only runs while the CAPTCHA verification dialog is visible (i.e. the OK button exists).
Once clicking Sign in succeeds, the CAPTCHA dialog disappears or the flow navigates forward, so existsOptionalNativeItems(...) returns false and the loop exits automatically.
Additionally, maxRetries is used as a safety guard to prevent an infinite loop if the dialog keeps reappearing.

integration_test/scenarios/chat_detail_scenario.dart
thuyenh1081: fixed, not using parseUserLevel

integration_test/robots/profile_information_robot.dart
thuyenh1081: I think #0xEFA9 is quite unique

integration_test/robots/profile_information_robot.dart
thuyenh1081: fixed, I change other way to verrify the display name

Issue #2665, opened 2025-11-10
If you select a message, you jump to the bottom of the chat


tddang-linagora:
## Reproduction steps

1. Open a chat 
2. Scroll up 
3. Select some message 
4. You jump to the very bottom of the chat, need to scroll back up 

## Demo / Video / Screenshots

https://github.com/user-attachments/assets/f9fa3cab-15ac-4353-8429-02550a0a54cb

jbrnzl:
Fixed by #2666

PR #2666, opened 2025-11-10
TW-2665 Fix select message jump to bottom

tddang-linagora:
## Ticket
- #2665 

## Resolved

https://github.com/user-attachments/assets/3d321ee1-5121-46c6-a329-2237b1d76a38

### DIFF
--- a/lib/pages/chat/optional_selection_area.dart
+++ b/lib/pages/chat/optional_selection_area.dart
@@ -1,6 +1,6 @@
 import 'package:flutter/material.dart';
 
-class OptionalSelectionArea extends StatelessWidget {
+class OptionalSelectionArea extends StatefulWidget {
   final Widget child;
   final bool isEnabled;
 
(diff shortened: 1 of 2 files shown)


lib/pages/chat/optional_selection_area.dart
```diff
@@ -10,8 +10,18 @@ class OptionalSelectionArea extends StatelessWidget {
...
+class _OptionalSelectionAreaState extends State<OptionalSelectionArea> {
+  final _bucket = PageStorageBucket();
   @override
   Widget build(BuildContext context) {
-    return isEnabled ? SelectionArea(child: child) : child;
+    return PageStorage(
+      bucket: _bucket,
+      child:
+          widget.isEnabled ? SelectionArea(child: widget.child) : widget.child,
+    );
```
hoangdat, line 25: can you explain about this solution, any impact to the list ?

tddang-linagora: PageStorage to save scroll position when widget rebuilds

PR #2675, opened 2025-11-13
TW-2674 Update message status when sent with error

tddang-linagora:
## Ticket
- #2674 

## Root cause
https://github.com/linagora/twake-on-matrix/blob/c12283c9de15b2975fadd3ed751cc64c6e525b84/lib/pages/chat/seen_by_row.dart#L50-L52
When sending message with error, `seenByUsers.isEmpty` is `true`

## Solution
https://github.com/linagora/twake-on-matrix/blob/c12283c9de15b2975fadd3ed751cc64c6e525b84/lib/pages/chat/seen_by_row.dart#L46-L48
Put this condition higher

## Resolved
Simulator Screenshot - iPhone 16 Plus - 2025-11-13 at 11 14 33

### DIFF
--- a/lib/pages/chat/seen_by_row.dart
+++ b/lib/pages/chat/seen_by_row.dart
@@ -39,12 +39,12 @@ class SeenByRow extends StatelessWidget {
     List<User> seenByUsers, {
     EventStatus? eventStatus,
   }) {
-    if (eventStatus == null || eventStatus == EventStatus.sending) {
-      return MessageStatus.sending;
-    }
-
     if (eventStatus == EventStatus.error) {
       return MessageStatus.error;
+    }
+
+    if (eventStatus == null || eventStatus == EventStatus.sending) {
+      return MessageStatus.sending;
     }
 
     if (eventStatus == EventStatus.sent || seenByUsers.isEmpty) {


lib/pages/chat/seen_by_row.dart
```diff
@@ -43,14 +43,14 @@ class SeenByRow extends StatelessWidget {
       return MessageStatus.sending;
     }
 
-    if (eventStatus == EventStatus.sent || seenByUsers.isEmpty) {
-      return MessageStatus.sent;
-    }
-
     if (eventStatus == EventStatus.error) {
       return MessageStatus.error;
     }
```
hoangdat, line 52: how about to keep the error at the top

lib/pages/chat/seen_by_row.dart
```diff
@@ -43,14 +43,14 @@ class SeenByRow extends StatelessWidget {
...
 
+    if (eventStatus == EventStatus.sent || seenByUsers.isEmpty) {
```
hoangdat, line 50: why `seenByUsers.isEmpty` should have status = sent?

lib/pages/chat/seen_by_row.dart
tddang-linagora: Done

hoangdat (approved):

PR #2689, opened 2025-11-17
Handle remove event exception

tddang-linagora:


lib/utils/matrix_sdk_extensions/event_list_extension.dart
```diff
@@ -0,0 +1,215 @@
+import 'package:fluffychat/utils/matrix_sdk_extensions/event_extension.dart';
+import 'package:fluffychat/utils/matrix_sdk_extensions/filtered_timeline_extension.dart';
+import 'package:matrix/matrix.dart';
+
+/// Result of syncing event lists between old and new states.
+class EventListSyncResult {
```
nqhhdev: Should to add `EquatableMixin` ?

tddang-linagora: Done

docs/adr/0031-event-list-synchronization-in-chat-scroll-view.md
```diff
@@ -0,0 +1,133 @@
...
+7. **Reusability**: Event synchronization logic can be used in other widgets
+
+### Negative
+
+1. Memory overhead from maintaining event maps (minimal impact)
+2. Additional layer of abstraction to understand
+
+### Known Considerations
+
+- **Matrix SDK Event Loops**: The Matrix SDK can sometimes add/remove the same event in rapid succession, creating a loop. The build-time safety check ensures such events are properly hidden when removed.
```
Crash--, line 122: Is it still the case (duplicate event) with the updated matrix sdk? 

If yes then, maybe we should fix it upstream no? Is it really at the app level we should fix this kind of issues ?

tddang-linagora: Not sure it happens at upstream.

nqhhdev (approved):

Issue #2693, opened 2025-11-18
Make it possible to execute individual Patrol test


Insalien:
### Describe the feature

Right now we have to execute all the test cases in a single *_test.dart file using --target during development.
The idea is to be able to execute individual test case using --tags by adding unique tag for each test case.

Here is a patch to add a tag in 1 single test:
[test_tagging.patch](https://github.com/user-attachments/files/23597373/test_tagging.patch)

And we can execute the single test with the following command:
`patrol run --tags 'chat_group_test_test01'`

### Describe the solution you'd like or your expectation

Have an individual test for each tag with the form "$filename_testXX', ex: chat_group_test_test01, chat_list_test_test02

PR #2701, opened 2025-11-19
Tw 2693 make it possible execute an individualtest

thuyenh1081:
## Ticket
https://github.com/linagora/twake-on-matrix/issues/2693

## Root cause
Can not run a specific script

## Solution
Add tags parameter into the method runPatrolTest

### DIFF
--- a/integration_test/tests/chat/chat_group_test.dart
+++ b/integration_test/tests/chat/chat_group_test.dart
@@ -41,7 +41,7 @@ Future<(String, String)> prepareTwoMessages(PatrolIntegrationTester $) async {
 
 void main() {
   TestBase().runPatrolTest(
-    tags: "chat_group_test_test01",
+    tags: ["chat_group_test_test01"],
     description: 'verify the display of pull down menu in a direct chat',
     test: ($) async {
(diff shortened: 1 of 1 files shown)


integration_test/base/test_base.dart
```diff
@@ -9,7 +9,7 @@ class TestBase {
   void runPatrolTest({
     required String description,
     required Function(PatrolIntegrationTester $) test,
-    NativeAutomatorConfig? nativeAutomatorConfig,
+    NativeAutomatorConfig? nativeAutomatorConfig, dynamic tags = const [],
```
Insalien: make a line return here

integration_test/tests/chat/chat_group_test.dart
```diff
@@ -41,6 +41,7 @@ Future<(String, String)> prepareTwoMessages(PatrolIntegrationTester $) async {
 
 void main() {
   TestBase().runPatrolTest(
+    tags: "chat_group_test_test01",
```
Insalien: it should be a list

Insalien (approved):
Nothing to add for me

PR #2707, opened 2025-11-20
TW-2706 Add mobile browser promotion banner

tddang-linagora:
## Ticket
- #2706 

#### Installed

[android-installed.webm](https://github.com/user-attachments/assets/9abdd5a7-c18a-44f5-be33-e3a1621314d1)

#### Not installed

[android-uninstalled.webm](https://github.com/user-attachments/assets/58a58523-ba4c-4654-b068-720753f5a6dd)

#### Installed

https://github.com/user-attachments/assets/fc9cadd3-aeaa-4b51-a301-f41815c694df

#### Not installed
Only works when universal link is implemented


Crash--:
Why don't we use native smart banner? (https://developer.apple.com/documentation/webkit/promoting-apps-with-smart-app-banners /

hoangdat:
> Why don't we use native smart banner? (https://developer.apple.com/documentation/webkit/promoting-apps-with-smart-app-banners / 

The reason:
- We have customized behavior: always display banner, so we choose to manage it ourself
- Android still not support native banner
- And we will have universal link, I think we should improve banner also

Issue #2700, opened 2025-11-19
[Test] Patrol test continues to run even if the previous step fails


Insalien:
Here is an example when I run a test 
<img width="1121" height="321" alt="Image" src="https://github.com/user-attachments/assets/274f2799-a3ed-4d9c-bfe3-3bc46140d8c8" />

We expect the step to stop when the previous step fails

PR #2708, opened 2025-11-20
Tw 2700 [Test] avoid exception for optional items

thuyenh1081:
## Ticket
[**Related issue**](https://github.com/linagora/twake-on-matrix/issues/2700)

## Root cause
Use the waitUntilVisible function for cases where it’s acceptable that the element might not appear, while waitUntilVisible will throw an exception if the element is not found within the timeout. This explains why some steps are marked as failed but the following steps can still continue to run.

## Solution
To solve this issue, replace waitUntilVisible with another function that simply waits and does not throw an exception if the element is not found.

## Impact description
In this PR, I updated the login function, because the login screen only appears when the user has not logged in successfully before.

## evidenta for 2 cases: one with login screen and one without
The goal of these PRs is to handle several optional elements that may appear in certain situations but not in others. These include the login page, the share information dialog, the browser login prompt, and the CAPTCHA verification dialog.
In addition to the code changes, I attached videos of two test cases to demonstrate that all optional items are handled correctly and no steps fail.

https://github.com/user-attachments/assets/aab6bfc0-822c-4f07-a27c-30f2c3d8a4d1

https://github.com/user-attachments/assets/b7302fd8-2a9d-4031-9138-c44b61cad5c3

https://github.com/user-attachments/assets/10ffe0e7-dbca-4c30-b838-14a6fc6750a9

## Summary by CodeRabbit

* **Tests**
  * Improved login-flow tests with platform-aware selectors, native-view synchronization utilities, more robust enter/tap sequences, optional captcha handling, and renamed readiness checks; removed an obsolete test helper.
* **Chores**
  * Minor build configuration edit with no behavioral change.
  * Test runner script runs with increased verbosity and a simplified test target invocation.

### DIFF
--- a/integration_test/robots/login_robot.dart
+++ b/integration_test/robots/login_robot.dart
@@ -41,7 +41,7 @@ class LoginRobot extends CoreRobot {
     }
   }
 
-  Selector getOKBtnInVerifyCapchaDialog() {
+  Selector getOKBtnInVerifyCaptchaDialog() {
     if (Platform.isIOS) {
       return Selector(
         text: 'Close',
@@ -52,7 +52,7 @@ class LoginRobot extends CoreRobot {
     }
   }
 
-  Future<bool> isWelComePageVisible() async {
+  Future<bool> isWelcomePageVisible() async {
     final welcomePage = $(TwakeWelcome);
     try {
       await welcomePage.waitUntilVisible();
@@ -207,16 +207,16 @@ class LoginRobot extends CoreRobot {
       appId: getBrowserAppId(),
     );
 
-    // if "verify ...please wait for Capcha" dialog is shown, click OK for continue wait
+    // if "verify ...please wait for Captcha" dialog is shown, click OK to continue waiting
     // and click Sign in again
     if (await CoreRobot($).existsOptionalNativeItems(
       $,
-      getOKBtnInVerifyCapchaDialog(),
+      getOKBtnInVerifyCaptchaDialog(),
       appId: getBrowserAppId(),
       timeout: const Duration(seconds: 2),
     )) {
       await $.native.tap(
-        getOKBtnInVerifyCapchaDialog(),
+        getOKBtnInVerifyCaptchaDialog(),
         appId: getBrowserAppId(),
       );
       await $.native.tap(
--- a/integration_test/scenarios/login_scenario.dart
+++ b/integration_test/scenarios/login_scenario.dart
@@ -20,7 +20,7 @@ class LoginScenario extends BaseScenario {
 
   Future<void> login() async {
     final loginRobot = LoginRobot($);
-    if (await loginRobot.isWelComePageVisible()) {
+    if (await loginRobot.isWelcomePageVisible()) {
       await loginRobot.tapOnUseYourCompanyServer();
       await _handleWaitUntilVisibleHomeServerPickerView(loginRobot);
       await loginRobot.enterServerUrl(serverUrl);


integration_test/base/core_robot.dart
```diff
@@ -91,6 +91,23 @@ class CoreRobot {
     }
   }
 
+  Future<bool> existsNiceToHaveNativeItems(
```
Insalien: please use **Optional** instead of **NiceToHave**

Insalien: and it seems more as a waitUntilVisible?

integration_test/robots/login_robot.dart
```diff
@@ -57,17 +57,18 @@ class LoginRobot extends CoreRobot {
...
-    try {
-      // A tiny settle helps after navigation/animation
-      await Future<void>.delayed(timeout);
-      await $.native.enterText(
-        getLoginBtn(),
-        text: "",
-      );
-      return true; // If tap didn’t throw, it’s interactable enough
-    } catch (_) {
+  Selector getSignIn() {
```
Insalien: Do we really need 1-line method?

thuyenh1081: I’m not sure if something strange happened when I pushed to Git, but when I view this file on Git, it actually looks complete like this:
  Selector getSignIn() {
    return Selector(text: 'Sign in');
  }
 Ah, I see the method is collapsed in your picture

integration_test/robots/login_robot.dart
```diff
@@ -25,6 +25,7 @@ class LoginRobot extends CoreRobot {
   }
 
   Future<void> confirmServerUrl() async {
+    await $.waitUntilVisible($('Continue'));
```
tddang-linagora: From Patrol's documentation: https://patrol.leancode.co/documentation/finders/usage#performing-actions

>What's very cool about Patrol's tap() is that it doesn't immediately fail if the finder finds no visible widgets – instead, it waits for some time (which you can specify globally in [PatrolTesterConfig](https://pub.dev/documentation/patrol_finders/latest/patrol_finders/PatrolTesterConfig-class.html) or as argument to the tap() method) and taps on the first widget as soon as it becomes visible. This lets you get rid of fixed timeouts and test your app just like a real user would.

integration_test/robots/login_robot.dart
tddang-linagora: @thuyenh1081 No he meant this method only contains `return Selector(text: 'Sign in');`, without any additional logic. Is it better that we use `Selector(text: 'Sign in')` directly?

integration_test/base/core_robot.dart
```diff
@@ -91,6 +91,23 @@ class CoreRobot {
...
+  }) async {
+    final end = DateTime.now().add(timeout);
+    while (DateTime.now().isBefore(end)) {
+      final views = await $.native.getNativeViews(selector);
+      if (views.isNotEmpty) {
+        return true;
+      }
+      await Future<void>.delayed(interval);
+    }
+    return false;
```
tddang-linagora, line 143:
```suggestion
    await $.pumpAndTrySettle(duration: timeout);
    final views = await $.native.getNativeViews(selector);
    return views.isNotEmpty;
```

tddang-linagora (commented):
Please, I'm begging you, run this command and create another commit for it.
```console
dart format integration_test
```

integration_test/base/core_robot.dart
thuyenh1081: getSignIn() is also used in pressSignInSsoLogin() method

integration_test/base/core_robot.dart
thuyenh1081: - waitUntilVisible will throw an exception if the element is not found within the timeout
- so, I create this method to simply waits and does not throw an exception if the element is not found.

integration_test/robots/login_robot.dart
thuyenh1081: getSignIn() is also used in pressSignInSsoLogin() method

integration_test/base/core_robot.dart
thuyenh1081: It seems they don’t provide the same effectiveness. pumpAndTrySettle is useful for Flutter UI (WidgetTester), but it doesn’t help much with native views.

tddang-linagora: Please run the test with it first and see if it will fail because of these lines.

integration_test/robots/login_robot.dart
thuyenh1081: I understand that you are suggesting me using global timeout, right ? If yes, I have created a PR for setting global timeout. Thank you

integration_test/robots/login_robot.dart
```diff
@@ -25,6 +25,7 @@ class LoginRobot extends CoreRobot {
   }
 
   Future<void> confirmServerUrl() async {
```
Insalien: the method name and the code is not related at all

thuyenh1081: updated in the new commit

integration_test/base/core_robot.dart
thuyenh1081: @tddang-linagora I updated the script based on your suggestion, and it still worked fine until the new login page code was deployed (maybe it was deployed yesterday). Waiting for the login page elements to appear has become unstable. It seems like $.pumpAndTrySettle() doesn’t work well with native elements, so I’d prefer to go back to the old approach.

integration_test/robots/login_robot.dart
```diff
@@ -49,35 +56,89 @@ class LoginRobot extends CoreRobot {
...
+      return Selector(text: label);
+    } else {
+      return Selector(
+        className: 'android.widget.Button',
+        textContains: label,
+      );
+    }
+  }
+
+  Selector getErrorMgs() {
```
Insalien: This method is not used

thuyenh1081: It has been removed, thanks!

hoangdat:
can you add test video? @thuyenh1081

android/gradle.properties
```diff
@@ -4,3 +4,4 @@ android.enableJetifier=true
 android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
+org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home
```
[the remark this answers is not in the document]
hoangdat: agree, please remove it and setup locally only

Issue #2699, opened 2025-11-19
[Test] Adjust Patrol testing timeout


Insalien:
### Describe the feature

We have different timeouts defined in Patrol tests which are too long than the real usage of the app, ex: waitUntilVisible can wait up to 2 minutes.
And it can drag the global testing time longer than expected in case of multiple test failures

### Describe the solution you'd like or your expectation

We would like to have:
- reasonable timeout for Patrol test cases, ex: waitUntilVisible timeout should be closer to a real usage of the app instead of waiting 2 minutes
- having global timeouts defined in a configuration file and we can overwrite the value for others testing environments

Insalien:
Please define timeout for individual test as defined by this API
https://pub.dev/documentation/patrol/latest/patrol/patrolTest.html

PR #2709, opened 2025-11-20
TW-2699 [Test] setting for global timeout

thuyenh1081:
## Ticket
[**Related issue**](https://github.com/linagora/twake-on-matrix/issues/2699)

## Root cause
The waiting time is not controlled yet

## Solution 
I've set up the global timeout to control how long we wait for an element to appear.

## Resolved
I’ve tested the case where the global timeout is set to 2 seconds, and it waits exactly 2 seconds. If the element is not found, it throws an error
Screenshot 2025-11-20 at 17 46 04


integration_test/base/test_base.dart
```diff
@@ -12,13 +12,39 @@ class TestBase {
     NativeAutomatorConfig? nativeAutomatorConfig,
     dynamic tags = const [],
   }) {
+    const visibleTimeoutMs = int.fromEnvironment(
```
Insalien, line 20: why the defaultValue is not the same at the default values in your .env file?
It should be the most suitable values for your tests

integration_test/base/test_base.dart
```diff
@@ -12,13 +12,39 @@ class TestBase {
...
+    );
     patrolTest(description,
-        config: const PatrolTesterConfig(
-          printLogs: true,
-          visibleTimeout: Duration(minutes: 1),
-        ),
-        nativeAutomatorConfig:
-            nativeAutomatorConfig ?? const NativeAutomatorConfig(),
+        config: patrolConfig,
+        nativeAutomatorConfig: nativeAutomatorConfig ?? defaultNativeConfig,
```
Insalien, line 55: how about the global timeout of a test here:
https://github.com/linagora/twake-on-matrix/issues/2699#issuecomment-3555636284

integration_test/base/test_base.dart
thuyenh1081: If we don’t set a global timeout, the default value will be 10 seconds. I just want to use that value as the default when a global timeout is not applied. If we don’t set a value here, it will become 0
The value in the .env file can be changed based on the specific environment.

integration_test/base/test_base.dart
thuyenh1081: yes, it has been added in the new commit

integration_test/.env
```diff
@@ -1,4 +1,11 @@
+#set timeout
+GLOBAL_VISIBLE_TIMEOUT_MS=30000
+GLOBAL_EXISTS_TIMEOUT_MS=6000
+GLOBAL_SETTLE_TIMEOUT_MS=6000
+GLOBAL_NATIVE_FIND_TIMEOUT_MS=6000
+GLOBAL_TEST_TIMEOUT_MS=600000
```
Insalien: is it normal to finish a test in 10 minutes?

thuyenh1081: It usually takes under 5 minutes, so I set 10 for the maximum

Insalien: I don't know why a test should take up to 5 minutes, you should talk to others to define a reasonable test timeout.

PR #2712, opened 2025-11-21
HOTFIX Thumbnail is distorted when upload image

tddang-linagora:
## Ticket
Screenshot 2025-11-21 at 9 56 59 AM

## Resolved
Screenshot 2025-11-21 at 9 56 52 AM

### DIFF
--- a/lib/pages/chat/events/images_builder/unencrypted_image_builder_web.dart
+++ b/lib/pages/chat/events/images_builder/unencrypted_image_builder_web.dart
@@ -38,9 +38,10 @@ class UnencryptedImageWidget extends StatelessWidget {
     }
     return Image.network(
       event
-          .attachmentOrThumbnailMxcUrl(getThumbnail: isThumbnail)!
-          .getDownloadLink(event.room.client)
-          .toString(),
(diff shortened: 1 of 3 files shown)


lib/pages/chat/events/images_builder/unencrypted_image_builder_web.dart
```diff
@@ -38,9 +38,10 @@ class UnencryptedImageWidget extends StatelessWidget {
     }
     return Image.network(
       event
-          .attachmentOrThumbnailMxcUrl(getThumbnail: isThumbnail)!
-          .getDownloadLink(event.room.client)
-          .toString(),
+              .attachmentOrThumbnailMxcUrl(getThumbnail: isThumbnail)
+              ?.getDownloadLink(event.room.client)
+              .toString() ??
+          '',
```
hoangdat, line 44: what happen if it is empty string?

tddang-linagora: errorBuilder is displayed. If we keep the code as is, it will throw null exception

PR #2724, opened 2025-11-26
HOTFIX Promotion banner redirection

tddang-linagora:


Crash--:
Can't we just wait to have universal link instead of trying to do hacky solution?

tddang-linagora:
>Can't we just wait to have universal link instead of trying to do hacky solution?

@Crash-- 
Android:
https://developer.android.com/training/app-links/about
>Because App Links use HTTP URLs and association with a website, users who don't have your app installed go directly to content on your site

iOS:
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html
>If your app isn’t installed, tapping a link to your website opens your website in Safari

The redirection must happen on whoever deploy the "links.twake.app", which already is a part of the TODOs:
Screenshot 2025-11-28 at 1 17 35 PM

And to answer your previous question on why we don't use iOS Smart App Banner, since our app is inside Cozy's iframe:
Screenshot 2025-11-28 at 2 23 56 PM

Crash--:
I don't understand why UL is not the right solution:

You'll generate depending of the OS Universalinks like that:

- iOS: https://links.twake.app/....&fallback=https://appstore.com/.... 
- Android: https://link.twake.app/...&fallback=https://playstore.com/... 

And I think the behavior will be the same for both OS: 

If the user click on the UL, then OS can intercept the click because app is installed and in that case, OS opens the app. If the app is not installed, then it'll open a new tab on the Browser on links.twake.app, and then, links.twake.app will redirect to the fallback parameters. 

That's all. 

We just need to confirm that OS intercept UL (if the app is installed) even from a click within an iframe.

tddang-linagora:
@Crash-- It is now testable on stg

hoangdat (approved):

Issue #2739, opened 2025-12-01
Chat investigation


hoangdat:
### Desc
- why a lot of `/sync` call without parameter? 
- missing last event for very old chat?

PR #2740, opened 2025-12-01
TW #2739 Remove unnecessary /sync calling

hoangdat:
### Desc:
- `tryClientSync` use for trigger quickly `setPresent`, but it is too expensive to call it whenever change app app state


Crash--:
OK so before we had 0ms delay to change the presence, now we have to wait the periodic sync. It's more than ok :+1:

PR #2755, opened 2025-12-08
TW-2739: Fix duplicate /sync calls - Option A (wait for existing sync)

### LINKED ISSUE #2739: Chat investigation

guimard:
## Summary
- Fixes #2739
- **Option A**: Wait for existing sync to complete before triggering a new one
- Replaces the first `oneShotSync()` call with `onSync.stream.first.timeout()` to actually wait for an existing sync

## Problem
The code was calling `oneShotSync()` twice when `syncPending` was true:
1. First call was supposed to "wait for existing sync" but actually triggered a NEW sync
2. Second call triggered another sync

This resulted in duplicate `/sync` calls visible in DevTools.

## Solution
Replace the first `oneShotSync()` with `client.onSync.stream.first.timeout()` which properly waits for the current sync to complete.

## Test plan
- [ ] Trigger a push notification during an ongoing sync
- [ ] Verify in DevTools that only ONE `/sync` call is made
- [ ] Verify events are correctly received
- [ ] Test behavior when `syncPending = false`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background synchronization reliability with enhanced timeout handling and error recovery to ensure sync operations complete properly or fail gracefully.

### DIFF
--- a/lib/utils/background_push.dart
+++ b/lib/utils/background_push.dart
@@ -538,12 +538,15 @@ class BackgroundPush {
         Logs().v('[Push] Got new clearing push');
         var syncErrored = false;
         if (client.syncPending) {
-          Logs().v('[Push] waiting for existing sync');
-          // we need to catchError here as the Future might be in a different execution zone
-          await client.oneShotSync().catchError((e) {
+          Logs().v('[Push] waiting for existing sync to complete');
+          try {
+            await client.onSync.stream.first.timeout(
+              const Duration(seconds: 30),
+            );
+          } catch (e) {
+            Logs().v('[Push] Error waiting for sync', e);
             syncErrored = true;
-            Logs().v('[Push] Error one-shot syncing', e);
-          });
+          }
         }
         if (!syncErrored) {
           Logs().v('[Push] single oneShotSync');


hoangdat:
Hi @guimard, thanks for your contribution.

Your changes still focus on `_onClearingPush` method with `getFromServer = false`. But we always use `_onClearingPush(getFromServer: false);` with Timer.periodic 20s, So it doesn't make any duplicate `/sync` call in our production now.

guimard:
> Hi @guimard, thanks for your contribution.
> 
> Your changes still focus on `_onClearingPush` method with `getFromServer = false`. But we always use `_onClearingPush(getFromServer: false);` with Timer.periodic 20s, So it doesn't make any duplicate `/sync` call in our production now.
> 
> The fix should be removing deadcode and find any problem with push notification to improve, not only change the way of calling oneShotSync.

Hi,

it's a pure AI response to this issue with 3 options. You can safely close those that have no interest :wink:

Crash--:
Oh my bad, I checked again the video I send and in this video we don't have duplicate call. We have distinct id. My bad.

hoangdat:
@Crash-- 
Our production have duplicated call, for sure. But it is not related to these changes in this PR. 
Not yet deployed.

PR #2762, opened 2025-12-09
HOTFIX Create support chat multiple times

tddang-linagora:
## Summary by CodeRabbit

* **New Features**
  * Adds an explicit initial state for support chat creation so startup progress is observable.

* **Refactor**
  * Improves internal state and sync flow during support chat creation to avoid duplicate processing.

* **Bug Fixes**
  * Reduces race conditions and tightens null-safety when changing active client, improving stability.

* **Notice**
  * Newly created support chats are configured with encryption disabled by default.

### DIFF
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -924,9 +924,9 @@ class MatrixState extends State<Matrix>
     Logs().d(
       'Matrix::_setUpToMServicesWhenChangingActiveClient: Old twakeSupported - $twakeSupported',
     );
-    if (client == null && client?.userID == null) return;
+    if (client == null || client.userID == null) return;
     try {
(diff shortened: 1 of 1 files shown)


lib/widgets/matrix.dart
```diff
@@ -173,9 +174,12 @@ class MatrixState extends State<Matrix>
   }
 
   void _createSupportChat(Client client) {
+    Object createSupportChatState = CreateSupportChatInitial();
     supportChatRoomId = null;
     StreamSubscription? syncListener;
     syncListener = client.onSync.stream.listen((_) {
+      if (createSupportChatState is! CreateSupportChatInitial) return;
```
hoangdat, line 181: how about if state is failure or completed?

hoangdat (approved):

PR #2773, opened 2025-12-12
Enhance mention parsing to support multiple formats and prevent duplicates

nqhhdev:
## Need merge

https://github.com/linagora/matrix-dart-sdk/pull/80

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/9ca57958-9487-4448-b502-6fe9830bfe35

## Summary by CodeRabbit

* **New Features**
  * Improved mention detection using formatted message content to mark mentions more reliably.

* **Bug Fixes**
  * Mention icon, width and color now update consistently for rich/formatted HTML messages.

* **Tests**
  * Added unit tests for the new mention-detection behavior; removed outdated legacy mention extraction tests.

* **Chores**
  * Updated dependency reference to a version with improved Matrix ID validation.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1868,8 +1868,8 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: improve-matrix-ID-validation-logic
-      resolved-ref: a0e0bac40169fc02e751bcaf8668df433e6ac9bc
+      ref: "twake-supported-0.22.6"
+      resolved-ref: "9330b21237b690b358bbc09dada0f3667e559e34"
       url: "git@github.com:linagora/matrix-dart-sdk.git"
     source: git
     version: "0.22.6"
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   matrix:
     git:
       url: git@github.com:linagora/matrix-dart-sdk.git
-      ref: improve-matrix-ID-validation-logic
+      ref: twake-supported-0.22.6
 
   receive_sharing_intent:
     git:


tddang-linagora (approved):

lib/utils/string_extension.dart
```diff
@@ -60,31 +60,154 @@ extension StringCasingExtension on String {
...
+      if (mention != null) {
+        matches.add((start: match.start, text: mention));
+      }
+    }
+
+    // Match @userid:domain.com format (Matrix user IDs)
+    final matrixIdRegex = RegExp(
+      r'@[a-zA-Z0-9._=\-/]+:[a-zA-Z0-9.\-]+',
+      caseSensitive: false,
+    );
```
chibenwa: > My godness it stinks the good old redos!!!!

You’re right to be suspicious — *this pattern is absolutely vulnerable to catastrophic backtracking* and can be weaponized for a RegEx DoS (ReDoS).

Let’s break down *why*, and then I’ll give you a **safe, fast alternative**.

---

# ❗ Why your regex is vulnerable

Your pattern:

```regex
@[a-zA-Z0-9._=\-/]+:[a-zA-Z0-9.\-]+
```

The dangerous part is:

```
[a-zA-Z0-9._=\-/]+
```

It contains:

* `.` (literal)
* `_`
* `=`
* `-`
* `/`

Those are *fine*, **individually**.

The problem is: **the character class contains a dash (`-`) in the middle**, which creates ranges and can behave unexpectedly unless carefully escaped or placed at the start or end of the class.

But even more important:

### 🚨 The entire regex is built from two adjacent greedy `+` quantifiers, with nothing preventing ambiguous backtracking between them.

```
@[allowed]+ : [allowed]+
```

On malicious input like:

```
@AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA:
BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB
```

…or even worse, a string with **no colon at all** after the long run:

```
@AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

the regex engine will try *every possible split* between the two `+` segments, causing exponential backtracking.

---

# ✔️ Safe, fast version (no backtracking traps)

Matrix IDs follow the form:

```
@localpart:domain
```

Using safe classes, and no ambiguous quantifiers:

### **Option A (exact equivalent, ReDoS-safe)**

```dart
final matrixIdRegex = RegExp(
  r'@[A-Za-z0-9._=\-/]+:[A-Za-z0-9.-]+',
  caseSensitive: false,
);
```

Why this is safe:

* The `-` is placed at **end of the class**, removing ambiguity.
* There’s no possible catastrophic backtracking pattern:

  * A character class with `+` does *not* cause backtracking by itself.
  * Two such classes separated by a literal `:` are safe.

---

# ✔️ Even safer / stricter Matrix ID validation

If you want a more *spec-compliant* regex:

### Localpart

```
[A-Za-z0-9._=\-/]+
```

### Domain

Better to use a real domain pattern:

```
[A-Za-z0-9.-]+\.[A-Za-z]{2,}
```

Final:

```dart
final matrixIdRegex = RegExp(
  r'@[A-Za-z0-9._=\-/]+:[A-Za-z0-9.-]+\.[A-Za-z]{2,}',
  caseSensitive: false,
);
```

---

# 💡 Want zero-regex risk?

You can also parse Matrix IDs with a simple split:

```dart
bool isMatrixId(String s) {
  if (!s.startsWith('@')) return false;
  final parts = s.substring(1).split(':');
  if (parts.length != 2) return false;
  return parts[0].isNotEmpty && parts[1].isNotEmpty;
}
```

No backtracking, no ambiguity.

---

If you want, tell me your **exact constraints** (strictness, performance, allowed characters), and I can craft the **minimal, fully safe regex** for your use case.

chibenwa (changes requested):
ALWAYS BE CAREFUL ABOUT REDOS

https://cwe.mitre.org/data/definitions/1333.html

In java I would FUZZ ( https://github.com/CodeIntelligenceTesting/jazzer ) the hell out of anything that contains the word `parse`

hoangdat:
> Use Matrix SDK's MatrixIdExtension

It also another ReDoS vulnerability too :(

lib/utils/string_extension.dart
```diff
@@ -60,31 +61,62 @@ extension StringCasingExtension on String {
...
 
-  List<String> getMentionsFromMessage() {
-    final RegExp regex = RegExp(r"@\[([^\]]+)\]");
-    final Iterable<Match> matches = regex.allMatches(this);
+  /// Extracts all mentioned user IDs from HTML formatted body.
+  ///
+  /// Parses HTML using DOM parser to find all <a> tags with matrix.to URLs.
+  /// Uses Matrix SDK's `isValidMatrixId` for canonical validation.
+  /// Strips query parameters from matrix.to URLs to prevent malformed IDs.
+  List<String> getAllMentionedUserIdsFromMessage(Room room) {
```
hoangdat: `room` not in use?

lib/utils/string_extension.dart
```diff
@@ -60,31 +61,62 @@ extension StringCasingExtension on String {
...
+        try {
+          // Decode percent-encoded characters (e.g., %3A -> :)
+          final decodedId = Uri.decodeComponent(encodedId);
+
+          // Validate using Matrix SDK's canonical validation
+          if (decodedId.isValidMatrixId) {
+            mentionUserIds.add(decodedId);
+          }
+        } catch (e) {
+          // If decoding fails, skip this href
```
hoangdat: put the log here,

lib/utils/string_extension.dart
```diff
@@ -60,31 +61,62 @@ extension StringCasingExtension on String {
...
+          continue;
+        }
       }
+    } catch (e) {
+      // If HTML parsing fails, return empty list
```
hoangdat: also here

lib/utils/string_extension.dart
```diff
@@ -60,31 +61,62 @@ extension StringCasingExtension on String {
...
+        }
+
+        final encodedId = href.substring(idStart, idEnd);
+
+        try {
+          // Decode percent-encoded characters (e.g., %3A -> :)
+          final decodedId = Uri.decodeComponent(encodedId);
+
+          // Validate using Matrix SDK's canonical validation
+          if (decodedId.isValidMatrixId) {
```
hoangdat: we should fix SDK for removing ReDoS

lib/utils/string_extension.dart
```diff
@@ -60,31 +61,62 @@ extension StringCasingExtension on String {
...
+            mentionUserIds.add(decodedId);
+          }
```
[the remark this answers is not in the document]
nqhhdev: No need, isValidMatrixId already has that condition pls resolve conversation

lib/utils/string_extension.dart
nqhhdev: https://github.com/linagora/matrix-dart-sdk/pull/80 pls review

lib/utils/matrix_sdk_extensions/event_extension.dart
```diff
@@ -536,4 +536,19 @@ extension LocalizedBody on Event {
...
+    // Check formatted_body (HTML format with matrix.to links)
+    final formattedBody = content.tryGet<String>('formatted_body');
+    if (formattedBody != null && formattedBody.isNotEmpty) {
+      final mentionedUserIds =
+          formattedBody.getAllMentionedUserIdsFromMessage();
+      return mentionedUserIds.contains(currentUserId);
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
hoangdat: 1. Edited content still be in `content['formatted_body']`, so the functional still good.
2. What is your suggestion to optimize?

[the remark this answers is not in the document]
nqhhdev: 1. 
- when a message is edited → automatically contain the edited version.
- Edit adds/removes a mention → lastEvent should return new data 'formatted_body'

2. Updated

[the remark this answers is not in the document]
nqhhdev: @coderabbitai

nqhhdev:
@coderabbitai

PR #2784, opened 2025-12-16
[Draft] Trying to fix problem of select text 

hoangdat:


hoangdat:
handled in https://github.com/linagora/twake-on-matrix/pull/2803

Issue #2826, opened 2026-01-14
Jump to pinned message doesn’t always work


tddang-linagora:
2.20.0, staging, tested on Chrome 

1. Pin several messages 
2. Click on the pinned message section several time 
3. You jump to pinned message okay until you reach the last one, then jumping is broken 

https://github.com/user-attachments/assets/47585663-41b9-4446-98b7-5be89799bb55

PR #2827, opened 2026-01-14
TW-2826 Fix cannot jump to pinned messages

tddang-linagora:
## Ticket
- #2826 

## Root cause
Check last visible event method did not work as expected

## Resolved

https://github.com/user-attachments/assets/87d0caf2-844e-477a-a87f-787545aaecaa

## Summary by CodeRabbit

* **Refactor**
  * Improved message navigation logic to make scrolling to specific messages more reliable and efficient.

* **New Features**
  * Enhanced tracking of which message is currently visible in the chat timeline to improve read-status and visibility-related behaviors.

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1437,7 +1437,7 @@ class ChatController extends State<Chat>
     if (eventsCount == 0) return;
 
     // Find nearest rendered message to determine scroll direction
-    final nearestRenderedIndex = _findNearestRenderedMessageIndex(targetIndex);
+    final nearestRenderedIndex = _findVisibleEventIndex();
     if (nearestRenderedIndex == null) return;
 
     // Determine scroll direction and estimate distance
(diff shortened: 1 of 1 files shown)


hoangdat:
> The method now returns the index of the currently visible event via visibleEventId rather than searching for the "nearest rendered" message. The name _findNearestRenderedMessageIndex no longer accurately describes what it does.

Agree on this

lib/pages/chat/chat.dart
```diff
@@ -1514,22 +1514,13 @@ class ChatController extends State<Chat>
 
   /// Finds the index of the nearest rendered message to [targetIndex].
   int? _findNearestRenderedMessageIndex(int targetIndex) {
```
hoangdat: `targetIndex` not in use

hoangdat: method name should change to indicate the logic?

tddang-linagora: Done

tddang-linagora: Done

Issue #2834, opened 2026-01-15
Message isn’t displayed in chat after you jump to a notification


tddang-linagora:
# Reproduction steps

NB: It’s a sometimes issue, might be hard to reproduce 

1. Receive a notification 
2. Jump to that notification 
3. The message from the notification along with some other messages aren’t displayed inside the chat. You need to reload to view it

## Demo / Video / Screenshots

https://github.com/user-attachments/assets/6cf18a93-ae48-49a2-9120-e16e20864959

PR #2835, opened 2026-01-15
[DO NOT MERGE] TW-2834 Fix missing messages when open chat room by notification

tddang-linagora:
## Ticket
- #2834 

## Root cause
In the ADR

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition so notifications and deep links reliably open the correct conversation and message history.
  * Improved timeline/history loading to avoid showing stale or missing messages during room transitions.

* **Stability**
  * Improved lifecycle handling to reduce UI side effects and make chat view rendering more predictable.
  * Reset pinned-message state and cancel related subscriptions to prevent stale pinned items after room changes.

* **Documentation**
  * Added ADR documenting the notification redirection race-condition fix.

### DIFF
--- a/docs/adr/0034-fix-notification-redirection-race-condition.md
+++ b/docs/adr/0034-fix-notification-redirection-race-condition.md
@@ -1,0 +1,64 @@
+# 34. Fix Notification Redirection Race Condition
+
+Date: 2026-01-15
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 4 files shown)


nqhhdev:
This is a POC PR. I discussed it with @tddang-linagora, and he told me that we couldn't reproduce the issue, and the proposed solution doesn't cover all cases. Therefore, I think we should keep the issue open, close this PR, and look for another solution once we can reliably reproduce the problem.

FYI @pm-McFly

Issue #2881, opened 2026-02-12
Cache avatars


tddang-linagora:
Currently avatars are loaded every time you open the app. We need to cache them and only update them: 

- For avatars: only when they are changed

PR #2888, opened 2026-02-23
TW-2881 Integrate CachedNetworkImage to MxcImage

tddang-linagora:
## Ticket
- #2881 

## Disclaimer
This PR only benefits the avatars and any image that use url as source. All images that are part of an matrix event will still blink on any part of the app's re-render.


tddang-linagora:
> > All images that are part of an matrix event will still blink on any part of the app's re-render
> 
> Souds like an issue, no?

Yes that's why I don't even consider this solution when implement new cache system.

jbrnzl:
Still relevent ? @9clg6

pm-McFly:
Not relevant for now, let's see how #3062 works out first and re-open here later

PR #2892, opened 2026-03-03
fix: chat list shows message as read whereas this was not the case

9clg6:
## Ticket
No issue related

## Root cause
lastEvent.receipts includes the sender's own receipt, and the list does not filter the current user, unlike getSeenByUsers() in the conversation details (cf. RoomStatusExtension.getSeenByUsers).

Capture d’écran 2026-03-03 à 17 20 10
Capture d’écran 2026-03-03 à 17 16 15

**‼️ According to the https://spec.matrix.org/v1.13/client-server-api/#receipts, clients SHOULD NOT send read receipts for events sent by their own user. However, the matrix-dart-sdk does not enforce this rule, resulting in the sender's own receipt being stored in receiptState and exposed via lastEvent.receipts.**  

## Solution
lastEvent.receipts includes the sender's own read receipt, causing the chat list to incorrectly display the sent message as read. The fix filters out the current user from lastEvent.receipts using whereNot, aligning the chat list behavior with the conversation detail which already correctly excludes the sender via the method getSeenByUsers() in extension RoomStatusExtension (room_status_extension.dart : line 84).

## Refactoring
  As part of this fix, a few small improvements were made to ChatListItemSubtitle following Dart/Flutter best practices:
  - Matrix.of(context).client.userID replaced with room.client.userID to avoid unnecessary BuildContext usage
  - Read receipt logic extracted into RoomStatusExtension as hasLastEventBeenSeenByOthers to align with the existing pattern of getSeenByUsers()
  - Theme properties (colorScheme, textTheme) grouped at the top of build() to avoid repeated Theme.of(context) calls
  - Strict typing enforced where previously implicit, in line with Dart best practices
  - Shorthand property syntax applied where applicable
  - Container() replaced with const SizedBox.shrink() for semantic clarity and performance

## Test recommendations
Before fix : To test this, you can take account A and send a message to account B. Account B should not read the incoming message, leaving it unopened/unread. Then, on account A, the icon will turn blue after a few seconds. 
After fix: Same test, but the icon will not turn blue.

## Resolved
- Web:
- Android:
- IOS:
 
Capture d’écran 2026-03-03 à 18 17 18

## Summary by CodeRabbit

* **New Features**
  * Read-receipt indicator now reflects when other users have seen the latest message.

* **Bug Fixes**
  * Fixed read-receipt logic to exclude the current user from the "seen by" count.

* **UI**
  * Improved theming, colors, padding, and text styles across message list, mention bubbles, badges, and timestamps; seen-by indicators updated for consistency.

* **Refactor**
  * Simplified seen-by/status rendering to centralize color and style decisions.

* **Tests**
  * Added tests covering multiple seen-by scenarios to verify read-receipt behavior.

### DIFF
--- a/lib/pages/chat/events/message_time_style.dart
+++ b/lib/pages/chat/events/message_time_style.dart
@@ -1,33 +1,32 @@
 import 'package:flutter/material.dart';
 import 'package:linagora_design_flutter/linagora_design_flutter.dart';
 
-class MessageTimeStyle {
+final class MessageTimeStyle {
   static Color? timelineColor(bool timelineOverlayMessage) =>
       timelineOverlayMessage
       ? Colors.white
       : LinagoraRefColors.material().neutral[50];
 
   static double get timelineLetterSpacing => 0.4;
-
   static double get paddingTimeAndIcon => 8;
   static double get seenByRowIconSize => 16;
-  static Color seenByRowIconPrimaryColor(
+
+  static Color readReceiptColor(bool seenByOthers) => seenByOthers
+      ? LinagoraSysColors.material().secondary
+      : LinagoraRefColors.material().neutral[50]!;
+
+  static Color seenByRowIconPrimaryColor(bool timelineOverlayMessage) =>
+      timelineOverlayMessage ? Colors.white : readReceiptColor(true);
+
+  static Color seenByRowIconSecondaryColor(
     bool timelineOverlayMessage,
-    BuildContext context,
-  ) => timelineOverlayMessage
-      ? Colors.white
-      : LinagoraSysColors.material().secondary;
-  static Color? seenByRowIconSecondaryColor(
-    bool timelineOverlayMessage,
-    BuildContext context,
-  ) => timelineOverlayMessage
-      ? Theme.of(context).colorScheme.onPrimary
-      : LinagoraRefColors.material().neutral[50];
+    ColorScheme colorScheme,
+  ) => timelineOverlayMessage ? colorScheme.onPrimary : readReceiptColor(false);
 
   static TextStyle? textStyle(
-    BuildContext context,
     bool timelineOverlayMessage,
-  ) => Theme.of(context).textTheme.bodySmall?.merge(
+    TextTheme textTheme,
+  ) => textTheme.bodySmall?.merge(
     TextStyle(color: timelineColor(timelineOverlayMessage), letterSpacing: 0.4),
   );
 }
--- a/lib/pages/chat_list/chat_list_item_subtitle.dart
+++ b/lib/pages/chat_list/chat_list_item_subtitle.dart
@@ -2,17 +2,15 @@ import 'dart:async';
 
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/config/themes.dart';
+import 'package:fluffychat/generated/l10n/app_localizations.dart';
+import 'package:fluffychat/pages/chat/events/message_time_style.dart';
 import 'package:fluffychat/pages/chat/typing_timer_wrapper.dart';
-import 'package:fluffychat/presentation/mixins/chat_list_item_mixin.dart';
 import 'package:fluffychat/pages/chat_list/chat_list_item_style.dart';
+import 'package:fluffychat/presentation/mixins/chat_list_item_mixin.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/event_extension.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/matrix_locals.dart';
 import 'package:fluffychat/utils/room_status_extension.dart';
-import 'package:fluffychat/widgets/matrix.dart';
 import 'package:flutter/material.dart';
-import 'package:fluffychat/generated/l10n/app_localizations.dart';
-import 'package:linagora_design_flutter/colors/linagora_ref_colors.dart';
-import 'package:linagora_design_flutter/colors/linagora_sys_colors.dart';
 import 'package:matrix/matrix.dart';
 
 class ChatListItemSubtitle extends StatelessWidget with ChatListItemMixin {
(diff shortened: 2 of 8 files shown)


hoangdat:
hi @9clg6 , Nice to see your first PR. 

Please dont delete comment of rabbit. Moreover, we should address them all, fix until it is no more comment from it.

lib/utils/room_status_extension.dart
```diff
@@ -80,11 +80,24 @@ extension RoomStatusExtension on Room {
...
       lastReceipts.removeWhere((user) => user.id == client.userID);
     }
     return lastReceipts.toList();
   }
 
+  /// Returns true if the last event has been seen by at least one other user.
+  ///
+  /// [receipts] is a getter that reads from [room.receiptState], so it is always
+  /// up to date without requiring the [Timeline] to be loaded.
+  bool get hasLastEventBeenSeenByOthers {
```
hoangdat, line 94: can we apply the same method for chat?

nqhhdev, line 94: Should create a unit test for this method.

9clg6: If by "chat" you mean precisly "message_time.dart"/"seen_by_row.dart", no we can't : because MessageTime is rendered per message and always needs a specific eventId. hasLastEventBeenSeenByOthers only works for lastEvent, so it cannot be applied here.

event.receipts only contains receipts whose position is exactly that event. If the recipient read a more recent message, their receipt wouldn't appear on the checked event — the icon would incorrectly stay grey. This is why getSeenByUsers uses isEventIdOlderOrSameAs — it checks if the user's last known receipt is on this event or any newer one, as per the https://spec.matrix.org/v1.13/client-server-api/#receipts:
                                                                                                                                                                                                                                          "A receipt is a marker that a user has read up to a given event."

9clg6: Before creating Unit Tests i'm waiting @pm-McFly response about the server-behavior, because as I said, in documentation it's written that the sender shouldn't send a receipt on his own message. Then this fix could be unecessary regarding to this. 
If the current behavior is OK i'll create tests

pm-McFly: `hasLastEventBeenSeenByOthers` sounds like a good fix, thx!

To clarify the situation spec wise. It is written:

> a client **should not** send read receipts for events authored by its own user.

Understand, if `$event.author.mxid == $self.connected_users.current.mxid` then do not send a read receipt for that `$event`.

On the server side, the spec is UB (Undefined Behavior), so its at the discretion of the devs of that said server to do as they deem is logical to do.
In our case with Synapse Community, it seems that it's only transferring the events, regardless of the receiver and the emitter. Filtering is then on the client side.

Note: it is one of the possibility, another server might do the filtering or even bounce the read receipt emitting!

But anyhow, it means two things:

1. Why is our client sending read receipts for "own" events? -> I suspect for the "mark as unread" feature?...
2. Why weren't we ignoring those UB read receipts before? -> sounds like fixed now (no justification expected 😉)

If we send own read receipts for a reason, we better check with @tprudentova if that is a feature to keep, as this might be broken depending on the backend (implem. or even version update, _cf. UB and developer discretion above_).
And depending we will check what to prompt to the user in the case of this own read receipt not being usable.

As for what to do now:

- [ ] fix the filtering logic client side: ok -> ensure we are not polluted by clients not respecting the **should not** (including ourselves) [sounds done, i'll let @9clg6 and @hoangdat validate that checkbox once tested/reviewed]
- [ ] investigate/fix the own read receipts sending
- [ ] use same logic for read ticks in both chat list and chat view

9clg6: >  investigate/fix the own read receipts sending

The self-read receipt is functionally useless but harmless. At first glance removing it is possible without impact, but it's not worth the effort of adding code to filter it. However, I discussed this topic with Taya and she confirmed that starting in April, she would like us to develop this feature: "like on Messenger, having the members' heads line up with the last message they saw. This would allow us to have our head on OUR message in case we haven't read the following messages yet." So, given this decision, it is preferable to leave the self-read receipt as it is, since we already filter the display.

> use same logic for read ticks in both chat list and chat view

You're right that they should ideally be consistent, but the chat list and chat view operate in fundamentally different contexts: 
  - Chat list: Only has access to room.lastEvent from the room summary — no timeline is loaded.                                                                                                                                            
  - Chat view: Has the full timeline loaded in memory with real-time read receipt sync.

lib/pages/chat_list/chat_list_item_subtitle.dart
```diff
@@ -59,50 +63,44 @@ class ChatListItemSubtitle extends StatelessWidget with ChatListItemMixin {
...
               return const SizedBox.shrink();
             }
             final isMentioned = lastEvent.isMention == true;
-            return lastEvent.senderId == Matrix.of(context).client.userID
+            return lastEvent.senderId == room.client.userID
                 ? Icon(
                     Icons.done_all,
-                    color: lastEvent.receipts.isEmpty
+                    color: room.hasLastEventBeenSeenByOthers
                         ? LinagoraRefColors.material().tertiary[30]
```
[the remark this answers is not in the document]
9clg6: The blue check state is actually correct in practice. room.lastEvent is always equal to or newer than the filtered this.lastEvent in the timeline, and if others have seen room.lastEvent, they've necessarily seen this.lastEvent too. The worst case is a briefly conservative grey check that turns blue moments later. No false positive is possible.

lib/pages/chat_list/chat_list_item_subtitle.dart
```diff
@@ -59,50 +63,44 @@ class ChatListItemSubtitle extends StatelessWidget with ChatListItemMixin {
...
                         : LinagoraSysColors.material().secondary,
```
hoangdat: IMO, it should be.
```suggestion
                    color: room.hasLastEventBeenSeenByOthers
                        ? LinagoraSysColors.material().secondary,
                        : LinagoraRefColors.material().tertiary[30]
```

when someone read our message, we should show the double check icon in secondary?

lib/pages/chat_list/chat_list_item_subtitle.dart
hoangdat: more over, in message, sent icon was `LinagoraRefColors.material().neutral[50]`. Can we use the same?

lib/pages/chat_list/chat_list_item_subtitle.dart
9clg6: Yes you are right, I inversed the condition

tddang-linagora (approved):

nqhhdev (approved):

lib/pages/chat/events/message_time_style.dart
```diff
@@ -1,33 +1,36 @@
...
-  static Color seenByRowIconPrimaryColor(
-    bool timelineOverlayMessage,
-    BuildContext context,
-  ) => timelineOverlayMessage
+
+  static Color readReceiptColor(bool seenByOthers) => seenByOthers
+      ? LinagoraSysColors.material().secondary
+      : LinagoraRefColors.material().neutral[50]!;
+
+  static Color seenByRowIconPrimaryColor(bool timelineOverlayMessage) =>
```
hoangdat, line 18:
```dart
static Color seenByRowIconPrimaryColor(bool timelineOverlayMessage) =>
      timelineOverlayMessage ? Colors.white : readReceiptColor(true);
```

lib/pages/chat/events/message_time_style.dart
```diff
@@ -1,33 +1,36 @@
...
       : LinagoraSysColors.material().secondary;
+
   static Color? seenByRowIconSecondaryColor(
     bool timelineOverlayMessage,
-    BuildContext context,
+    ColorScheme colorScheme,
   ) => timelineOverlayMessage
-      ? Theme.of(context).colorScheme.onPrimary
+      ? colorScheme.onPrimary
       : LinagoraRefColors.material().neutral[50];
```
hoangdat:
```dart
static Color seenByRowIconSecondaryColor(
    bool timelineOverlayMessage,
    ColorScheme colorScheme,
  ) => timelineOverlayMessage ? colorScheme.onPrimary : readReceiptColor(false);
```

hoangdat (approved):

PR #2897, opened 2026-03-05
Fix/tw 2236 transfer ownership bricks legacy rooms

9clg6:
## Ticket
TW-2236: Transfer Ownership button bricks legacy rooms

## Root cause
Hardcoded power levels in transferOwnership() :
The old implementation used hardcoded power levels (DefaultPowerLevelMember.owner.powerLevel = 90 and DefaultPowerLevelMember.admin.powerLevel = 80) instead of the room's actual power levels. In legacy rooms where the owner had a different power level (e.g., 100), this would:                                                                            
  - Set the target user to level 90 instead of the actual owner level — they never truly became owner
  - Demote the original owner to 80 (admin) instead of the room's users_default

## Solution
1. Use dynamic room power levels instead of hardcoded values
transferOwnership() now reads the actual owner's power level (room.ownUser.powerLevel) to assign to the target user, and demotes the current owner to the room's users_default level via room.getUserDefaultLevel(). This ensures correct behavior regardless of the room's power level configuration.

2. Fix getUserDefaultLevel() fallback
Changed the fallback from 0 to DefaultPowerLevelMember.guest.powerLevel to ensure consistency with the app's power level definitions.

3. Add E2E integration test
Added a Patrol integration test that verifies the full transfer ownership flow: opens a group chat, navigates to a member's profile, transfers ownership, then verifies the transfer button is no longer visible (confirming the current user lost owner privileges).

## Test recommendations
1. Manual pre-condition setup
This test is not idempotent — after a successful run, ownership has been transferred. Before re-running, the owner must be manually restored to the original user.

## Refactoring & code hygiene:

- Widget lifecycle & member ordering: Reordered ProfileInfoBodyController members following the [doc](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#order-other-class-members-in-a-way-that-makes-sense).
- Strict typing on class attributes: Added explicit types to class fields (e.g., final ResponsiveUtils responsive instead of implicit final responsive)
- Inverted condition fix in _goToDraftChat(): The guard clause used != instead of ==, preventing draft chats from opening for other users. Replaced with an early return pattern.
- Switch expressions: Replaced if/return chains in _handleSetPermissionState() with switch pattern matching for exhaustive state handling.
- .gitignore updates: Added test_bundle.dart (Patrol auto-generated, marked DO NOT COMMIT) and .fvm/ (Flutter Version Management local config).

## Resolved
- Web:
image
image
image

## Summary by CodeRabbit

* **New Features**
  * Transfer group ownership UI flow added and covered by a new integration test
  * New profile action controls exposed for test automation

* **Improvements**
  * Expanded localization across chat scenarios and UI text
  * Lifecycle, animation and reliability improvements in profile UI
  * Better contact-permission handling in search flow
  * Member list stability improved via identity keys
  * Minor UI text tweak ("Group info")

* **Bug Fixes**
  * More reliable waits for progress indicators and dialogs

* **Chores**
  * Updated ignore rules and test environment variables

### DIFF
--- a/integration_test/robots/chat/chat_profile_info_robot.dart
+++ b/integration_test/robots/chat/chat_profile_info_robot.dart
@@ -6,6 +6,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:patrol/patrol.dart';
 
+import 'package:fluffychat/presentation/enum/profile_info/profile_info_body_enum.dart';
+
 import '../../base/core_robot.dart';
 
 class ChatProfileInfoRobot extends CoreRobot {
@@ -181,6 +183,79 @@ class ChatProfileInfoRobot extends CoreRobot {
       await $(confirmButtons.first).tap();
     }
     await $.pumpAndSettle(timeout: const Duration(seconds: 5));
+  }
+
+  /// Find a profile action button by its enum value
+  PatrolFinder getProfileActionButton(ProfileInfoActions action) {
+    return $(find.byKey(Key('profile_action_${action.name}')));
+  }
+
+  /// Verify a profile action button is visible
+  Future<void> verifyProfileActionButtonVisible(
+    ProfileInfoActions action, {
+    bool expected = true,
+  }) async {
+    if (expected) {
+      try {
+        await getProfileActionButton(
+          action,
+        ).waitUntilVisible(timeout: const Duration(seconds: 5));
+      } catch (_) {
+        // Fall through to the expect below for a clear error message
+      }
+    } else {
+      await $.pump(const Duration(seconds: 2));
+    }
+    final button = getProfileActionButton(action);
+    expect(
+      button.exists,
+      expected,
+      reason: expected
+          ? '${action.name} button not found'
+          : '${action.name} button should not be visible',
+    );
+  }
+
+  /// Tap on a profile action button
+  Future<void> tapProfileActionButton(ProfileInfoActions action) async {
+    final button = getProfileActionButton(action);
+    await button.waitUntilVisible();
+    await button.tap();
+    await $.pump(const Duration(seconds: 2));
+  }
+
+  /// Confirm the transfer ownership dialog
+  Future<void> confirmTransferOwnership() async {
+    await $.waitUntilVisible(
+      $(find.byKey(TwakeDialog.showConfirmAlertDialogKey)),
+    );
+
+    final dialogFinder = find.byKey(TwakeDialog.showConfirmAlertDialogKey);
+    final confirmButton = find.descendant(
+      of: dialogFinder,
+      matching: find.textContaining('Confirm'),
+    );
+
+    if (confirmButton.evaluate().isNotEmpty) {
+      await $(confirmButton).tap();
+      await $.pump(const Duration(seconds: 3));
+      return;
+    }
+
+    // Fallback: tap the first confirm-style button
+    final buttons = find.descendant(
+      of: dialogFinder,
+      matching: find.byWidgetPredicate(
+        (widget) => widget is TextButton || widget is ElevatedButton,
+      ),
+    );
+
+    if (buttons.evaluate().length >= 2) {
+      await $(buttons.at(1)).tap();
+    } else if (buttons.evaluate().isNotEmpty) {
+      await $(buttons.first).tap();
+    }
+    await $.pump(const Duration(seconds: 3));
   }
 
   /// Navigate back from profile info screen
--- a/integration_test/robots/chat_group_detail_robot.dart
+++ b/integration_test/robots/chat_group_detail_robot.dart
@@ -47,7 +47,7 @@ class ChatGroupDetailRobot extends CoreRobot {
 
   Future<void> tapOnChatBarTitle() async {
     await getChatAppBarTitle().tap();
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($("Group info"));
   }
 
   Future<void> tapOnChatBarTitleForDM() async {
--- a/integration_test/scenarios/chat_scenario.dart
+++ b/integration_test/scenarios/chat_scenario.dart
@@ -1,5 +1,6 @@
 import 'dart:developer';
 import 'dart:io';
+
 import 'package:fluffychat/generated/l10n/app_localizations.dart';
 import 'package:fluffychat/pages/chat/chat_app_bar_title.dart';
 import 'package:fluffychat/pages/chat/chat_input_row.dart';
@@ -8,6 +9,7 @@ import 'package:fluffychat/pages/chat/chat_pinned_events/pinned_events_view.dart
 import 'package:fluffychat/pages/chat/chat_pinned_events/pinned_messages_screen.dart';
 import 'package:fluffychat/pages/chat/chat_view.dart';
 import 'package:fluffychat/pages/chat/event_info_dialog.dart';
+import 'package:fluffychat/pages/chat/events/images_builder/sending_image_info_widget.dart';
 import 'package:fluffychat/pages/chat/events/message/multi_platform_message_container.dart';
 import 'package:fluffychat/pages/chat/events/message_content.dart';
 import 'package:fluffychat/pages/chat/events/message_time.dart';
@@ -16,18 +18,18 @@ import 'package:fluffychat/pages/chat_draft/draft_chat_view.dart';
 import 'package:fluffychat/pages/chat_list/chat_list_body_view.dart';
 import 'package:fluffychat/widgets/context_menu/context_menu_action_item_widget.dart';
 import 'package:fluffychat/widgets/twake_components/twake_icon_button.dart';
-import 'package:fluffychat/pages/chat/events/images_builder/sending_image_info_widget.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:linagora_design_flutter/images_picker/image_item_widget.dart';
 import 'package:linkfy_text/linkfy_text.dart';
 import 'package:patrol/patrol.dart';
+
 import '../base/core_robot.dart';
 import '../help/soft_assertion_helper.dart';
 import '../robots/add_member_robot.dart';
 import '../robots/chat_group_detail_robot.dart';
 import '../robots/chat_list_robot.dart';
-import 'package:flutter/material.dart';
 import '../robots/group_information_robot.dart';
 import '../robots/menu_robot.dart';
 import '../robots/new_chat_robot.dart';
@@ -39,6 +41,11 @@ enum UserLevel { member, admin, owner, moderator }
 
 class ChatScenario extends CoreRobot {
   ChatScenario(super.$);
+
+  L10n get _l10n {
+    final context = $.tester.element(find.byType(Scaffold));
+    return L10n.of(context)!;
+  }
 
   Future<void> enterSearchText(String searchText) async {
     await SearchRobot($).enterSearchText(searchText);
@@ -66,7 +73,7 @@ class ChatScenario extends CoreRobot {
     }
     await AddMemberRobot($).getNextIcon().tap();
     await AddMemberRobot($).getAgreeInviteMemberBtn().tap();
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($(_l10n.groupInfo));
     return (await GroupInformationRobot($).getListOfMembers()).length;
   }
 
@@ -77,7 +84,7 @@ class ChatScenario extends CoreRobot {
       await GroupInformationRobot($).clickOnRemoveFromGroup();
       await GroupInformationRobot($).clickOnAgreeIRemoveMemberBtn();
     }
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($(_l10n.groupInfo));
     return (await GroupInformationRobot($).getListOfMembers()).length;
   }
 
@@ -149,6 +156,9 @@ class ChatScenario extends CoreRobot {
 
     // 2) Get MaterialLocalizations from the input's own context (NOT WidgetsApp)
     final BuildContext inputCtx = $.tester.element(input.finder);
+    if (inputCtx.mounted == false) {
+      throw StateError('Input context is not mounted');
+    }
     final String pasteLabel = MaterialLocalizations.of(
       inputCtx,
     ).pasteButtonLabel;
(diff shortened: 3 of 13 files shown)


lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -71,15 +68,35 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
+    _avatarToggleTimer?.cancel();
+    _setPermissionLevelSubscription?.cancel();
+    super.dispose();
+  }
+
   void getUserInfoAction() {
     if (user == null) return;
-    userInfoNotifierSub = _getUserInfoInteractor
+    userInfoNotifierSub = getIt
+        .get<GetUserInfoInteractor>()
```
nqhhdev, line 99: Why do we need to change it?

integration_test/scenarios/chat_scenario.dart
```diff
@@ -66,7 +66,7 @@ class ChatScenario extends CoreRobot {
     }
     await AddMemberRobot($).getNextIcon().tap();
     await AddMemberRobot($).getAgreeInviteMemberBtn().tap();
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($("Group info"));
```
nqhhdev: Should use l10n

integration_test/scenarios/chat_scenario.dart
```diff
@@ -77,7 +77,7 @@ class ChatScenario extends CoreRobot {
       await GroupInformationRobot($).clickOnRemoveFromGroup();
       await GroupInformationRobot($).clickOnAgreeIRemoveMemberBtn();
     }
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($("Group info"));
```
nqhhdev: Should use l10n

lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -230,60 +251,72 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
-                user!.room.ownUser: DefaultPowerLevelMember.admin.powerLevel,
-              },
-            )
-            .listen((state) => _handleSetPermissionState(state));
-      }
+      if (result == ConfirmResult.cancel) return;
+
+      final Room room = user!.room;
+      _setPermissionLevelSubscription = getIt
+          .get<SetPermissionLevelInteractor>()
```
nqhhdev, line 276: Why do we need to change it?

lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -230,60 +251,72 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
-        }
-
-        if (failure is NoPermissionFailure) {
-          TwakeDialog.hideLoadingDialog(context);
-          TwakeSnackBar.show(
-            context,
-            L10n.of(context)!.permissionErrorChangeRole,
-          );
-          return;
+      (Failure failure) {
```
nqhhdev, line 292: Should check mounted state before using context to avoid errors when the widget is disposed while async operations are in progress.

integration_test/robots/chat_group_detail_robot.dart
```diff
@@ -47,7 +47,7 @@ class ChatGroupDetailRobot extends CoreRobot {
 
   Future<void> tapOnChatBarTitle() async {
     await getChatAppBarTitle().tap();
-    await $.waitUntilVisible($("Group information"));
+    await $.waitUntilVisible($("Group info"));
```
nqhhdev, line 50: Should use l10n

integration_test/robots/chat/chat_profile_info_robot.dart
```diff
@@ -183,6 +183,80 @@ class ChatProfileInfoRobot extends CoreRobot {
...
+  /// Confirm the transfer ownership dialog
+  Future<void> confirmTransferOwnership() async {
+    await $.waitUntilVisible(
+      $(find.byKey(TwakeDialog.showConfirmAlertDialogKey)),
+    );
+
+    final dialogFinder = find.byKey(TwakeDialog.showConfirmAlertDialogKey);
+    final confirmButton = find.descendant(
+      of: dialogFinder,
+      matching: find.textContaining('Confirm'),
```
nqhhdev, line 236: Should use l10n

integration_test/robots/chat/chat_profile_info_robot.dart
```diff
@@ -183,6 +183,80 @@ class ChatProfileInfoRobot extends CoreRobot {
     await $.pumpAndSettle(timeout: const Duration(seconds: 5));
   }
 
+  /// Find the "Transfer ownership" button in the profile info screen
+  PatrolFinder getTransferOwnershipButton() {
+    return $(find.byKey(const Key('profile_action_transferOwnership')));
+  }
```
nqhhdev: Currently we have 4 actions. Should we make it dynamic so we can reuse it in other cases? WDYT?

```dart
  sendMessage,
  downgradeToReadOnly,
  removeFromGroup,
  transferOwnership;
```

lib/pages/profile_info/profile_info_body/profile_info_body.dart
9clg6: GetUserInfoInteractor is only used once in getUserInfoAction(), so there's no need to store it as a class attribute. It's a stateless service resolved via DI — calling getIt.get() inline is equivalent and keeps the class leaner.
Declaring it as a class attribute "for later" would go against [YAGNI](https://martinfowler.com/bliki/Yagni.html) (You Aren't Gonna Need It), we shouldn't add complexity for hypothetical future use. The StreamSubscription (userInfoNotifierSub) is still kept as a class attribute to properly cancel it in dispose().

lib/pages/profile_info/profile_info_body/profile_info_body.dart
9clg6: For the same reason as above : 

> GetUserInfoInteractor is only used once in getUserInfoAction(), so there's no need to store it as a class attribute. It's a stateless service resolved via DI — calling getIt.get() inline is equivalent and keeps the class leaner.
> Declaring it as a class attribute "for later" would go against [YAGNI](https://martinfowler.com/bliki/Yagni.html) (You Aren't Gonna Need It), we shouldn't add complexity for hypothetical future use. The StreamSubscription (userInfoNotifierSub) is still kept as a class attribute to properly cancel it in dispose().

lib/pages/profile_info/profile_info_body/profile_info_body.dart
9clg6: I will add the rule to the linter, but this will cause a lot of changes that will need to be reviewed. I will create a new dedicated merge request.

integration_test/tests/chat/transfer_ownership_test.dart
```diff
@@ -0,0 +1,61 @@
...
+);
+
+void main() {
+  TestBase().runPatrolTest(
+    tags: ['transfer_ownership_test_01'],
+    description:
+        'Transfer ownership sets target to own level and self to users_default',
+    test: ($) async {
+      // 1. Open the group chat
+      await ChatListRobot($).openSearchScreen();
```
nqhhdev, line 27: I have an idea for this test case. Currently, we search for an existing room and then perform the remaining steps. If the room isn’t found, the test breaks.

IMO, we should create a new group and invite some members first, then continue with the remaining steps. What do you think?

cc: @9clg6

hoangdat (approved):
LGTM

integration_test/tests/chat/transfer_ownership_test.dart
```diff
@@ -0,0 +1,61 @@
...
+const memberMatrixID = String.fromEnvironment(
+  'MemberMatrixID',
+  defaultValue: '@member:localhost',
+);
+
+void main() {
+  TestBase().runPatrolTest(
+    tags: ['transfer_ownership_test_01'],
+    description:
+        'Transfer ownership sets target to own level and self to users_default',
```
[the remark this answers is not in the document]
9clg6: Good point. The current E2E test validates the happy path from the UI perspective (button visibility before/after). Verifying the actual power levels via the Matrix API would make it more robust but adds significant complexity for an integration test. The regression itself is covered by the code fix — this test ensures the flow doesn't break. But i'll change the description to make it fit better with the test scenario.

integration_test/tests/chat/transfer_ownership_test.dart
```diff
@@ -0,0 +1,61 @@
...
+      await $.pump(const Duration(seconds: 3));
+
+      // 6. After transfer, the profile page pops back to Group Info.
+      //    Re-open the same member's profile to verify the button is gone.
+      await GroupInformationRobot($).openMemberDetail(matrixID: memberMatrixID);
+      await ChatProfileInfoRobot($).verifyProfileActionButtonVisible(
+        ProfileInfoActions.transferOwnership,
+        expected: false,
+      );
+    },
```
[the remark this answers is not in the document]
9clg6: Acknowledged. Ideally the test should be self-cleaning. However, setting up a disposable room or restoring ownership via API in teardown adds significant complexity. For now, the test relies on a pre-configured room, same pattern as our other E2E tests. 

This is consistent with all our existing E2E tests : none implement teardown or self-cleaning. For example, chat_group_test pins/unpins messages without cleanup, create_new_group_chat_test creates groups without deleting them. Making tests self-cleaning would be a cross-cutting improvement for the entire suite, not specific to this PR.

integration_test/tests/chat/transfer_ownership_test.dart
9clg6: Good idea, it would make the test self-contained. However, creating a room + inviting a member + having them accept requires a second active account and significantly increases test complexity and duration. This is consistent with how all our other E2E tests work today : they all rely on pre-existing rooms via SearchByTitle. I'd prefer to tackle this as a cross-cutting improvement for the whole test suite rather than just this test. WDYT?

lib/domain/model/room/room_extension.dart
```diff
@@ -355,11 +355,15 @@ extension RoomExtension on Room {
...
+  ///
   int getUserDefaultLevel() {
+    const usersDefaultKey = 'users_default';
+
     return getState(
           EventTypes.RoomPowerLevels,
-        )!.content.tryGet<int>('users_default') ??
-        0;
+        )!.content.tryGet<int>(usersDefaultKey) ??
+        DefaultPowerLevelMember.guest.powerLevel;
```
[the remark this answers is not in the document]
9clg6: You are right, i'll fix it. Even if we have no feedback about a potential issue related to.

lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -71,15 +68,35 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
 
+  @override
+  void initState() {
+    super.initState();
+    animationController = AnimationController(
+      vsync: this,
+      duration: const Duration(milliseconds: _animationDuration),
+    );
+    getUserInfoAction();
+  }
```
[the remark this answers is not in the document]
9clg6: This is pre-existing behavior unrelated to this PR. I'd prefer to address it in a separate ticket to keep this PR focused.

lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -230,60 +251,72 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
+          .get<SetPermissionLevelInteractor>()
+          .execute(
+            room: room,
+            userPermissionLevels: {
+              user!: room.ownUser.powerLevel,
+              room.ownUser: room.getUserDefaultLevel(),
+            },
+          )
+          .listen(_handleSetPermissionState);
     });
```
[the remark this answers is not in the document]
9clg6: transferOwnership() is only reachable when !isOwnProfile (the button is hidden otherwise), so user! and room.ownUser can never be the same user. No duplicate key issue can occur.

tddang-linagora (approved):

integration_test/robots/chat_group_detail_robot.dart
nqhhdev: Any update @9clg6

integration_test/robots/chat/chat_profile_info_robot.dart
nqhhdev: Any update @9clg6

integration_test/tests/chat/transfer_ownership_test.dart
nqhhdev: I agree. Could you update this test case first, please?

9clg6: Thanks! Since we agree this should be a cross-cutting improvement for the whole test suite, I'd rather not change just this one test in isolation — it would make it inconsistent with all the others.

integration_test/robots/chat/chat_profile_info_robot.dart
9clg6: I fixed it, it now uses l10n

lib/pages/profile_info/profile_info_body/profile_info_body.dart
```diff
@@ -230,60 +251,72 @@ class ProfileInfoBodyController extends State<ProfileInfoBody>
...
+
+      final Room room = user!.room;
+      _setPermissionLevelSubscription = getIt
+          .get<SetPermissionLevelInteractor>()
+          .execute(
+            room: room,
+            userPermissionLevels: {
+              user!: room.ownUser.powerLevel,
+              room.ownUser: room.getUserDefaultLevel(),
+            },
```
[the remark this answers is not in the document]
9clg6: This feedback has already been answered before.

integration_test/robots/chat/chat_profile_info_robot.dart
```diff
@@ -183,6 +185,79 @@ class ChatProfileInfoRobot extends CoreRobot {
...
+      await $.pump(const Duration(seconds: 2));
+    }
+    final button = getProfileActionButton(action);
+    expect(
+      button.exists,
+      expected,
+      reason: expected
+          ? '${action.name} button not found'
+          : '${action.name} button should not be visible',
+    );
```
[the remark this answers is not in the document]
9clg6: Tests worked all on all runs

lib/pages/profile_info/profile_info_body/profile_info_body.dart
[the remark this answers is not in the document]
pm-McFly: @coderabbitai
This is a peculiar situation that could arrive when this matrix client is used against a room created outside a Twake Chat server. In that case the fallback is legitimate to 0, as per community 0 is usually the default level where people can write messages.
In specific cases where 0 cannot write (e.g. our Twake Chat model), rooms will always have a `user_default` set and this would correctly sets the level to the according value.

No error here.

tddang-linagora (approved):

PR #2918, opened 2026-03-16
feat(search): validate mxid profile via API before displaying external contact

9clg6:
## Ticket
TID-2474 - Improve searching by the non-exist matrix address

## Root cause
`SearchExternalContactWidget` displayed any syntactically valid mxid as a contact result without verifying it exists on the server. The check was purely format-based (`isValidMatrixId`), so typing `@noexist:server.com` showed a fake profile instead of "No Results".

## Solution
Converted `SearchExternalContactWidget` from `StatelessWidget` to `StatefulWidget` with a `FutureBuilder` that calls `client.getProfileFromUserId()` before rendering:
- **200** → display the contact with real profile data
- **404/403/5xx** → display `EmptySearchWidget` ("No Results")
- **Loading** → display a spinner

## Impact description
N/A (bug fix)

## Test recommendations
1. Search `@validuser:your-server.com` → should show profile info
2. Search `@nonexistent:your-server.com` → should show "No Results" mascot

## Pre-merge
No

## Resolved
- Web:
image

- Android:
- IOS:

https://github.com/user-attachments/assets/c9907079-b9ad-425b-ba80-7fb7a0078300

## Summary by CodeRabbit

* **New Features**
  * External contact search now fetches remote profile data, shows a loading indicator, and accepts an optional test client for testing.
* **Bug Fixes**
  * Shows an explicit empty/no-results state when profile lookup fails or returns no data.
* **Improvements**
  * UI updates responsively when query/ID changes and displays validated contact details; taps use fetched profile info.
* **Tests**
  * Added widget and integration tests covering loading, no-result, and successful profile fetch flows.

### DIFF
--- a/lib/pages/contacts_tab/contacts_tab_body_view.dart
+++ b/lib/pages/contacts_tab/contacts_tab_body_view.dart
@@ -18,6 +18,7 @@ import 'package:fluffychat/widgets/contacts_warning_banner/contacts_warning_bann
 import 'package:fluffychat/widgets/sliver_expandable_list.dart';
 import 'package:flutter/material.dart';
 import 'package:fluffychat/generated/l10n/app_localizations.dart';
+import 'package:matrix/matrix.dart';
 
 class ContactsTabBodyView extends StatelessWidget {
   final ContactsTabController controller;
@@ -268,7 +269,7 @@ class _SliverContactsList extends StatelessWidget {
   }
 }
 
-class _SilverExternalContact extends StatelessWidget {
+class _SilverExternalContact extends StatefulWidget {
   final ContactsTabController controller;
   final PresentationContact externalContact;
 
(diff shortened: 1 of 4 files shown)


lib/pages/search/search_external_contact.dart
```diff
@@ -16,31 +20,84 @@ class SearchExternalContactWidget extends StatelessWidget {
...
   final SearchController searchController;
 
+  @override
+  State<SearchExternalContactWidget> createState() =>
+      _SearchExternalContactWidgetState();
+}
+
+class _SearchExternalContactWidgetState
+    extends State<SearchExternalContactWidget> {
+  Future<CachedProfileInformation>? _profileFuture;
```
hoangdat, line 32: can you separate it as a inter-actor to reuse in other controller? like search in contact, add member, ...

lib/pages/search/search_external_contact.dart
```diff
@@ -16,31 +20,84 @@ class SearchExternalContactWidget extends StatelessWidget {
...
+    }
+  }
+
+  void _fetchProfileIfNeeded() {
+    if (_currentKeyword != widget.keyword) {
+      _currentKeyword = widget.keyword;
+      final client = Matrix.of(context).client;
+      _profileFuture = client.getUserProfile(
+        widget.keyword,
+        maxCacheAge: Duration.zero,
```
hoangdat, line 55: how about to keep longer in cache?

9clg6: This is actually the existing behavior — `Duration.zero` was already in place in `SearchExternalContactWidget` before this PR, we just replicated the same pattern on the Contacts side.

For a search context, we want fresh data: if a user just created their account or updated their profile, we need the lookup to reflect that immediately. Caching here could lead to stale "not found" results persisting across searches.

I'd prefer to keep it as is.

lib/pages/search/search_external_contact.dart
9clg6: ight now the profile lookup is a single SDK call (client.getUserProfile) — there's no extra business logic around it that would benefit from an interactor layer.

Introducing one at this point would add indirection without a clear gain. That said, abstracting SDK interactions is something I'd like to propose as part of a broader discussion on the app's overall architecture. We can revisit this then.

hoangdat: but right now we should have the same behavior for searching in contact and adding member in group chat. IMO we should apply the same

9clg6: Out of scope, create a ticket for that

hoangdat: no, we should not apply for other screen but at least we should separte to a interactor for later use

lib/pages/search/search_external_contact.dart
```diff
@@ -4,43 +4,102 @@ import 'package:fluffychat/pages/search/search.dart';
...
-      child: InkWell(
-        onTap: () {
-          searchController.onSearchItemTap(
-            ContactPresentationSearch(
-              matrixId: newContact.matrixId,
-              displayName: newContact.displayName,
+        final profile = snapshot.data!;
+        final newContact = PresentationContact(
+          matrixId: widget.keyword,
+          displayName: profile.displayname ?? widget.keyword.substring(1),
```
hoangdat, line 79: should precheck for `substring(1)`?

lib/pages/search/search_external_contact.dart
9clg6: An interactor in this codebase earns its place when there's business logic to encapsulate - state sequencing, failure mapping, repository abstraction. Here it would be an empty wrapper around a single SDK call with no identified future consumer ("for later" is not a use case).                                                                                                                                                                                         

YAGNI applies: we add the interactor when the logic justifies it, not in anticipation of it.

If we were to abstract the SDK call, the right place would be a datasource, not an interactor - that's what respects the Dependency Inversion Principle: the domain layer (interactors) depends on abstractions, and datasources are those abstractions over external SDKs/infrastructure.

Adding an interactor that directly calls client.getUserProfile() would actually violate DIP by coupling domain logic to the SDK implementation. The current direct call in the widget is at least honest about what it is.

lib/pages/search/search_external_contact.dart
9clg6: No because we already check 
`searchWord.isValidMatrixId && searchWord.startsWith('@');` then keyword can't be empy

lib/pages/contacts_tab/contacts_tab_body_view.dart
```diff
@@ -277,24 +278,93 @@ class _SilverExternalContact extends StatelessWidget {
...
+      _profileFuture = null;
+      return;
+    }
+    if (_currentMatrixId != matrixId) {
+      _currentMatrixId = matrixId;
+      _profileFuture = widget.controller.client.getUserProfile(
+        matrixId,
+        maxCacheAge: Duration.zero,
+      );
+    }
```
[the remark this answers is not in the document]
9clg6: Not in scope and even if the user switch account the whole tree is rebuilt

lib/pages/search/search_external_contact.dart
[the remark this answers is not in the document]
9clg6: If the name is a whitespace is it really my problem ? I will not handle every edge-cases, this should be check at account creation. 
Not in scope

PR #2920, opened 2026-03-16
fix: move state mutations inside setState callbacks [2/7]

9clg6:
## Summary
Move all state mutations inside `setState()` callbacks instead of mutating before calling empty `setState(() {})`. Remove redundant `setState()` calls and simplify control flow.

## PR Chain (Performance Optimization Series)
```
PR1 ← main
PR2 ← PR1            ← YOU ARE HERE
PR3 ← PR2     PR4 ← PR2     PR5 ← PR2
       \         |
        PR6 ← PR3 + PR4
              |
             PR7 ← PR6
```
Depends on: #2919

## Changes
- **9 files** with setState hygiene fixes
- State mutations moved inside callbacks in: `tom_bootstrap_dialog`, `key_verification_dialog`, `login`, `homeserver_picker`, `add_widget_tile`, `media_viewer`, `chat_settings_popup_menu`, `mxc_image`
- Import reordering in `matrix.dart` and `homeserver_picker.dart`
- Redundant switch-case removed in `homeserver_picker.dart`

## Why
Empty `setState(() {})` is a code smell — Flutter still schedules a rebuild, but the intent is unclear. Moving mutations inside the callback makes the relationship between state change and rebuild explicit, and prevents bugs where mutations happen but `setState` is never reached (e.g., early return).


hoangdat:
hi @9clg6 so sorry with the format change in weblate.
- weblate removed all `"@xxx": {},`, so it made a lot of change. Now I think we should find the way to follow weblate or config weblate to follow our format. (but I cannot config it now)

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: fluffychat
 description: A convenient Matrix-based tool for personal and corporate communication.
 publish_to: none
-version: 2.20.4+2330
+version: 2.20.3+2330
```
hoangdat: revert it

lib/pages/homeserver_picker/homeserver_picker.dart
```diff
@@ -211,10 +209,11 @@ class HomeserverPickerController extends State<HomeserverPicker>
...
-        state = HomeserverState.otherLoginMethod;
         context.push('/connect');
         FocusManager.instance.primaryFocus?.unfocus();
-        setState(() {});
+        setState(() {
+          state = HomeserverState.otherLoginMethod;
+        });
       }
     } catch (e) {
       state = HomeserverState.wrongServerName;
```
hoangdat: how to move it inside too?

lib/pages/homeserver_picker/homeserver_picker.dart
```diff
@@ -224,18 +223,8 @@ class HomeserverPickerController extends State<HomeserverPicker>
...
-      case HomeserverState.wrongServerName:
-        await checkHomeserverAction();
-        break;
-      default:
-        await checkHomeserverAction();
-        break;
-    }
-    setState(() {});
+    // All branches call checkHomeserverAction which already handles setState
+    await checkHomeserverAction();
```
hoangdat, line 229: should do a manual test to ensure it work as expect

hoangdat (commented):
quite good to me. still 2 minor comments

pubspec.yaml
9clg6: Yes, it was a rebase mistake

PR #2931, opened 2026-03-17
perf: offload /sync JSON parsing to background isolate/web worker

9clg6:
## Summary

Move the expensive `/sync` response parsing off the main thread to eliminate UI freezes during Matrix sync.

## Problem

The `/sync` JSON response parsing (`utf8.decode` + `jsonDecode` + `SyncUpdate.fromJson`) was running entirely on the main thread, blocking the UI for **~380ms per sync call** (measured with CPU 4x throttling).

## Solution

Introduce `TwakeClient extends Client` that overrides `sync()` to offload parsing via `compute()`:

- **Native** (iOS/Android/Desktop): the entire pipeline runs in a background Dart isolate
- **Web**: only `utf8.decode` + `jsonDecode` (~80% of cost) runs in a web worker. `SyncUpdate.fromJson` stays on the main thread because web workers can only transfer Maps/Lists/primitives efficiently through structured clone — not custom Dart class instances

### Why a subclass instead of modifying the SDK?

- `Api.sync()` is **generated code** in `dart_matrix_api_lite` — modifications would be overwritten on regeneration
- The override is non-intrusive: `Client` exposes `httpClient`, `baseUri`, `bearerToken`, and `unexpectedResponse()` — everything needed to replicate the HTTP request
- Error handling, `FixedTimeoutHttpClient` timeout wrapper, soft logout, and token refresh all work identically to the original (verified by code review of the SDK internals)

> A maintenance notice is included in the code: when upgrading the `matrix` SDK, the generated `Api.sync()` signature should be compared with this override.

## Results

> Metrics collected via Chrome DevTools performance trace with CPU 4x throttling (simulating a low-end device). The raw flame chart JSON exceeded 350,000 lines — analysis was automated using the Chrome DevTools MCP on Claude Code to run the app, navigate through conversations, record performance traces, and extract metrics programmatically.

| Metric | Before | After | Delta |
|---|---|---|---|
| **Sync processing (main thread)** | **381.7ms** | **8.7ms** | **-97.7%** |

## Files changed

| File | Change |
|---|---|
| `lib/utils/matrix_sdk_extensions/twake_client.dart` | New file |
| `lib/utils/client_manager.dart` | Import + `Client(` → `TwakeClient(` |

## Summary by CodeRabbit

* **Refactor**
  * Sync operation rewritten for safer error handling and platform-specific parsing/offloading to improve reliability and performance on web and native.
  * Sync request/response processing now buffers and validates responses more robustly to reduce malformed-sync failures.
  * Client creation now produces an updated client implementation while keeping the public creation API unchanged.

### DIFF
--- a/lib/utils/matrix_sdk_extensions/twake_client.dart
+++ b/lib/utils/matrix_sdk_extensions/twake_client.dart
@@ -1,0 +1,122 @@
+import 'dart:convert';
+
+import 'package:flutter/foundation.dart';
+import 'package:http/http.dart' as http;
+import 'package:matrix/matrix.dart';
+
+/// Top-level function for native isolate-based sync response parsing.
+/// Returns a fully constructed [SyncUpdate] since Dart isolates can
+/// transfer complex objects without serialization overhead.
+SyncUpdate _parseSyncResponseNative(Uint8List responseBody) {
+  final responseString = utf8.decode(responseBody);
+  final json = jsonDecode(responseString);
+  return SyncUpdate.fromJson(json as Map<String, Object?>);
+}
+
+/// Top-level function for web JSON decoding via `compute()`.
+/// Returns only the raw [Map] because structured clone (used by web workers)
+/// transfers Maps/Lists/primitives efficiently, whereas custom Dart class
+/// instances like [SyncUpdate] do not. [SyncUpdate.fromJson] is therefore
+/// called on the main thread after the transfer.
+Map<String, Object?> _jsonDecodeSyncResponseWeb(Uint8List responseBody) {
+  final responseString = utf8.decode(responseBody);
+  return jsonDecode(responseString) as Map<String, Object?>;
+}
+
+/// A [Client] subclass that offloads the heavy JSON parsing of `/sync`
+/// responses to a background isolate / web worker.
+///
+/// On **native** platforms (iOS, Android, macOS, Linux), the entire parsing
+/// pipeline (`utf8.decode` + `jsonDecode` + `SyncUpdate.fromJson`) runs in
+/// a background isolate via `compute()`.
+///
+/// On **web**, only `utf8.decode` + `jsonDecode` (~80% of parsing time) runs
+/// via `compute()`. On Flutter Web with WASM, this executes at near-native
+/// speed (not in a separate thread). `SyncUpdate.fromJson` remains on the
+/// main thread because structured clone cannot transfer custom Dart class
+/// instances — only Maps/Lists/primitives pass without overhead.
+///
+/// ## Maintenance notice
+///
+/// This override replicates the HTTP request logic from the generated
+/// `Api.sync()` in `matrix_api_lite`. If the upstream SDK changes the
+/// sync endpoint (new parameters, headers, or path), this override must
+/// be updated accordingly. When upgrading the `matrix` SDK dependency,
+/// compare the generated `Api.sync()` signature and body with this
+/// implementation.
+// Validated against matrix SDK ^6.0.0
+class TwakeClient extends Client {
+  TwakeClient(
+    super.clientName, {
+    required super.database,
+    super.legacyDatabaseBuilder,
+    super.verificationMethods,
+    super.httpClient,
+    super.importantStateEvents,
+    super.roomPreviewLastEvents,
+    super.pinUnreadRooms,
+    super.pinInvitedRooms,
+    super.requestHistoryOnLimitedTimeline,
+    super.supportedLoginTypes,
+    super.mxidLocalPartFallback,
+    super.formatLocalpart,
+    super.nativeImplementations,
+    super.logLevel,
+    super.syncFilter,
+    super.defaultNetworkRequestTimeout,
+    super.sendTimelineEventTimeout,
+    super.customImageResizer,
+    super.shareKeysWith,
+    super.enableDehydratedDevices,
+    super.receiptsPublicByDefault,
+    super.onSoftLogout,
+    super.typingIndicatorTimeout,
+    super.convertLinebreaksInFormatting,
+  });
+
+  @override
+  Future<SyncUpdate> sync({
+    String? filter,
+    String? since,
+    bool? fullState,
+    PresenceType? setPresence,
+    int? timeout,
+    bool? useStateAfter,
+  }) async {
+    // Build the request identically to the generated Api.sync().
+    final requestUri = Uri(
+      path: '_matrix/client/v3/sync',
+      queryParameters: {
+        if (filter != null) 'filter': filter,
+        if (since != null) 'since': since,
+        if (fullState != null) 'full_state': fullState.toString(),
+        if (setPresence != null) 'set_presence': setPresence.name,
+        if (timeout != null) 'timeout': timeout.toString(),
+        if (useStateAfter != null) 'use_state_after': useStateAfter.toString(),
+      },
+    );
+    final request = http.Request('GET', baseUri!.resolveUri(requestUri));
+    request.headers['authorization'] = 'Bearer ${bearerToken!}';
+    final response = await httpClient.send(request);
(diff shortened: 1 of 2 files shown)


lib/utils/matrix_sdk_extensions/twake_client.dart
nqhhdev, line 24: - Why did you choose Top-level and explain it?
- IMO: Should make it static inside the class? WDYT

tddang-linagora (commented):
- [ ] Add before after screenshots of the improvement

hoangdat:
what did you profile on this PR. I tried but not much thing

image

Crash--:
If you need help to test that in real condition, I'll be more than happy to be able to test on web.

hoangdat:
in my site, it quite be stucked at login step of the app. How you can bypass this @9clg6

9clg6:
> yes, agree your performance metrics very detail and interesting. But for me, I still not really understand what is the comparision.
> 
> Fx:
> 
> * how the AI can measure Sync main thread? the sync is the same between 2 snapshot?
> * what is the jank? in which feature? behavior?
> * bla bla bla ....
> * and so on?

The AI doesn't invent metrics. It connects to Chrome via the [DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/) (same protocol behind the Performance panel in your browser), records a standard trace (Tracing.start() / Tracing.end()), and parses the resulting .json.gz — the exact same file you get when you click Record manually.

Sync main thread: Chrome records XHRLoad events for every /sync response. The dur field = time the main thread was blocked parsing the JSON. 379.8ms → 13.5ms is not an estimate, it's a Chrome-measured microsecond-precision timestamp.

Same sync?: No, /sync payloads depend on server activity. But same account, same rooms, same navigation path (~60s). The difference shows on large responses — the heavy parsing is now offloaded to a web worker.

Jank: Any frame > 16.7ms (60fps budget). Not feature-specific — it's global UI smoothness. Scroll stutters, transition freezes. Severe jank (> 50ms) dropped from 43 frames to 1.

The raw traces are available — load them in Chrome DevTools > Performance > Load profile and verify every number yourself.

lib/utils/matrix_sdk_extensions/twake_client.dart
```diff
@@ -0,0 +1,123 @@
...
+    if (response.statusCode != 200) {
+      unexpectedResponse(response, responseBody);
+    }
+
+    try {
+      if (kIsWeb) {
+        // On web: offload utf8.decode + jsonDecode to a web worker,
+        // then do the lightweight SyncUpdate.fromJson on the main thread.
+        final json = await compute(_jsonDecodeSyncResponseWeb, responseBody);
+        return SyncUpdate.fromJson(json);
```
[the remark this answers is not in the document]
9clg6: You are right but compute works with WASM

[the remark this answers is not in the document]
9clg6: Anyway the performance gain has been mesured, then we will merge it and see if there is a improvment for final users. It needs to come with others suggestions.

lib/utils/matrix_sdk_extensions/twake_client.dart
```diff
@@ -0,0 +1,123 @@
...
+    final request = http.Request('GET', baseUri!.resolveUri(requestUri));
+    request.headers['authorization'] = 'Bearer ${bearerToken!}';
+    final response = await httpClient.send(request);
+    final responseBody = await response.stream.toBytes();
+    if (response.statusCode != 200) {
+      unexpectedResponse(response, responseBody);
+    }
+
+    try {
+      if (kIsWeb) {
```
hoangdat, line 107: how about apply it also for other platform

lib/utils/matrix_sdk_extensions/twake_client.dart
```diff
@@ -0,0 +1,123 @@
...
+///
+/// ## Maintenance notice
+///
+/// This override replicates the HTTP request logic from the generated
+/// `Api.sync()` in `matrix_api_lite`. If the upstream SDK changes the
+/// sync endpoint (new parameters, headers, or path), this override must
+/// be updated accordingly. When upgrading the `matrix` SDK dependency,
+/// compare the generated `Api.sync()` signature and body with this
+/// implementation.
+class TwakeClient extends Client {
```
hoangdat, line 48: we should have a comment about `// Validated against matrix-6.0.0` to keep track with the upgrade of SDK. I worry if `Client` 's constructor be changed, maybe we miss it.

lib/utils/matrix_sdk_extensions/twake_client.dart
```diff
@@ -0,0 +1,123 @@
...
+        return SyncUpdate.fromJson(json);
+      }
+
+      // On native: offload the entire pipeline to a background isolate.
+      return await compute(_parseSyncResponseNative, responseBody);
+    } catch (e) {
+      // Re-throw parsing errors with context so _innerSync's catch chain
+      // can distinguish them from network/connection errors.
+      throw FormatException(
+        'Failed to parse /sync response in background isolate: $e',
```
hoangdat: `'Failed to parse /sync response: $e',`

hoangdat (commented):
@9clg6 

Good work overall — native isolate path is solid. One comment to fix before merge:              
                                                                                                  
compute() on web (_isolates_web.dart) is just await null; return callback(message) — no web
worker, no thread. The parsing still runs on the main thread. The XHRLoad metric looks great    
because await null breaks it into a separate task in the trace, but jank only dropped 4 pts     
which reflects the real picture.                                                                
                                                                                                
The gains are still real (task splitting helps frame scheduling), just update the comment to say
 "defers parsing to next microtask" rather than "web worker offload" to avoid misleading future 
readers.

And dont forget to resolve comment from Rabbit

lib/utils/matrix_sdk_extensions/twake_client.dart
9clg6: Top-level is required for `compute()` — it needs a top-level or static function (no closure over instance state). Both work, top-level is the idiomatic convention for `compute()` callbacks in Flutter.

lib/utils/matrix_sdk_extensions/twake_client.dart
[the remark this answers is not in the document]
9clg6: Fixed — updated comments to reflect the actual behavior (WASM near-native speed, not a separate thread).

lib/utils/matrix_sdk_extensions/twake_client.dart
9clg6: Intentional — on web, `SyncUpdate.fromJson` can't be returned from a `compute()` callback because structured clone doesn't serialize custom Dart class instances. Only `Map`/`List`/primitives transfer cleanly, so `fromJson` has to stay on the main thread.

lib/utils/matrix_sdk_extensions/twake_client.dart
9clg6: Fixed — added `// Validated against matrix SDK ^6.0.0`.

lib/utils/matrix_sdk_extensions/twake_client.dart
9clg6: Fixed.

lib/utils/matrix_sdk_extensions/twake_client.dart
```diff
@@ -0,0 +1,123 @@
+import 'dart:convert';
+
+import 'package:flutter/foundation.dart';
+import 'package:http/http.dart' as http;
+import 'package:matrix/matrix.dart';
```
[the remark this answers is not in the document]
9clg6: Fixed.

hoangdat (changes requested):

hoangdat (approved):

PR #2933, opened 2026-03-17
perf: evict image cache on chat dispose to prevent OOM kills

9clg6:
## Problem

Flutter's global LRU image cache evicts only on insertion, not on idle. In a room with 200+ media messages, decoded bitmaps accumulate in memory indefinitely once the user stops scrolling.

Decoded bitmaps are expensive: a 2 MP photo is ~8 MB in memory (RGBA uncompressed). A dense room pushes the cache to 80–100 MB with no automatic relief. On iOS, there is no swap — that's a direct path to OOM kills.

## Solution

Call `imageCache.clear()` + `clearLiveImages()` in `ChatState.dispose()`, releasing decoded bitmaps when the user leaves the room.

## Measured on device (profile mode, same room, ~30s scroll)

| | This branch | main |
|---|---|---|
| Dart heap after leaving room | 64.7 MB | 96.8 MB |
| External memory (decoded bitmaps) | 0.6 MB | 2.9 MB |

Dart heap 32 MB lighter. Decoded bitmap memory down 5×.

## Trade-off

Global eviction also clears avatars and images from other screens. Acceptable because `dispose()` fires post-navigation — the window where other screens are affected is minimal, and evicted images re-decode from disk cache without a network fetch.

## Long-term

`ImageProvider.evict()` per MXC URL, scoped per room. Requires tracking providers at the room level — noted in a TODO.

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -523,7 +523,7 @@ class ChatController extends State<Chat>
     if (!timeline!.canRequestHistory) return;
     Logs().v('Chat::requestHistory(): Requesting history...');
     try {
-      return timeline!.requestHistory(
+      await timeline!.requestHistory(
         historyCount: historyCount ?? _loadHistoryCount,
         filter: filter,
       );
@@ -1267,6 +1267,9 @@ class ChatController extends State<Chat>
   void scrollDown() async {
     if (timeline == null) return;
     if (!timeline!.allowNewEvent) {
+      // Cancel subscriptions BEFORE nulling the reference to avoid
+      // leaking the old Timeline's 5 stream subscriptions.
+      timeline!.cancelSubscriptions();
       setState(() {
         timeline = null;
         loadTimelineFuture = _getTimeline().onError((e, s) {
@@ -1318,6 +1321,9 @@ class ChatController extends State<Chat>
         );
         return;
       }
+      // Cancel subscriptions BEFORE nulling the reference to avoid
+      // leaking the old Timeline's 5 stream subscriptions.
+      timeline?.cancelSubscriptions();
       setState(() {
         timeline = null;
         loadTimelineFuture = _getTimeline(eventContextId: eventId).onError((
@@ -1607,6 +1613,9 @@ class ChatController extends State<Chat>
   }) async {
     Logs().d('$logContext: Reloading timeline centered on $eventId');
 
+    // Cancel subscriptions BEFORE nulling the reference to avoid
+    // leaking the old Timeline's 5 stream subscriptions.
+    timeline?.cancelSubscriptions();
     setState(() {
       timeline = null;
       loadTimelineFuture = _getTimeline(eventContextId: eventId).onError((
@@ -3690,6 +3699,23 @@ class ChatController extends State<Chat>
     disposeAutoMarkAsReadMixin();
     timeline?.cancelSubscriptions();
     timeline = null;
+
+    // Evict decoded images from Flutter's image cache on chat close.
+    //
+    // Trade-off: this is a global eviction — avatars and images from other
+    // screens are also cleared. However, Flutter's default LRU policy alone
+    // is insufficient here: a dense room (200+ messages with photos) can
+    // accumulate 50–100 decoded bitmaps well above 80MB before LRU starts
+    // evicting, since eviction only triggers when new images are added.
+    // On iOS (no swap), this causes OOM kills. The dispose() fires after
+    // navigation has already occurred, so the visibility window where other
+    // screens are affected is minimal.
+    //
+    // TODO: replace with per-room selective eviction via ImageProvider.evict()
+    // once MXC image providers are tracked at the room level.
+    PaintingBinding.instance.imageCache.clear();
+    PaintingBinding.instance.imageCache.clearLiveImages();
+
     inputFocus.dispose();
     searchEmojiFocusNode.dispose();
     composerDebouncer.cancel();


tddang-linagora (commented):
- [ ] Add before after screenshots of the improvement

lib/pages/chat/chat.dart
```diff
@@ -3690,6 +3726,13 @@ class ChatController extends State<Chat>
     disposeAutoMarkAsReadMixin();
     timeline?.cancelSubscriptions();
     timeline = null;
+
+    // Evict decoded images from Flutter's image cache to free GPU/raster
+    // memory. This is important because decoded bitmaps can consume
+    // significantly more memory than their compressed form.
+    PaintingBinding.instance.imageCache.clear();
+    PaintingBinding.instance.imageCache.clearLiveImages();
+
```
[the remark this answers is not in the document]
9clg6: The global eviction is intentional. Flutter's LRU alone is insufficient for a dense chat room: 200+ messages with photos decode to 50–100 bitmaps that can exceed 80MB before LRU kicks in (it only evicts when *new* images are added, not passively). On iOS there's no swap, so keeping that memory until the next image load means OOM kills in practice.

The `dispose()` fires after navigation has already happened, so the window where other screens are visibly affected is minimal — avatars in the list are being rebuilt anyway at that point.

The global clear is a deliberate compromise. Selective per-room eviction via `ImageProvider.evict()` would be the correct long-term fix and has been added as a TODO in the code.

lib/pages/chat/chat.dart
```diff
@@ -3690,6 +3726,23 @@ class ChatController extends State<Chat>
...
+    // evicting, since eviction only triggers when new images are added.
+    // On iOS (no swap), this causes OOM kills. The dispose() fires after
+    // navigation has already occurred, so the visibility window where other
+    // screens are affected is minimal.
+    //
+    // TODO: replace with per-room selective eviction via ImageProvider.evict()
+    // once MXC image providers are tracked at the room level.
+    PaintingBinding.instance.imageCache.clear();
+    PaintingBinding.instance.imageCache.clearLiveImages();
+
```
tddang-linagora, line 3718: Please let Flutter do its job. This will essentially make every cache image by Flutter load again, including the ones in the chat list.

lib/pages/chat/chat.dart
```diff
@@ -535,6 +542,26 @@ class ChatController extends State<Chat>
...
+    if (events == null || events.length <= _maxTimelineEvents) return;
+
+    final excess = events.length - _maxTimelineEvents;
+    Logs().d(
+      'Chat::_trimTimelineIfNeeded(): Trimming $excess old events '
+      '(${events.length} -> $_maxTimelineEvents)',
+    );
+    events.removeRange(_maxTimelineEvents, events.length);
+  }
+
```
tddang-linagora: Don't do this please. Imagine, you start loading a timeline in the middle of somewhere, and you keep loading the previous batch, exceeding 500 and some events are removed. Now you scroll to the opposite direction and load the next batch of your starting position, now all the removed event are missing for good until you exit the room and enter again.

lib/pages/chat/chat.dart
9clg6: "Letting Flutter do its job" means LRU, which only evicts on new insertions - never on idle. We measured 32 MB unreclaimed without this.

Re-decoded images come from disk cache, not the network, so the UX cost is a few-millisecond placeholder. Also, `dispose()` fires post-navigation: the chat list isn't visible yet when the clear happens.

9clg6: “Letting Flutter do the work” is clearly a solution that seems to work, given that performance is a major issue

lib/pages/chat/chat.dart
9clg6: You are right, I didn't think of this case

lib/pages/chat/chat.dart
tddang-linagora: From `clearLiveImages` doc:
>Calling this method does not relieve memory pressure, since the live image caching only tracks image instances that are also being held by at least one other object.

So calling this does not solve any memory issue.
And from `ImageCache` doc:
>Implements a least-recently-used cache of up to 1000 images, and up to 100 MB. The maximum size can be adjusted using [maximumSize] and [maximumSizeBytes].

So if you find the cache size of image is an issue, you can reduce `maximumSizeBytes`. It is not even recommended to call `imageCache`:
>The cache is used internally by [ImageProvider] and should generally not be accessed directly.

lib/pages/chat/chat.dart
```diff
@@ -3690,6 +3699,23 @@ class ChatController extends State<Chat>
...
+    // accumulate 50–100 decoded bitmaps well above 80MB before LRU starts
+    // evicting, since eviction only triggers when new images are added.
+    // On iOS (no swap), this causes OOM kills. The dispose() fires after
+    // navigation has already occurred, so the visibility window where other
+    // screens are affected is minimal.
+    //
+    // TODO: replace with per-room selective eviction via ImageProvider.evict()
+    // once MXC image providers are tracked at the room level.
+    PaintingBinding.instance.imageCache.clear();
+    PaintingBinding.instance.imageCache.clearLiveImages();
```
hoangdat, line 3717: https://github.com/user-attachments/assets/49426a39-e395-4e29-bd3c-fd5994b5eec0

When I close the chat screen to jump to profile screen, avatar in chat list will be effected. So we should keep the imageCache of Flutter instead of clear it our self.

9clg6: Only one image is flickering on navigate back: the avatar of the conversation you just came from, which was just cleared from cache. If there is a performance improvement of ~32 MB freed on iOS (no swap), I think this is an acceptable trade-off - especially since the image is re-decoded from disk cache, not re-fetched over the network.

lib/pages/chat/chat.dart
9clg6: On clearLiveImages(): The doc is correct - it doesn't relieve memory pressure on its own. But that's not its role here. Without clearLiveImages(), Flutter can re-promote live image references back into the LRU after clear() runs, partially undoing the eviction. The two calls are complementary, not redundant.

On maximumSizeBytes: This is a steady-state capacity control - it limits how large the cache grows going forward, but doesn't free anything already resident. A photo-heavy room can fill the cache to its limit and then stall: LRU only evicts on new insertions, not on idle. Lowering the cap globally doesn't free what's already resident in memory - it only limits future growth. The problem here is a spike caused by opening a photo-heavy room; clear() on dispose is the only way to actively reclaim that memory at the right moment.

On "do not access imageCache directly": That's a defensive note against arbitrary clearings at unpredictable moments. dispose() is a well-defined lifecycle boundary - the room is being torn down. That's exactly the right place for a targeted eviction.

The 32 MB measured freed is consistent with this model. Happy to revisit once we have per-room  ImageProvider.evict() in place as the long-term solution.

And @hoangdat 's video shows the only visible side effect: a single avatar flickering on navigate back, re-decoded from disk cache, not re-fetched over the network. The effect is minimal. On the other side, the gain is proven and measured (~32 MB freed on a photo-heavy room, critical on iOS where there is no swap). Given that the trade-off has been clearly demonstrated, I'm not sure what's left to debate here.

When leaving a room, we want to shed the residual memory associated with it. The images are still on disk cache. On iOS with no swap, keeping 80 MB of decoded bitmaps resident is a concrete OOM risk. The trade-off is clear.

tddang-linagora: This way you're disregarding any cache effort made by the Flutter framework to claim 100mb of memory at maximum. If you're fine with this then I have no other objection.

PR #2936, opened 2026-03-18
Config support email

hoangdat:
## Summary by CodeRabbit

* **New Features**
  * Support email is now configurable, allowing customization of the email address used in account deletion flows instead of using a hardcoded value.

* **Documentation**
  * Updated configuration documentation to include the new support email configuration field with examples for mobile and web platforms.

### DIFF
--- a/config.sample.json
+++ b/config.sample.json
@@ -17,5 +17,6 @@
   "qr_code_download_url": "https://sign-up.twake.app/download/chat",
   "enable_logs": false,
   "support_url": "https://twake.app/support",
+  "support_email": "support@twake.app",
   "cozy_external_bridge_version": "0.8.0"
 }
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,7 +1,7 @@
 name: fluffychat
 description: A convenient Matrix-based tool for personal and corporate communication.
 publish_to: none
-version: 2.20.4+2330
+version: 2.20.5+2330
 
 environment:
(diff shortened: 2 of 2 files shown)


lib/pages/settings_dashboard/settings/settings.dart
```diff
@@ -294,7 +294,7 @@ class SettingsController extends State<Settings> with ConnectPageMixin {
         }
 
         final emailUri = Uri.parse(
-          'mailto:software@linagora.com?subject=${Uri.encodeComponent(generateEmailSubject)}',
+          'mailto:${AppConfig.supportEmail}?subject=${Uri.encodeComponent(generateEmailSubject)}',
```
Crash--, line 297: can we at least add the matrix ID somewhere? Because we receive the email but this email can be sent from email address we don't know.

hoangdat: you mean we should put the matrixId of user in the subject of email to easily to identify the requester?

Crash--: or in the body with &body but yes. We need to have something to identify the user. The matrix id is ok, but if you can have access to the email address of the user, it'll be better.

hoangdat: address it here https://github.com/linagora/twake-on-matrix/pull/2937

PR #2950, opened 2026-03-20
perf: prevent iOS OOM kills with image download queue and decode resize

9clg6:
## Summary

Prevent iOS out-of-memory kills caused by unbounded concurrent image downloads when scrolling fast through rooms with many images. Two complementary mechanisms:

1. **Download queue** — limits concurrent image downloads/decodes to 4 via a semaphore in \`MxcImage._tryLoad()\`
2. **Decode resize** — adds \`cacheWidth\`/\`cacheHeight\` to \`ImageBubble\` so images are decoded at display resolution instead of native (~130x memory reduction per image)

### Changes

- **Add \`ImageDownloadQueue\`** (semaphore pattern, max 4 concurrent): Each \`MxcImage\` acquires a slot before loading and releases it when done or disposed. O(1) cancel for fast-scrolling scenarios where widgets appear and disappear in rapid succession.
- **Integrate queue into \`MxcImage._tryLoad()\`**: The actual image consumer — not preview URLs — is now throttled. Pending tickets are cancelled on dispose to prevent slot starvation.
- **Add \`cacheWidth\`/\`cacheHeight\` to \`ImageBubble\`**: Chat images are decoded at display resolution (e.g., 768px on 3x iPhone) instead of native (e.g., 4000px), reducing per-image memory from ~48 MB to ~360 KB.
- **13 unit tests** for \`ImageDownloadQueue\`: FIFO ordering, cancel, starvation prevention, stress test, configurable maxConcurrent, singleton management.

### How it works

```
MxcImage._tryLoad()
  ├─ imageData already in memory? → skip queue, render immediately
  ├─ _releaseDownloadTicket()     → cancel/release any previous ticket
  ├─ acquire() slot from queue    → wait if 4 downloads already running
  ├─ await ticket.ready           → proceeds when slot available
  ├─ !mounted guard               → release slot if widget disposed while waiting
  ├─ _load() → HTTP download + decrypt
  ├─ setState()                   → render image
  └─ _releaseDownloadTicket()     → free slot for next image
  
dispose()
  └─ _releaseDownloadTicket()     → no orphaned slots
```

### Known limitations / follow-up

- \`AvifImage\` does not receive \`cacheWidth\`/\`cacheHeight\` (AVIF is rare in Matrix, \`flutter_avif\` may not support it)
- No retry limit in \`_tryLoad\` — if server returns 500s, retries are unbounded (mitigated by slot release before retry)

---

## Performance results

Measured on a physical iPhone (release build) using a Patrol integration test (`image_scroll_perf_test`) that navigates to two image-heavy rooms ("Bug Me Harder" and "tech-radar"), scrolls for 30 seconds in each, and records `imageCache` size + process RSS at each step.

**Protocol:**
- Script: `scripts/integration_test_patrol.sh` (`patrol test --release --no-generate-bundle --show-flutter-logs`)
- Test: `integration_test/tests/chat/image_scroll_perf_test.dart`
- 3 runs per branch, **median** value retained
- Branches compared: `main` vs `perf/image-download-throttle`
- Integration test infrastructure enabled by #2980

### imageCache bytes (decoded image memory)

| Checkpoint | `main` | `perf/image-download-throttle` | Delta |
|---|---|---|---|
| initial lobby | 2.6 MB | 2.6 MB | — |
| room 1 entered | 15.2 MB | 9.8 MB | **-35%** |
| room 1 after 30s scroll | 24.8 MB | 18.3 MB | **-26%** |
| room 1 left | 4.3 MB | 2.4 MB | **-45%** |
| room 2 entered | 25.7 MB | 15.8 MB | **-39%** |
| room 2 after 30s scroll | 57.1 MB | 42.9 MB | **-25%** |
| room 2 left | 4.7 MB | 2.4 MB | **-50%** |
| final lobby | 4.7 MB | 2.4 MB | **-50%** |

### Process RSS (total memory)

| Checkpoint | `main` | `perf/image-download-throttle` | Delta |
|---|---|---|---|
| room 1 after 30s scroll | 262 MB | 246 MB | **-6%** |
| room 2 after 30s scroll | 292 MB | 278 MB | **-5%** |
| final lobby | 296 MB | 280 MB | **-5%** |

**The throttle reduces decoded image cache by 25–50% during scrolling and at rest, with a consistent ~5% reduction in total process memory.**

---

## Ticket

No ticket — discovered during performance profiling.

## Summary by CodeRabbit

* **Improvements**
  * Image downloads now execute in a controlled queue (max 4 concurrent operations), preventing system overload and improving app responsiveness.
  * Enhanced image caching with context-aware memory management for more efficient resource usage.

* **Tests**
  * Added comprehensive test coverage for image download queue functionality.

### DIFF
--- a/lib/pages/chat/events/images_builder/image_bubble.dart
+++ b/lib/pages/chat/events/images_builder/image_bubble.dart
@@ -1,5 +1,6 @@
 import 'dart:typed_data';
 
+import 'package:fluffychat/utils/extension/build_context_extension.dart';
 import 'package:fluffychat/pages/chat/events/images_builder/image_builder_web.dart';
 import 'package:fluffychat/pages/chat/events/images_builder/image_placeholder.dart';
 import 'package:fluffychat/pages/chat/events/message_content_style.dart';
@@ -108,6 +109,8 @@ class ImageBubble extends StatelessWidget {
                     event: event,
                     width: width,
                     height: height,
+                    cacheWidth: context.getCacheSize(width),
+                    cacheHeight: context.getCacheSize(height),
                     fit: fit,
                     animated: animated,
                     isThumbnail: thumbnailOnly,
--- a/lib/utils/image_download_queue.dart
+++ b/lib/utils/image_download_queue.dart
@@ -1,0 +1,130 @@
+import 'dart:async';
+import 'dart:collection';
+
+import 'package:flutter/foundation.dart';
+
+/// Limits the number of concurrent image download/decode operations
+/// to prevent memory spikes that cause iOS OOM kills.
+///
+/// Each [MxcImage] acquires a slot before starting its load and
+/// releases it when done (or when disposed). If all slots are taken,
+/// the request is queued and processed in FIFO order.
+class ImageDownloadQueue {
+  static ImageDownloadQueue _instance = ImageDownloadQueue._internal();
+  static ImageDownloadQueue get instance => _instance;
+
+  @visibleForTesting
+  static set instance(ImageDownloadQueue queue) => _instance = queue;
+
+  @visibleForTesting
+  static void resetInstance() {
+    _instance = ImageDownloadQueue._internal();
+  }
+
+  ImageDownloadQueue._internal({this.maxConcurrent = 4});
+
+  @visibleForTesting
+  factory ImageDownloadQueue.forTesting({int maxConcurrent = 4}) =>
+      ImageDownloadQueue._internal(maxConcurrent: maxConcurrent);
+
+  /// Maximum number of image loads running simultaneously.
+  /// 4 is a good balance: enough parallelism for fast scrolling,
+  /// low enough to avoid memory pressure on iOS.
+  final int maxConcurrent;
+
+  int _running = 0;
+  final Queue<_QueueEntry> _queue = Queue<_QueueEntry>();
+
+  /// Acquires a slot in the download queue.
+  ///
+  /// Returns an [ImageDownloadTicket] whose [ready] future completes
+  /// with `true` when a slot is available, or `false` if cancelled.
+  ///
+  /// The ticket must be released via [release] when work is finished,
+  /// or cancelled via [ImageDownloadTicket.cancel] if no longer needed.
+  ImageDownloadTicket acquire() {
+    final ticket = ImageDownloadTicket._();
+
+    if (_running < maxConcurrent) {
+      _running++;
+      ticket._granted = true;
+      ticket._completer.complete(true);
+    } else {
+      final entry = _QueueEntry(ticket);
+      _queue.add(entry);
+      // Entry held in queue; ticket uses cancel() to skip when drained.
+    }
+
+    return ticket;
+  }
+
+  /// Releases a slot, allowing the next queued request to proceed.
+  void release(ImageDownloadTicket ticket) {
+    if (!ticket._granted) {
+      // Was still queued — mark as cancelled and let _drainQueue skip it.
+      // This is O(1) instead of O(n) queue.remove().
+      ticket.cancel();
+      return;
+    }
+
+    ticket._granted = false;
+    _running--;
+
+    _drainQueue();
+  }
+
+  void _drainQueue() {
+    while (_running < maxConcurrent && _queue.isNotEmpty) {
+      final next = _queue.removeFirst();
+      if (next.ticket._cancelled) {
+        if (!next.ticket._completer.isCompleted) {
+          next.ticket._completer.complete(false);
+        }
+        continue;
+      }
+      _running++;
+      next.ticket._granted = true;
+      // Ticket granted; no longer in queue.
+      if (!next.ticket._completer.isCompleted) {
+        next.ticket._completer.complete(true);
+      }
+    }
+  }
+
+  /// Current number of active downloads (for debugging/testing).
+  int get activeCount => _running;
+
+  /// Current number of queued downloads (for debugging/testing).
+  /// Note: may include cancelled entries not yet drained.
+  int get queuedCount => _queue.length;
+}
+
+/// A ticket representing a pending or active slot in the download queue.
+class ImageDownloadTicket {
+  ImageDownloadTicket._();
+
+  final Completer<bool> _completer = Completer<bool>();
+  bool _granted = false;
+  bool _cancelled = false;
+
+  /// Resolves to `true` when a slot is available and work should start,
+  /// or `false` if the ticket was cancelled before being granted.
+  Future<bool> get ready => _completer.future;
+
+  /// Whether this ticket currently holds an active slot.
+  bool get isActive => _granted;
+
+  /// Cancel this ticket. If still queued, it will be skipped by drainQueue.
+  /// If already active, call [ImageDownloadQueue.release] instead.
+  void cancel() {
+    _cancelled = true;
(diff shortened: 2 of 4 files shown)


lib/widgets/mxc_image.dart
```diff
@@ -281,6 +317,7 @@ class _MxcImageState extends State<MxcImage>
 
   @override
   void dispose() {
+    _releaseDownloadTicket();
```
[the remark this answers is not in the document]
9clg6: This behavior is already handled. `_releaseDownloadTicket()` already distinguishes the two cases:

```dart
void _releaseDownloadTicket() {
  final ticket = _downloadTicket;
  if (ticket == null) return;
  if (ticket.isActive) {
    ImageDownloadQueue.instance.release(ticket);  // slot was granted — return it
  } else {
    ticket.cancel();  // still queued — cancel without touching _running
  }
  _downloadTicket = null;
}
```

`ticket.isActive` is checked at line 272 — queued tickets are cancelled, active ones are released. This was already in place before this PR's diff. The `isActive` getter you reference as missing exists on `DownloadTicket`.

lib/pages/chat/events/images_builder/image_bubble.dart
```diff
@@ -108,6 +109,8 @@ class ImageBubble extends StatelessWidget {
                     event: event,
                     width: width,
                     height: height,
+                    cacheWidth: context.getCacheSize(width),
+                    cacheHeight: context.getCacheSize(height),
```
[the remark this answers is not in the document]
9clg6: Valid point. `cacheWidth`/`cacheHeight` are currently passed unconditionally, which bypasses `MxcImage`'s own `noResize` policy for callers using `noResizeThumbnail: true`.

That said, this is already addressed on the `perf/image-cache-dimensions` branch (linagora/twake-on-matrix#2980), which introduced the `cacheWidth`/`cacheHeight` plumbing in `MxcImage`. The fix there gates the resize on `noResize`. This PR will be rebased on top of that one before merge, so the behavior will be correct end-to-end.

lib/utils/image_download_queue.dart
```diff
@@ -0,0 +1,130 @@
...
+  @visibleForTesting
+  static void resetInstance() {
+    _instance = ImageDownloadQueue._internal();
+  }
+
+  ImageDownloadQueue._internal({this.maxConcurrent = 4});
+
+  @visibleForTesting
+  factory ImageDownloadQueue.forTesting({int maxConcurrent = 4}) =>
+      ImageDownloadQueue._internal(maxConcurrent: maxConcurrent);
```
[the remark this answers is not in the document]
9clg6: Agreed, adding a fast-fail guard is the right call. Fixed:

```dart
ImageDownloadQueue._internal({this.maxConcurrent = 4}) {
  if (maxConcurrent < 1) {
    throw ArgumentError.value(
      maxConcurrent,
      'maxConcurrent',
      'Must be greater than 0',
    );
  }
}
```

jbrnzl (approved):

lib/widgets/mxc_image.dart
```diff
@@ -221,19 +223,53 @@ class _MxcImageState extends State<MxcImage>
...
       _imageData = loadResult.imageData;
       filePath = loadResult.filePath;
-      setState(() {});
+      if (mounted) setState(() {});
     } catch (e) {
       if (mounted && !isLoadDone) {
         isLoadDone = true;
+        // Release before retry so the retry can acquire its own slot.
+        _releaseDownloadTicket();
         _tryLoad(context);
```
hoangdat: Retry sets isLoadDone = true. If retry also fails, cleanup is skipped. Slot permanently leaked.

lib/utils/image_download_queue.dart
```diff
@@ -0,0 +1,130 @@
...
+
+  @visibleForTesting
+  static set instance(ImageDownloadQueue queue) => _instance = queue;
+
+  @visibleForTesting
+  static void resetInstance() {
+    _instance = ImageDownloadQueue._internal();
+  }
+
+  ImageDownloadQueue._internal({this.maxConcurrent = 4});
```
hoangdat, line 24: how about in case of 4 heavy download in room A, then we swtich to room B? user should wait for  the download in room A completed then room B?

lib/widgets/mxc_image.dart
```diff
@@ -221,19 +223,53 @@ class _MxcImageState extends State<MxcImage>
...
+    if (!shouldProceed || !mounted) {
+      if (shouldProceed) {
+        ImageDownloadQueue.instance.release(ticket);
+        _downloadTicket = null;
+      }
+      return;
     }
+
     try {
       final loadResult = await _load(context);
```
hoangdat: I also concern this method still take a long time to response, especially the download from server? maybe we need timeout to ensure it not take the ticket permanent?

lib/widgets/mxc_image.dart
9clg6: Good catch on the readability — the flow was correct (the `_releaseDownloadTicket()` after the try/catch was always reached on the retry-failure path because the catch doesn't rethrow), but relying on fallthrough made it hard to reason about.

Moved the release into a `finally` block in a25fa8b so the guarantee is now structural. `_releaseDownloadTicket()` is idempotent (no-op if `_downloadTicket` is already null), so the early release + return in the catch path still works fine alongside the finally.

lib/utils/image_download_queue.dart
9clg6: When switching from room A to room B, the MxcImage widgets in room A are disposed — `dispose()` calls `_releaseDownloadTicket()`, which immediately releases active slots and cancels queued ones. So the 4 slots free up right away and room B images start loading without waiting.

That's the main reason tickets are tied to the widget lifecycle: images that are no longer visible never hold slots.

lib/widgets/mxc_image.dart
9clg6: The slot can't be held permanently in practice:
- The HTTP download already has its own timeout at the Matrix SDK level
- If `_load()` throws (network timeout, server error…), the `finally` block releases the ticket
- If the widget is disposed during a slow download, `dispose()` releases it immediately

Adding a separate timeout at the queue level would add complexity (which duration? what if the download was almost done?) for a case that's already covered by existing layers. If we ever see slots getting stuck in production we can revisit, but the current safeguards should be sufficient.

lib/widgets/mxc_image.dart
```diff
@@ -221,19 +223,53 @@ class _MxcImageState extends State<MxcImage>
...
       if (mounted && !isLoadDone) {
         isLoadDone = true;
+        // Release before retry so the retry can acquire its own slot.
+        _releaseDownloadTicket();
         _tryLoad(context);
+        return;
       }
+    } finally {
+      _releaseDownloadTicket();
+    }
```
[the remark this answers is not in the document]
9clg6: Good catch — the recursive call was indeed problematic with the `finally` block. The recursive `_tryLoad()` runs synchronously until its first `await`, storing a new ticket in `_downloadTicket`. Then the original call's `finally` releases *that* ticket, not the original one.

Replaced the recursion with a `do/while` loop in 590f8ae so the ticket lifecycle is deterministic within a single `finally` block. Each iteration acquires, uses, and releases its own ticket cleanly.

hoangdat (approved):

Issue #2963, opened 2026-03-27
Forward message to several people


tddang-linagora:
# Description

We need to change the user picker when forwarding a message/media from single-select to multi-select. If user selects several addressees, the forwarded message is sent separately to each of them 

<img width="1342" height="710" alt="Image" src="https://github.com/user-attachments/assets/8318ee94-164f-4d95-bce9-1f0555e3b5cc" />

After I select users and click “Send”, I end up un the same location from where I forwarded the message (see tg behaviour)

https://github.com/user-attachments/assets/022dd14a-fc8b-4d5b-bd5c-43aa76594949

PR #2964, opened 2026-03-27
TW-2963 Forward to multiple rooms

tddang-linagora:
## Ticket
- #2963 

## Resolved

- Web:

https://github.com/user-attachments/assets/e5c544b6-7d92-4052-a7a1-96cd9cb5b481

- Mobile

https://github.com/user-attachments/assets/74ce271f-d6cc-4b28-809e-50020d9d0e62

## Summary by CodeRabbit

* **New Features**
  * Multi-select forwarding UI with checkboxes, a new recent-chats list, per-room forwarding, and an aggregated completion summary that informs navigation and snackbars.

* **Bug Fixes**
  * Per-target error handling so one failure won’t stop other forwards; improved loading/dialog behavior, guarded stream cancellation, and guaranteed share-state cleanup.

* **Localization**
  * Added English messages for success, partial success (counts), and failure.

* **Tests**
  * New end-to-end tests for single- and multi-recipient forwarding, search/select, and send flows.

### DIFF
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -26,13 +26,14 @@ class ForwardMessageInteractor {
           );
           if (room == null || room.membership != Membership.join) continue;
 
+          final hasMultipleMessages = matrixState.shareContentList.any(
+            (message) => message != null,
+          );
           final hasContent =
-              matrixState.shareContent != null ||
-              matrixState.shareContentList.isNotEmpty;
+              matrixState.shareContent != null || hasMultipleMessages;
           if (!hasContent) continue;
 
-          if (matrixState.shareContentList.isEmpty &&
-              matrixState.shareContent != null) {
+          if (!hasMultipleMessages && matrixState.shareContent != null) {
             yield* _forwardOneMessageAction(
               room: room,
               matrixState: matrixState,
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -26,6 +26,11 @@ class ForwardMessageInteractor {
           );
           if (room == null || room.membership != Membership.join) continue;
 
+          final hasContent =
+              matrixState.shareContent != null ||
+              matrixState.shareContentList.isNotEmpty;
+          if (!hasContent) continue;
+
           if (matrixState.shareContentList.isEmpty &&
               matrixState.shareContent != null) {
             yield* _forwardOneMessageAction(
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -45,9 +45,6 @@ class ForwardMessageInteractor {
         }
       }
 
-      matrixState.shareContent = null;
-      matrixState.shareContentList = null;
-
       yield Right(
         ForwardMessageAllDoneState(
           successCount: successCount,
@@ -56,6 +53,9 @@ class ForwardMessageInteractor {
       );
     } catch (exception) {
       yield Left(ForwardMessageFailed(exception: exception));
+    } finally {
+      matrixState.shareContent = null;
+      matrixState.shareContentList = null;
     }
   }
 
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -21,6 +21,11 @@ class ForwardMessageInteractor {
       final hasMultipleMessages = messages.any((message) => message != null);
       final hasContent = singleMessage != null || hasMultipleMessages;
 
+      if (!hasContent) {
+        yield Left(ForwardEmptyContentFailure());
+        return;
+      }
+
       yield Right(ForwardMessageLoading());
 
       int successCount = 0;
@@ -32,8 +37,6 @@ class ForwardMessageInteractor {
             (element) => element.id == roomId,
           );
           if (room == null || room.membership != Membership.join) continue;
-
-          if (!hasContent) continue;
 
           if (!hasMultipleMessages && singleMessage != null) {
             yield* _forwardOneMessageAction(room: room, message: singleMessage);
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -15,20 +15,42 @@ class ForwardMessageInteractor {
     try {
       yield Right(ForwardMessageLoading());
 
-      final room = rooms.firstWhere(
-        (element) => element.id == selectedEvents.first,
-      );
-      if (room.membership == Membership.join) {
-        if (matrixState.shareContentList.isEmpty &&
-            matrixState.shareContent != null) {
-          yield* _forwardOneMessageAction(room: room, matrixState: matrixState);
-        } else {
-          yield* _forwardMultipleMessagesAction(
-            room: room,
-            matrixState: matrixState,
-          );
+      int successCount = 0;
+      final totalCount = selectedEvents.length;
+
+      for (final roomId in selectedEvents) {
+        try {
+          final room = rooms.firstWhere((element) => element.id == roomId);
+          if (room.membership != Membership.join) continue;
+
+          if (matrixState.shareContentList.isEmpty &&
+              matrixState.shareContent != null) {
+            yield* _forwardOneMessageAction(
+              room: room,
+              matrixState: matrixState,
+            );
+          } else {
+            yield* _forwardMultipleMessagesAction(
+              room: room,
+              matrixState: matrixState,
+            );
+          }
+          successCount++;
+          yield Right(ForwardMessageSuccess(room));
+        } catch (exception) {
+          yield Left(ForwardMessageFailed(exception: exception));
         }
       }
+
+      matrixState.shareContent = null;
+      matrixState.shareContentList = null;
+
+      yield Right(
+        ForwardMessageAllDoneState(
+          successCount: successCount,
+          totalCount: totalCount,
+        ),
+      );
     } catch (exception) {
       yield Left(ForwardMessageFailed(exception: exception));
     }
@@ -53,15 +75,9 @@ class ForwardMessageInteractor {
     required Room room,
     required MatrixState matrixState,
   }) async* {
-    try {
-      final message = matrixState.shareContent;
-      if (message != null) {
-        yield* _forwardMessage(message, room);
-        matrixState.shareContent = null;
-      }
-      yield Right(ForwardMessageSuccess(room));
-    } catch (exception) {
-      yield Left(ForwardMessageFailed(exception: exception));
+    final message = matrixState.shareContent;
+    if (message != null) {
+      yield* _forwardMessage(message, room);
     }
   }
 
@@ -69,23 +85,13 @@ class ForwardMessageInteractor {
     required Room room,
     required MatrixState matrixState,
   }) async* {
-    try {
-      yield Right(ForwardMessageLoading());
-
-      final messages = matrixState.shareContentList;
-      if (messages.isNotEmpty) {
-        for (final message in messages) {
-          if (message != null) {
-            yield* _forwardMessage(message, room);
-          } else {
-            continue;
-          }
+    final messages = matrixState.shareContentList;
+    if (messages.isNotEmpty) {
+      for (final message in messages) {
+        if (message != null) {
+          yield* _forwardMessage(message, room);
         }
-        matrixState.shareContentList = null;
       }
-      yield Right(ForwardMessageSuccess(room));
-    } catch (exception) {
-      yield Left(ForwardMessageFailed(exception: exception));
     }
   }
 }
--- a/lib/domain/usecase/forward/forward_message_interactor.dart
+++ b/lib/domain/usecase/forward/forward_message_interactor.dart
@@ -14,6 +14,13 @@ class ForwardMessageInteractor {
     required MatrixState matrixState,
   }) async* {
     try {
+      final singleMessage = matrixState.shareContent;
+      final messages = List<Map<String, dynamic>?>.from(
+        matrixState.shareContentList,
+      );
+      final hasMultipleMessages = messages.any((message) => message != null);
+      final hasContent = singleMessage != null || hasMultipleMessages;
+
       yield Right(ForwardMessageLoading());
 
       int successCount = 0;
@@ -26,22 +33,14 @@ class ForwardMessageInteractor {
           );
           if (room == null || room.membership != Membership.join) continue;
 
-          final hasMultipleMessages = matrixState.shareContentList.any(
-            (message) => message != null,
-          );
-          final hasContent =
-              matrixState.shareContent != null || hasMultipleMessages;
           if (!hasContent) continue;
 
-          if (!hasMultipleMessages && matrixState.shareContent != null) {
-            yield* _forwardOneMessageAction(
-              room: room,
-              matrixState: matrixState,
-            );
+          if (!hasMultipleMessages && singleMessage != null) {
+            yield* _forwardOneMessageAction(room: room, message: singleMessage);
           } else {
             yield* _forwardMultipleMessagesAction(
               room: room,
-              matrixState: matrixState,
+              messages: messages,
             );
           }
           successCount++;
@@ -82,24 +81,18 @@ class ForwardMessageInteractor {
 
   Stream<Either<Failure, Success>> _forwardOneMessageAction({
     required Room room,
-    required MatrixState matrixState,
+    required Map<String, dynamic> message,
   }) async* {
-    final message = matrixState.shareContent;
-    if (message != null) {
-      yield* _forwardMessage(message, room);
-    }
+    yield* _forwardMessage(message, room);
   }
 
   Stream<Either<Failure, Success>> _forwardMultipleMessagesAction({
     required Room room,
-    required MatrixState matrixState,
+    required List<Map<String, dynamic>?> messages,
   }) async* {
-    final messages = matrixState.shareContentList;
-    if (messages.isNotEmpty) {
-      for (final message in messages) {
-        if (message != null) {
-          yield* _forwardMessage(message, room);
-        }
+    for (final message in messages) {
+      if (message != null) {
+        yield* _forwardMessage(message, room);
       }
     }
   }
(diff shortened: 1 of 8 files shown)


lib/pages/forward/forward.dart
```diff
@@ -114,35 +122,66 @@ class ForwardController extends State<Forward>
...
+  }
+
+  String _buildMultiForwardSnackbarMessage(ForwardMessageAllDoneState done) {
+    if (done.successCount == 0) {
+      return 'Failed to forward message';
+    }
+    if (done.successCount == done.totalCount) {
+      return 'Forwarded to ${done.successCount} chats';
+    }
+    return 'Forwarded to ${done.successCount} of ${done.totalCount} chats';
```
jbrnzl: I think we will need some localization here

lib/pages/forward/forward.dart
```diff
@@ -81,19 +83,25 @@ class ForwardController extends State<Forward>
...
+    selectedRoomIdNotifier.value = current;
   }
 
   final ActiveFilter _activeFilterAllChats = ActiveFilter.acceptedChats;
 
   List<Room> get filteredRoomsForAll =>
       Matrix.of(context).client.filteredRoomsForAll(_activeFilterAllChats);
 
-  void forwardAction() async {
+  void forwardAction(BuildContext context) async {
```
jbrnzl: @tddang-linagora can you please confirm if passing the context here is needed ?

lib/domain/usecase/forward/forward_message_interactor.dart
```diff
@@ -15,20 +15,42 @@ class ForwardMessageInteractor {
...
-          yield* _forwardOneMessageAction(room: room, matrixState: matrixState);
-        } else {
-          yield* _forwardMultipleMessagesAction(
-            room: room,
-            matrixState: matrixState,
-          );
+      int successCount = 0;
+      final totalCount = selectedEvents.length;
+
+      for (final roomId in selectedEvents) {
```
hoangdat, line 34: can we do it in parallel?

lib/domain/usecase/forward/forward_message_interactor.dart
```diff
@@ -15,20 +15,42 @@ class ForwardMessageInteractor {
...
+        try {
+          final room = rooms.firstWhere((element) => element.id == roomId);
```
hoangdat: how about `firstWhereOrNull`?

lib/pages/forward/forward.dart
```diff
@@ -114,35 +122,66 @@ class ForwardController extends State<Forward>
...
-              ),
+              builder: (c) => SendFileDialog(files: [shareFile], room: room),
             );
             break;
         }
       },
     );
   }
 
+  void _handleAllDone(BuildContext context, ForwardMessageAllDoneState done) {
```
hoangdat: Guard: if all forwards failed for a single-room selection, show error instead of navigating — avoids silent failure UX.

lib/domain/usecase/forward/forward_message_interactor.dart
nqhhdev: `_forwardMessage` yields `ForwardMessageIsShareFileState` which triggers a dialog in the UI — running that in parallel would open multiple dialogs simultaneously. Matrix SDK also queues sendEvent internally per room, so the perf gain is minimal. Sequential is safer here.

nqhhdev:
@coderabbitai The getter shareContentList already returns _shareContentList ?? [] — it's non-nullable by contract. The null checks in the proposed fix are unnecessary. Current code is safe. 

> ♻️ Duplicate comments (1)
> > lib/domain/usecase/forward/forward_message_interactor.dart (1)> `29-30`: _⚠️ Potential issue_ | _🔴 Critical_
> > > **Guard `shareContentList` before calling `.isEmpty` / `.isNotEmpty`.**
> > > Line 29 and Line 92 dereference `matrixState.shareContentList`, but the same flow resets that field to `null` on Line 49. If this batch is invoked with only `shareContent` populated, forwarding can fail before the first send.
> > > 
> > > Proposed fix
> > > ```diff
> > > -          if (matrixState.shareContentList.isEmpty &&
> > > -              matrixState.shareContent != null) {
> > > +          final messages = matrixState.shareContentList;
> > > +          if ((messages == null || messages.isEmpty) &&
> > > +              matrixState.shareContent != null) {
> > >              yield* _forwardOneMessageAction(
> > >                room: room,
> > >                matrixState: matrixState,
> > >              );
> > >            } else {
> > > @@
> > > -    final messages = matrixState.shareContentList;
> > > -    if (messages.isNotEmpty) {
> > > +    final messages = matrixState.shareContentList;
> > > +    if (messages != null && messages.isNotEmpty) {
> > >        for (final message in messages) {
> > >          if (message != null) {
> > >            yield* _forwardMessage(message, room);
> > >          }
> > >        }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > Use this to confirm the field type and nearby call sites:
> > > ```shell
> > > #!/bin/bash
> > > set -e
> > > rg -n -C3 '\bclass MatrixState\b|\bshareContentList\b|\bshareContent\b' lib/widgets/matrix.dart lib
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Also applies to: 91-92
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@lib/domain/usecase/forward/forward_message_interactor.dart` around lines 29 -
> > > 30, Guard accesses to matrixState.shareContentList in forward logic by
> > > null-checking it before calling .isEmpty / .isNotEmpty: update the checks in
> > > forward_message_interactor (where matrixState.shareContentList is evaluated
> > > around the block with matrixState.shareContent) to first ensure
> > > matrixState.shareContentList != null (or use ?.isEmpty/?.isNotEmpty) so the flow
> > > that sets shareContent and clears shareContentList to null won’t cause a
> > > null-deref; apply the same null-safe guard at the other occurrence around line
> > > 91-92 (the forwardAll/forward loop) so both code paths handle a null
> > > shareContentList correctly and still process a non-null shareContent.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Duplicate comments:
> In `@lib/domain/usecase/forward/forward_message_interactor.dart`:
> - Around line 29-30: Guard accesses to matrixState.shareContentList in forward
> logic by null-checking it before calling .isEmpty / .isNotEmpty: update the
> checks in forward_message_interactor (where matrixState.shareContentList is
> evaluated around the block with matrixState.shareContent) to first ensure
> matrixState.shareContentList != null (or use ?.isEmpty/?.isNotEmpty) so the flow
> that sets shareContent and clears shareContentList to null won’t cause a
> null-deref; apply the same null-safe guard at the other occurrence around line
> 91-92 (the forwardAll/forward loop) so both code paths handle a null
> shareContentList correctly and still process a non-null shareContent.
> ```
> 
> ℹ️ Review info

jbrnzl:
I know it is not only related to multiple rooms (as the same bug is here when forwarding to only one room) but when I forward a sticker, it says the message has been forwarded but in reality there is an error. (Also it tries to forward sticker description only). When I click retry, then the message is sent.

https://github.com/user-attachments/assets/c2b43bed-d8c1-4e0d-b73b-a9a63fe538af

Maybe we should at least not show the success message in this case ? Or remove the ability to forward a sticker for now

nqhhdev:
> I know it is not only related to multiple rooms (as the same bug is here when forwarding to only one room) but when I forward a sticker, it says the message has been forwarded but in reality there is an error. (Also it tries to forward sticker description only). When I click retry, then the message is sent.
> 
>  Enregistrement.de.l.ecran.2026-04-15.a.09.03.27.mov 
> Maybe we should at least not show the success message in this case ? Or remove the ability to forward a sticker for now

Thanks for your contribution. I have a few concerns from my side:

- Does this also occur when sending a message manually? If so, could you try to reproduce it using the normal send message flow?
- I believe this issue is more related to the message-sending flow itself. Could we create a separate ticket with reproduction steps and continue the discussion there?

cc: @jbrnzl

lib/pages/forward/forward.dart
```diff
@@ -93,7 +107,7 @@ class ForwardController extends State<Forward>
     forwardMessageInteractorStreamSubscription = _forwardMessageInteractor
```
hoangdat, line 119: should be cancel befor execute the new one. Fx: user double tap on forward button

lib/domain/usecase/forward/forward_message_interactor.dart
```diff
@@ -15,20 +16,44 @@ class ForwardMessageInteractor {
...
+          }
+          successCount++;
+          yield Right(ForwardMessageSuccess(room));
+        } catch (exception) {
+          yield Left(ForwardMessageFailed(exception: exception));
         }
       }
+
+      matrixState.shareContent = null;
+      matrixState.shareContentList = null;
```
hoangdat: should put in `finally` to ensure it is cleanup

lib/pages/forward/forward.dart
```diff
@@ -75,13 +78,24 @@ class ForwardController extends State<Forward>
     disposeContactsMixin();
     recentChatScrollController.dispose();
     forwardMessageInteractorStreamSubscription?.cancel();
+    // Clear share state in case the stream was cancelled mid-operation
+    // and the interactor's own cleanup lines never executed.
+    final matrixState = Matrix.of(context);
```
hoangdat: should wrap it in try/catch?

lib/pages/forward/forward.dart
```diff
@@ -114,35 +128,79 @@ class ForwardController extends State<Forward>
...
+  void _handleAllDone(BuildContext context, ForwardMessageAllDoneState done) {
+    if (!mounted) return;
+    if (done.totalCount != 1) {
+      if (Navigator.of(context).canPop()) {
+        Navigator.of(context).pop();
+      } else {
+        const RoomsRoute().go(context);
+      }
+      final message = _buildMultiForwardSnackbarMessage(context, done);
+      TwakeSnackBar.show(context, message);
```
hoangdat, line 185: we pop before, what is the context here?

jbrnzl:
> > I know it is not only related to multiple rooms (as the same bug is here when forwarding to only one room) but when I forward a sticker, it says the message has been forwarded but in reality there is an error. (Also it tries to forward sticker description only). When I click retry, then the message is sent.
> > Enregistrement.de.l.ecran.2026-04-15.a.09.03.27.mov
> > Maybe we should at least not show the success message in this case ? Or remove the ability to forward a sticker for now
> 
> Thanks for your contribution. I have a few concerns from my side:
> 
> * Does this also occur when sending a message manually? If so, could you try to reproduce it using the normal send message flow?
> * I believe this issue is more related to the message-sending flow itself. Could we create a separate ticket with reproduction steps and continue the discussion there?
> 
> cc: @jbrnzl

From what I know you can not send directly a sticker from Twake chat. Only receive. But you can forward it. I will create a seperate issue.

lib/pages/forward/forward_view.dart
```diff
@@ -3,7 +3,7 @@ import 'package:fluffychat/app_state/failure.dart';
 import 'package:fluffychat/app_state/success.dart';
 import 'package:fluffychat/domain/app_state/forward/forward_message_state.dart';
 import 'package:fluffychat/pages/forward/forward.dart';
-import 'package:fluffychat/pages/forward/recent_chat_list.dart';
```
9clg6, line 6: recent_chat_list.dart is not used anymore no ? delete the file if the class is dead code. Else np

9clg6:
Are Integration tests possible ? if yes please add

lib/pages/forward/forward_view.dart
nqhhdev: The file is still alive — lib/pages/share/share_view.dart imports it. Not dead code.

nqhhdev:
> Are Integration tests possible ? if yes please add

Screenshot 2026-04-17 at 12 30 27

Screenshot 2026-04-17 at 12 36 01

cc: @9clg6

lib/pages/forward/forward.dart
```diff
@@ -114,35 +135,79 @@ class ForwardController extends State<Forward>
...
         }
       },
     );
   }
 
+  void _handleAllDone(BuildContext context, ForwardMessageAllDoneState done) {
+    if (!mounted) return;
+    if (done.totalCount != 1) {
+      final message = _buildMultiForwardSnackbarMessage(context, done);
+      if (Navigator.of(context).canPop()) {
```
9clg6: You can refacto here to avoid `Navigator.of(context)` duplication

lib/pages/forward/forward.dart
```diff
@@ -114,35 +135,79 @@ class ForwardController extends State<Forward>
...
+      RoomRoute(roomid: roomId).go(context);
+    }
+  }
+
+  String _buildMultiForwardSnackbarMessage(
+    BuildContext context,
+    ForwardMessageAllDoneState done,
+  ) {
+    if (done.successCount == 0) {
+      return L10n.of(context)!.forwardFailed;
```
9clg6: Same

lib/pages/forward/forward.dart
```diff
@@ -114,35 +135,79 @@ class ForwardController extends State<Forward>
...
-              ),
+              builder: (c) => SendFileDialog(files: [shareFile], room: room),
             );
             break;
         }
       },
     );
   }
 
+  void _handleAllDone(BuildContext context, ForwardMessageAllDoneState done) {
```
9clg6: Don 't pass the BuildContext in parameters, it should stay in the screen. Prefer to pass a Callback instead

lib/pages/forward/forward.dart
```diff
@@ -114,35 +135,79 @@ class ForwardController extends State<Forward>
...
+
+  String _buildMultiForwardSnackbarMessage(
+    BuildContext context,
+    ForwardMessageAllDoneState done,
+  ) {
+    if (done.successCount == 0) {
+      return L10n.of(context)!.forwardFailed;
+    }
+    if (done.successCount == done.totalCount) {
+      return L10n.of(context)!.forwardedToChats(done.successCount);
```
9clg6: Refacto to avoid L10 duplication et same as above; don't pass the BuildContext in parameters

lib/pages/forward/forward_recent_chat_list.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  final ScrollController recentChatScrollController;
+
+  final ValueNotifier<List<String>> selectedChatNotifier;
+
+  final void Function(String roomId) onSelectedChat;
+
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<List<String>>(
```
9clg6: Moving the ValueListenableBuilder further down the tree would be more efficient here — currently every visible item is rebuilt on selection change, even though only the Radio depends on selectedChat. Wrapping just the Radio (or using the child: param of ValueListenableBuilder) scopes the rebuild to what actually changes.

lib/pages/forward/forward_view.dart
```diff
@@ -84,13 +84,13 @@ class ForwardView extends StatelessWidget {
             _WebActionsButton(
               selectedChatNotifier: controller.selectedRoomIdNotifier,
               forwardMessageNotifier: controller.forwardMessageNotifier,
-              forwardAction: controller.forwardAction,
+              forwardAction: () => controller.forwardAction(),
```
9clg6: Let the tearoff

lib/pages/forward/forward_view.dart
```diff
@@ -84,13 +84,13 @@ class ForwardView extends StatelessWidget {
...
             ),
         ],
       ),
       floatingActionButton: controller.isFullScreen
           ? _ForwardButton(
-              forwardAction: controller.forwardAction,
+              forwardAction: () => controller.forwardAction(),
```
9clg6: Same

lib/pages/forward/forward_view.dart
```diff
@@ -116,9 +116,9 @@ class _WebActionsButton extends StatelessWidget {
   Widget build(BuildContext context) {
     return Padding(
       padding: ForwardViewStyle.webActionsButtonPadding,
-      child: ValueListenableBuilder<String>(
+      child: ValueListenableBuilder<List<String>>(
```
9clg6: Surround the changing widget with the ValueListenableBuilder; not the whole widget

lib/pages/forward/forward_view.dart
9clg6: Else not necessary here, remove brackets

lib/pages/forward/forward_view.dart
```diff
@@ -116,78 +116,73 @@ class _WebActionsButton extends StatelessWidget {
...
+              message: L10n.of(context)!.cancel,
+              borderHover: ForwardViewStyle.webActionsButtonBorder,
+              margin: ForwardViewStyle.webActionsButtonMargin,
+              buttonDecoration: BoxDecoration(
+                borderRadius: BorderRadius.circular(
+                  ForwardViewStyle.webActionsButtonBorder,
+                ),
+              ),
+              styleMessage: Theme.of(context).textTheme.labelLarge?.copyWith(
+                color: LinagoraSysColors.material().primary,
```
9clg6: LinagoraSysColors can be refacto to avoid duplication

lib/pages/forward/forward_view.dart
```diff
@@ -116,78 +116,73 @@ class _WebActionsButton extends StatelessWidget {
...
-                        ? LinagoraSysColors.material().primary
-                        : LinagoraStateLayer(
-                            LinagoraSysColors.material().onSurface,
-                          ).opacityLayer2,
-                    borderRadius: BorderRadius.circular(
-                      ForwardViewStyle.webActionsButtonBorder,
-                    ),
+              );
+            }
+            return const SizedBox();
```
9clg6: Not blocking but SizedBox.shrink > SizedBox

lib/pages/forward/forward_view.dart
```diff
@@ -116,78 +116,73 @@ class _WebActionsButton extends StatelessWidget {
...
-                      ForwardViewStyle.webActionsButtonBorder,
-                    ),
+      child: ValueListenableBuilder<Either<Failure, Success>?>(
+        valueListenable: forwardMessageNotifier,
+        builder: (context, forwardMessageState, child) {
+          if (forwardMessageState == null) {
+            return child!;
+          }
+          return forwardMessageState.fold((failure) => child!, (success) {
+            if (success is ForwardMessageLoading) {
```
9clg6, line 127: Not blocking but if the condition has only 1 term, as success is ForwardMessageLoading and a else, prefer inversing condition as if(success !is ForwardMessageLoading) return SizedBox.shrink(); = early return pattern.
It improve code visibility

9clg6 (approved):

jbrnzl (approved):

PR #2982, opened 2026-04-02
HOTFIX Add retry to NSE get notification

tddang-linagora:
## Ticket
Some encrypted rooms' notifications are still unable to be decrypted.

## Root cause
- Room key download on Matrix Rust SDK is an asynchronous fire and forget operation
- There's no way to pre-feed the room key to Matrix Rust SDK

## Solution
Add exponential back-off retry, maximum 2 times, to give Rust SDK more chance to get the room key and decrypt the encrypted notification

## Impact description
Notification might take longer to appear.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable notification retrieval with a short automatic retry sequence (up to 3 attempts with incremental delays) to reduce missed notifications on slow or unstable connections.
  * Improved detection of end-to-end encrypted notifications so encrypted messages are correctly identified and excluded from regular notification handling.

### DIFF
--- a/ios/NSE/Sources/Other/NSEUserSession.swift
+++ b/ios/NSE/Sources/Other/NSEUserSession.swift
@@ -65,10 +65,11 @@ final class NSEUserSession {
         var proxy: NotificationItemProxyProtocol? = await fetchNotificationItem(roomID: roomID, eventID: eventID)
         if let proxy, !proxy.isEncrypted { return proxy }
 
-        for delay in 1...3 {
-            MXLog.info("NSE: Notification is \(proxy == nil ? "nil" : "encrypted"), retrying in \(delay)s - roomID: \(roomID)")
-            try? await Task.sleep(nanoseconds: UInt64(delay) * 1_000_000_000)
-            
+        for attempt in 0..<2 {
+            let delaySeconds = 1 << attempt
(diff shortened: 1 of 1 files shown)


ios/NSE/Sources/Other/NSEUserSession.swift
```diff
@@ -62,6 +62,20 @@ final class NSEUserSession {
     }
     
     func notificationItemProxy(roomID: String, eventID: String) async -> NotificationItemProxyProtocol? {
+        var proxy: NotificationItemProxyProtocol? = await fetchNotificationItem(roomID: roomID, eventID: eventID)
+        if let proxy, !proxy.isEncrypted { return proxy }
+
```
hoangdat, line 67: how about add this line

```
guard proxy != nil else { return nil } 
```

ios/NSE/Sources/Other/NSEUserSession.swift
```diff
@@ -62,6 +62,20 @@ final class NSEUserSession {
...
+        for delay in 1...3 {
+            try? await Task.sleep(nanoseconds: UInt64(delay) * 1_000_000_000)
+
```
hoangdat: add log
```
MXLog.info("NSE: Notification is encrypted, retrying in \(delay)s - roomID: \(roomID)")
```

Issue #2983, opened 2026-04-02
[Bug] [WEB] 2nd, 3rd etc video is only displayed in the chat after it’s uploaded


nqhhdev:

tprudentova:
@nqhhdev I'm afraid the issue happens again... 2.22.1 

https://github.com/user-attachments/assets/ecab098b-4d25-408e-9822-f9767ccd44ba

nqhhdev:
@tprudentova I tried the new version, it works well. Help me verify it again.

https://github.com/user-attachments/assets/c5c6ec22-862b-4734-8307-0a545a10b0a1

tprudentova:
@nqhhdev I can still reproduce the issue in 2.23.4. Send 1st video in a chat -> good, send another one -> no loading preview, it only appears in the chat once fully uploaded 

https://github.com/user-attachments/assets/32f080e5-b7a3-4820-8a32-7acb33d90bcf

PR #2984, opened 2026-04-02
Enhance file picking and upload process with sequential reading and improved task management

nqhhdev:
## Ticket
**Related issue**

- #2983

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/6db07549-0cc9-4430-ae61-6f45d9d35b37

## Summary by CodeRabbit

* **Refactor**
  * Enhanced file selection workflow to use sequential reading when picking multiple files during upload operations, improving how files are processed.
  * Optimized web file upload initialization with a refined two-phase approach: files are now registered first with proper state setup, then queued for upload, improving upload reliability and task management.

### DIFF
--- a/lib/utils/manager/upload_manager/upload_manager.dart
+++ b/lib/utils/manager/upload_manager/upload_manager.dart
@@ -320,6 +320,10 @@ class UploadManager {
     String? caption,
     Event? inReplyTo,
   }) async {
+    // Store txids and fake events for all files
+    final uploadTasks = <String, SyncUpdate>{};
+
+    // First pass: Create all placeholders and fake events
     for (final matrixFile in files.asMap().entries) {
       final txid = room.client.generateUniqueTransactionId();
       final fileIndex = matrixFile.key;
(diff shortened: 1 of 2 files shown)


lib/utils/manager/upload_manager/upload_manager.dart
```diff
@@ -343,6 +347,25 @@ class UploadManager {
         uploadInfo: _eventIdMapUploadFileInfo[txid]?.toJson(),
       );
 
+      uploadTasks[txid] = fakeFileEvent;
+
+      final streamController =
```
9clg6, line 352: Close the stream in case of error; Cf my comments

9clg6 (changes requested):

lib/utils/manager/upload_manager/upload_manager.dart
nqhhdev: Thanks for the review. Before addressing this, I have a few questions:
- What specific issue are you observing? Is there a reproduction case showing memory growth or functional problems?
- Have you verified this causes a memory leak? The map is bounded by the number of failed uploads per session.
- Did you review the retry flow? The retryUpload() method (line 86-194) depends on this state being preserved.

## Why the current logic is correct
The guard if (info?.isFailed != true) is intentional, not a bug:
1. Retry requires preserved state
  - When upload fails, we must keep the UploadFileInfo, streamController, and all context in memory
  - Without this, retry functionality breaks completely
2. Restoring from event data is insufficient
  - UI stream listeners are lost (new stream ≠ subscribed stream)
  - Binary data is not serialized (especially MatrixFile.bytes on web)
  - Active CancelToken references cannot be restored
3. Cleanup paths exist
  - Success/retry success → cleanup via onTaskCompleted
  - User cancels → cleanup via cancelUpload() (line 59-67)
  - Memory only retained when user neither retries nor cancels—acceptable trade-off
  
## Conclusion: 
- This follows the standard "optimistic retry" pattern. Removing this guard would break retry functionality. If you have specific memory concerns with reproduction steps, we can explore alternative solutions like room-level cleanup on navigation.

jbrnzl (approved):

lib/utils/manager/upload_manager/upload_manager.dart
9clg6, line 352: Is this controller ever closed ?

nqhhdev: Cleanup paths exist
- Success/retry success → cleanup via onTaskCompleted
- User cancels → cleanup via cancelUpload()

9clg6 (approved):

PR #2986, opened 2026-04-03
fix: edit rendering after merge and revoke web blob URLs

9clg6:
## Summary

Two unrelated fixes around message rendering and web media handling:

1. **Edit rendering** — `MessageContentBuilder` and `isDisplayOnlyEmoji` now consult the display event (latest edit) instead of the original event, so bubble sizing, text style, and emoji-only detection stay consistent with what `MessageContent` actually renders after an edit.
2. **Web blob URL cleanup** — `generateVideoThumbnail` and `getVideoDuration` now revoke their `Blob` object URLs in `finally` blocks, so the backing `ArrayBuffer` is released even on exception paths.

### Edit rendering uses the display event

**`message_content_builder.dart` / `message_content_builder_mixin.dart` / `message_content_with_timestamp_builder.dart`**

`MessageContentBuilder` now forwards the `Timeline` to the size-calculation mixin. Inside the mixin, `_paintMessageText` resolves the display event via `event.getDisplayEventWithoutEditEvent(timeline)` and paints against its body + style. The bubble decoration branch in `_MessageContentWithTimestampBuilderState` now calls `isDisplayOnlyEmoji(widget.timeline)` so edited messages with added text no longer render as enlarged emoji.

**`event_extension.dart` + `event_extension_test.dart`**

`isDisplayOnlyEmoji` takes an optional `Timeline`. When provided, it evaluates `onlyEmotes` / `numberEmotes` on the display event. Reply/caption short-circuit moved to the top so edited replies are handled correctly. Unit tests cover: original emoji-only / edited to text, original with text / edited to emoji-only, reply events, caption mode, boundary counts (0, 6, 7 emojis).

### Web blob URL leak fix

**`media_thumbnail_extension.dart`**

- `generateVideoThumbnail`: `url` is now declared before the `try`; the cleanup (file delete on mobile, `revokeWebUrl` on web) moved into a `finally` block so the blob URL and temp file are freed on every path, including exceptions.
- `getVideoDuration`: the created blob URL is stored and revoked in `finally` alongside the existing `VideoPlayerController.dispose()`.

**`web_url_creation_extension.dart`**

Adds the matching `revokeWebUrl(String url)` helper that calls `html.Url.revokeObjectUrl`. Without this, every video thumbnail or duration probe on web pinned an `ArrayBuffer` in browser memory for the lifetime of the tab.

## Test plan
- [x] `fvm flutter analyze` clean
- [x] Unit tests for `isDisplayOnlyEmoji` pass (edit → text, edit → emoji, reply, caption, boundaries)
- [x] Manual: edit a text message to add characters — bubble no longer oversized, no stale emoji styling
- [x] Manual: edit an emoji-only message to add text — emoji enlargement disabled
- [x] Manual web: upload a video — thumbnail renders, no growth of retained `ArrayBuffer`s in DevTools heap snapshots across repeated uploads

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,7 @@ test_bundle.dart
 sentry.properties
 
 .fvm
+
+# Playwright
+playwright-report/
+test-results/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+## [2.22.0] - 2026-04-22
+### Added
+- TW-2963: Forward to multiple rooms
+- Translated using Weblate (French, Irish, Russian, Vietnamese)
+
+### Changed
+- TW-2942: Refactor connection status handling in chat components
+- TW-3568: EventVisibilityResolver and chat-list/timeline visibility rules
+- TW-2973: Update chat room preset to private, remove power level overrides
+- Migrate routing to go_router_builder for type-safe navigation
+- Enhance file picking and upload with sequential reading and task management
+- Sentry: Track wrong member count and missing last message
+
+### Fixed
+- TW-3004: Enhance emoji display in timeline to consider edit events
+- TW-2957: Fix cannot drag scrollbar on web
+- TW-2952: Fix message time overlap and scroll glitch
+- TW-2988: Fix emoji display logic to exclude replies
+- TW-2993: Improve scroll behavior for message replies
+- fix: resolve GoRouter type cast crash when opening pinned messages
+- fix: replace FluffyChat deep link prefix with Twake scheme
+- fix: timeline incorrectly shows message as unread
+
+### Performance
+- perf: fix O(N²) hashCode cascade in ChatAppBarTitle room name resolver
+
 ## [2.21.17] - 2026-04-08
 ### Fixed
 - HOTFIX: Fix disoriented notification avatar (#2992)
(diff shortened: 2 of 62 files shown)


jbrnzl (approved):
LGTM

lib/widgets/mixins/get_preview_url_mixin.dart
```diff
@@ -43,6 +45,12 @@ mixin GetPreviewUrlMixin<T extends StatefulWidget> on State<T> {
         );
   }
 
+  void disposeGetPreviewUrl() {
+    _previewUrlSubscription?.cancel();
+    _previewUrlSubscription = null;
+    getPreviewUrlStateNotifier.dispose();
```
hoangdat: maybe it dispose twice (one in dispose method)

lib/widgets/mixins/get_preview_url_mixin.dart
```diff
@@ -27,6 +27,8 @@ mixin GetPreviewUrlMixin<T extends StatefulWidget> on State<T> {
     super.dispose();
   }
 
+  StreamSubscription? _previewUrlSubscription;
```
hoangdat: I still not got the idea of this?

hoangdat (approved):

nqhhdev (approved):
LGMT

jbrnzl (approved):

PR #2991, opened 2026-04-07
fix: timeline incorrectly shows message as unread

jbrnzl:
## Root cause
getSeenByUsers had split logic paths (depending on receiptState.**mainThread**) and could derive seen/unseen status inconsistently from timeline receipts.
That could mark timeline items as unread even when receipts indicated they had been read.

2991-debug-0

According to Matrix, read markers should be interpreted consistently from receipt progression (including threaded semantics where applicable). With threads, read progression is scoped by context (thread_id), so we must evaluate receipts in all scopes (main timeline vs a specific thread), or you can show inconsistent states.

## Solution
- Simplified getSeenByUsers in lib/utils/room_status_extension.dart to a single deterministic flow:
  - iterate timeline events down to the target eventId
  - aggregate receipts only within that range
  - exclude the current user directly during aggregation
- Removed now-unused import.
- Added/expanded tests in test/utils/room_status_extension_test.dart for:
  - empty timeline
  - own-user-only receipts
  - latest-event receipts
  - accumulation up to target event
  - exclusion of receipts older than target
  - mixed receipts where own user must be filtered out

### Web
First window: Hosted Staging / Second: Local Staging with fix

https://github.com/user-attachments/assets/a0429db5-54ad-4714-8b08-1326212599ec

### Mobile
Left: Hosted Staging / Right: Local Staging with fix

https://github.com/user-attachments/assets/43e84014-bd5c-49c0-a8ba-3330d152c08f

## Test recommendations

- Run unit tests for room_status_extension.
- Manual verification in timeline UI:
  - messages with other-user receipts are no longer incorrectly shown unread
  - own user is not included in “seen by” users
  - behavior is correct when a specific eventId is provided.
- Validate in rooms with different timeline depths to ensure no regression.
- Chat with an other client (ie. element) in a thread, read the message and come back to Twake Chat: message is "unread".

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate determination of who has seen a message by limiting receipt aggregation to events up to the specified message and correctly excluding the current user from results.
  * Returns empty results when the target message isn't found.

* **Tests**
  * Added thorough tests covering empty timelines, own-only receipts, multi-event accumulation, missing target message, and filtering out the current user.

### DIFF
--- a/lib/utils/room_status_extension.dart
+++ b/lib/utils/room_status_extension.dart
@@ -1,7 +1,6 @@
 import 'package:collection/collection.dart';
 import 'package:fluffychat/generated/l10n/app_localizations.dart';
 import 'package:fluffychat/utils/date_time_extension.dart';
-import 'package:fluffychat/utils/string_extension.dart';
 import 'package:flutter/widgets.dart';
 import 'package:matrix/matrix.dart';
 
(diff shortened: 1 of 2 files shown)


hoangdat:
Maybe, we should filter out if the reader is the current user.

9clg6 (approved):

hoangdat (approved):

PR #2996, opened 2026-04-09
refactor: extract duplicated event-type sets into EventTypeSets

jbrnzl:
## Summary

- Introduces `EventTypeSets` (`lib/config/event_type_sets.dart`), a single source of truth for the event-type sets that were duplicated as inline literals across 6 files. Later it will be the source of truth where we can add new events or change the visibility behavior of each.
- Replaces every occurrence with a reference to the shared constants — **zero behavioral change**.
- Adds 25 unit tests on `isVisibleInGui` and `isState` to freeze the current filtering contract before further refactoring or spec changes.

### Three sets extracted

| Constant | Contents | Replaced in |
|----------|----------|-------------|
| `messageContentTypes` | Message, Sticker, Encrypted | `isState`, `isSameSenderWith`, local notifications, chat-list preview, server history filter |
| `userContentTypes` | Message, Sticker, Encrypted, CallInvite | `Message.build` type guard, `hasNoMessageEvents`, `isEmptySupportChat` |
| `importantStateEvents` | Encryption, RoomCreate, RoomMember, RoomTombstone, CallInvite, RoomName, RoomAvatar | `isVisibleInGui` unimportant-state filter |

### What this does NOT change

- No logic in `isVisibleInGui` is modified (yet) (same chain of `&&`, same guards)
- No methods removed (`shouldHideBannedEvent`, `isActivateEndToEndEncryption`… stay as-is for now)
- `lastEventAvailableInPreview` is untouched for now
- `chat_list_item_subtitle.dart` is untouched for now

## Tests

- 25 new tests added in `filtered_timeline_extension_test.dart`
- You can open chat list, open a conversation, send message/sticker, verify timeline renders identically
- You can verify room list sorting and last-message preview unchanged

## Demos
No demos added as there is zero behavioral change

## Summary by CodeRabbit

* **Refactor**
  * Centralized event-type groupings used across the app (message-like, user-content, important state) for consistent filtering and rendering.
* **Tests**
  * Added comprehensive tests validating event categorization and GUI visibility under various configurations and edge cases.

### DIFF
--- a/lib/domain/matrix_events/event_type_sets.dart
+++ b/lib/domain/matrix_events/event_type_sets.dart
@@ -1,0 +1,36 @@
+import 'package:matrix/matrix.dart';
+
+/// Canonical event-type sets used for filtering, grouping and display.
+abstract final class EventTypeSets {
+  /// Message-like types: messages, stickers, encrypted payloads.
+  static const Set<String> messageContentTypes = {
+    EventTypes.Message,
+    EventTypes.Sticker,
+    EventTypes.Encrypted,
+  };
+
+  static const List<String> messageContentTypesList = [
+    EventTypes.Message,
+    EventTypes.Sticker,
+    EventTypes.Encrypted,
+  ];
+
+  /// [messageContentTypes] + signalling types produced by users (calls).
+  static const Set<String> userContentTypes = {
+    EventTypes.Message,
+    EventTypes.Sticker,
+    EventTypes.Encrypted,
+    EventTypes.CallInvite,
+  };
+
+  /// State events kept visible when "hide unimportant state events" is on.
+  static const Set<String> importantStateEvents = {
+    EventTypes.Encryption,
+    EventTypes.RoomCreate,
+    EventTypes.RoomMember,
+    EventTypes.RoomTombstone,
+    EventTypes.CallInvite,
+    EventTypes.RoomName,
+    EventTypes.RoomAvatar,
+  };
+}
--- a/test/utils/matrix_sdk_extensions/filtered_timeline_extension_test.dart
+++ b/test/utils/matrix_sdk_extensions/filtered_timeline_extension_test.dart
@@ -1,0 +1,352 @@
+import 'package:fluffychat/config/app_config.dart';
+import 'package:fluffychat/domain/matrix_events/event_type_sets.dart';
+import 'package:fluffychat/utils/matrix_sdk_extensions/filtered_timeline_extension.dart';
+import 'package:flutter_test/flutter_test.dart';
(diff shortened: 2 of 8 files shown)


lib/domain/matrix_events/event_type_sets.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:matrix/matrix.dart';
+
+/// Canonical event-type sets used for filtering, grouping and display.
+abstract final class EventTypeSets {
```
9clg6, line 4: I'm not convinced about the location of this file, why lib/config ?

jbrnzl: @9clg6 Thank you, it was a mistake. I moved it to lib/domain/matrix_events/ as it sounds good for me but let me know if you had an other location in mind.

9clg6: LGTM now, thanks

9clg6 (approved):

test/utils/matrix_sdk_extensions/filtered_timeline_extension_test.dart
```diff
@@ -0,0 +1,352 @@
...
+    AppConfig.hideRedactedEvents = false;
+    AppConfig.hideUnknownEvents = true;
+    AppConfig.hideUnimportantStateEvents = true;
+  });
+
+  tearDown(() {
+    AppConfig.hideRedactedEvents = savedHideRedacted;
+    AppConfig.hideUnknownEvents = savedHideUnknown;
+    AppConfig.hideUnimportantStateEvents = savedHideUnimportant;
+  });
```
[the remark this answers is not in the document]
jbrnzl: hideAllStateEvents is static const, no setter, so save/restore isn’t possible; the suggested patch fails to compile. I will not modify it

lib/domain/matrix_events/event_type_sets.dart
nqhhdev, line 1: My concern: .toList() allocation on every getter call. Since messageContentTypes is already a const Set, .toList() allocates a new List instance every time this getter is called. Would it make sense to either:                                                                                                                                                                                            
 1. Change the return type to Set<String> and return the const set directly — no allocation, and Set.contains() is O(1), or                                                                                                                                                                                                                            
 2. Add a static const List<String> messageContentTypesList to EventTypeSets if a List is truly required by callers?

jbrnzl: Thank you @nqhhdev . Switched to static const List.

nqhhdev (approved):

PR #3001, opened 2026-04-14
perf(chat): fix O(N²) hashCode cascade in ChatAppBarTitle room name resolver

9clg6:
## Summary

`ChatAppBarTitle._getRoomName` was the dominant CPU hotspot on room-switch rebuilds. A Set-fold with spread (`{...prev, ...contact.toPresentationContacts()}`) rebuilt a `LinkedHashSet<PresentationContact>` at every iteration, hashing Equatable-based contacts with expensive props on every rebuild of the two `ValueListenableBuilder`s bound to `getContactsNotifier()`.

Changes:
- Replace the Set-fold by a lazy `expand().firstWhereOrNull()` that short-circuits on the first `matrixId` match.
- Hoist a single `ValueListenableBuilder` at the InkWell level so both the avatar name and the title text share one rebuild cycle. `_ChatAppBarStatusContent` is passed via `child:` so it no longer rebuilds on contacts notifier emissions.
- Extract resolution logic into a top-level `resolveChatAppBarTitle` function (`@visibleForTesting`).
- 6 unit tests covering fallback, match, and failure branches.

## Measured impact (10-run bench, same methodology for baseline and fix)

Seeds 1–10, randomized 15-room parcours, profile-mode local build, Chrome fresh + auto-OIDC login per run.

| Metric | Baseline (median) | Fix (median) | Delta |
|---|---:|---:|---:|
| `script_total_s` | 14.28s | **6.41s** | **-55%** |
| `long_task_total_ms` | 9 206 ms | **846 ms** | **-91%** |
| `long_task_count` | 30 | 10 | -67% |
| `long_task_max_ms` | 418 ms | 133 ms | -68% |
| `wall_total_s` | 33.86s | 28.63s | -15% |
| `heap_end_mb` | 112.35 | 104.07 | -7% |

Cross-checked via inverted call tree on runs 1, 3, 6, 9: `_getRoomName` self-time = 0.000s, `mapPropsToHashCode` self-time ≤ 0.001s — the hotspot is gone.

Full report: `docs/plan/findings/2026-04-14-perf-web-cpu-report.md` (§10).

## Known blind spots (out of scope, follow-ups)

The bench exercises room switching (which matches the user-reported "5 s to open a room"). The following paths were not measured and may still contain independent hotspots in production:
- Typing lag (composer focus / onChanged rebuilds).
- Very long sessions (accumulated Matrix SDK state over 1 h+).
- Large address books (> 500 contacts — the new code is O(N) without memoization).
- Deep navigation stacks (space/room rapid switching).

Independent second opinions (wingspan + rodin-tech) validated semantic equivalence, metric consistency, and flagged these blind spots.

## Test plan

- [x] `fvm flutter analyze lib/pages/chat/chat_app_bar_title.dart test/pages/chat/chat_app_bar_title_test.dart` — clean
- [x] `fvm flutter test test/pages/chat/chat_app_bar_title_test.dart` — 6/6 passing
- [ ] Manual smoke: open a direct chat and verify the app bar shows the contact display name (not the MXID)
- [ ] Manual smoke: open a group chat and verify the app bar shows the localized room name
- [ ] Manual smoke: go offline while a direct chat is open; app bar still shows a sensible name (no crash on the `Failure` branch)

## Summary by CodeRabbit

* **Refactor**
  * Optimized chat app bar title display with improved contact matching for direct chats and reduced render cycles.

* **Tests**

https://github.com/user-attachments/assets/8744caa0-029e-4272-9c48-d9cd2f3ee571

### DIFF
--- a/lib/pages/chat/chat_app_bar_title.dart
+++ b/lib/pages/chat/chat_app_bar_title.dart
@@ -10,8 +10,8 @@ import 'package:fluffychat/domain/app_state/contact/get_contacts_state.dart';
 import 'package:fluffychat/domain/contact_manager/contacts_manager.dart';
 import 'package:fluffychat/pages/chat/chat_app_bar_title_style.dart';
 import 'package:fluffychat/presentation/extensions/contact/presentation_contact_extension.dart';
-import 'package:fluffychat/presentation/model/contact/presentation_contact.dart';
 import 'package:fluffychat/pages/chat/typing_timer_wrapper.dart';
+import 'package:fluffychat/presentation/model/contact/presentation_contact.dart';
 import 'package:fluffychat/resource/image_paths.dart';
 import 'package:fluffychat/utils/common_helper.dart';
 import 'package:fluffychat/utils/room_status_extension.dart';
@@ -54,34 +54,15 @@ class ChatAppBarTitle extends StatelessWidget {
     BuildContext context,
     Either<Failure, Success> getContactsState,
   ) {
-    final directChatMatrixId = room?.directChatMatrixID;
     final localizedRoomName = room?.getLocalizedDisplayname(
       MatrixLocals(L10n.of(context)!),
     );
-    if (directChatMatrixId == null) {
-      if (roomName != null) return roomName!;
-
-      return localizedRoomName ?? '';
-    }
-
-    final currentContacts = getContactsState.fold(
-      (failure) => <PresentationContact>[],
-      (success) => success is GetContactsSuccess
-          ? success.contacts
-                .fold(
-                  <PresentationContact>{},
-                  (previous, contact) => {
-                    ...previous,
-                    ...contact.toPresentationContacts(),
-                  },
-                )
-                .toList()
-          : <PresentationContact>[],
+    return resolveChatAppBarTitle(
+      directChatMatrixId: room?.directChatMatrixID,
+      fallbackRoomName: roomName,
+      localizedRoomName: localizedRoomName,
+      getContactsState: getContactsState,
     );
-    final availableContact = currentContacts.firstWhereOrNull(
-      (contact) => contact.matrixId == directChatMatrixId,
-    );
-    return availableContact?.displayName ?? localizedRoomName ?? '';
   }
 
   @override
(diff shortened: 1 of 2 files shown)


lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -378,3 +354,33 @@ class _ChatAppBarTitleTyping extends StatelessWidget {
...
+}) {
+  if (directChatMatrixId == null) {
+    return fallbackRoomName ?? localizedRoomName ?? '';
+  }
+  final availableContact = getContactsState
+      .getSuccessOrNull<GetContactsSuccess>()
+      ?.contacts
+      .expand((contact) => contact.toPresentationContacts())
+      .firstWhereOrNull((contact) => contact.matrixId == directChatMatrixId);
+  return availableContact?.displayName ?? localizedRoomName ?? '';
```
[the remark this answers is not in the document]
9clg6: Valid edge case, but it's pre-existing behavior — the old code had the same `availableContact?.displayName ?? localizedRoomName ?? ''` return. This PR is strictly a perf fix (no logic change), so I'd prefer to keep the scope tight and address the empty-displayName fallback in a separate PR + regression test. Tracking it separately.

nqhhdev:
Please don’t forget to include demos for both web and mobile. This is important because it helps verify that your changes actually work. Without it, I not only have to review the code but also spend extra time testing all cases and checking for side effects. I’d really appreciate it if you could include this going forward 🙏

pm-McFly:
> Please don’t forget to include demos for both web and mobile. This is important because it helps verify that your changes actually work. Without it, I not only have to review the code but also spend extra time testing all cases and checking for side effects. I’d really appreciate it if you could include this going forward 🙏

I'd argue that demos are especially essential when doing feature or bug fixing
here a before/after of the expected improvements is enough IMHO.

For breaking or regressions, we'd better rely on tests tbh

nqhhdev:
> 
> 
> 
> I'd argue that demos are especially essential when doing feature or bug fixing
> 
> here a before/after of the expected improvements is enough IMHO.
> 
> 
> 
> For breaking or regressions, we'd better rely on tests tbh


I agree it depends on the case, but if the PR changes code structure or UI/UX, a demo should be included.
It only takes a few minutes and provides clear evidence while helping avoid regressions.

lib/pages/chat/chat_app_bar_title.dart
```diff
@@ -378,3 +354,33 @@ class _ChatAppBarTitleTyping extends StatelessWidget {
...
+///
+/// For direct chats, walks the contacts state lazily to find the first
+/// [PresentationContact] whose `matrixId` matches [directChatMatrixId], and
+/// uses its display name. Falls back to [localizedRoomName] when no contact
+/// matches or when the state is not a success.
+///
+/// Lazy iteration + short-circuit on match is deliberate: do not rebuild a
+/// `Set<PresentationContact>` here. `PresentationContact` is `Equatable`
+/// with expensive props, and a Set-fold triggers an O(N²) `hashCode` cascade
+/// on every rebuild (cf. perf report 2026-04-14).
```
jbrnzl: Is it worth mentioning the report here ? I think it's not accessible for us

9clg6:
> > I'd argue that demos are especially essential when doing feature or bug fixing
> > here a before/after of the expected improvements is enough IMHO.
> > For breaking or regressions, we'd better rely on tests tbh
> 
> I agree it depends on the case, but if the PR changes code structure or UI/UX, a demo should be included. It only takes a few minutes and provides clear evidence while helping avoid regressions.

I'll do it np !

lib/pages/chat/chat_app_bar_title.dart
9clg6: done

jbrnzl (approved):

nqhhdev (approved):

PR #3007, opened 2026-04-17
fix: resolve GoRouter type cast crash when opening pinned messages

### LINKED ISSUE #3006: [BUG] Can't open pin screen
### Reproduction Steps
- Open a chat that has a pinned messages banner at the top.
- Click on “Pinned messages”.
- The app redirects to an error screen.

### Expected Result
- The pinned messages screen should open correctly.

https://github.com/user-attachments/assets/8257b23b-f753-461c-af95-f73c99bd39dd

9clg6:
## Summary
- **Bug**: Navigating to the pinned messages screen crashed with `GoException: type 'PinnedEventsArgument' is not a subtype of type 'ChatRouterInputArgument?' in type cast`
- **Root cause**: `RoomRoute.$extra` was typed as `ChatRouterInputArgument?`, so the generated `_fromState` hard-casts `state.extra`. When the child `PinnedMessagesRoute` passes a `PinnedEventsArgument`, the parent route's cast fails during redirect.
- **Fix**: Widen `RoomRoute.$extra` to `Object?` and add a safe `is` check in `buildPage`, preserving existing draft/share behavior.

Closes #3006

https://github.com/user-attachments/assets/6545952c-6cc8-4750-80be-b5868bd2c150

## Summary by CodeRabbit

* **Refactor**
  * Safer handling of routing parameters with runtime validation and guarded access.
  * Reduces risk of routing-related crashes while preserving existing draft and share behaviors (including prefix, room name, and shared files).
  * Internal change only; no visible UI or workflow changes.

### DIFF
--- a/lib/config/go_routes/app_routes.dart
+++ b/lib/config/go_routes/app_routes.dart
@@ -766,7 +766,7 @@ class RoomRoute extends GoRouteData with $RoomRoute {
   const RoomRoute({required this.roomid, this.event, this.$extra});
   final String roomid;
   final String? event; // query param: ?event=xxx
-  final ChatRouterInputArgument? $extra;
+  final Object? $extra;
 
   @override
   FutureOr<String?> redirect(BuildContext context, GoRouterState state) =>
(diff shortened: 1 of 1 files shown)


jbrnzl (approved):

nqhhdev:
- [ ] Doesn't work on mobile

https://github.com/user-attachments/assets/261431e8-7287-4278-8e74-67329221a1f3

Screenshot 2026-04-21 at 11 46 18

cc: @9clg6

nqhhdev:
LGTM, but need to fix Rabbit comment

jbrnzl (approved):

pm-McFly:
> LGTM, but need to fix Rabbit comment

Usually I'd tend to not ensure/enforce fixing nitpicks too much.
That one truly sounds real, and could be at certain point in time, but it's not a priority at the moment.

Therefore, I advocate to roll that PR out and to address this nitpick comment in another PR if we encounter the issue.

@9clg6 if you are confident that won't blow up day one, let's merge :)

hoangdat:
to look on this nitpick comment in detail 

image

I found the first one should be addressed because is related to casting data, especially it is the exchange data between routes. 

But agree we can ignore the second nitpick

9clg6:
> to look on this nitpick comment in detail
> 
> image
> I found the first one should be addressed because is related to casting data, especially it is the exchange data between routes.
> 
> But agree we can ignore the second nitpick

Fixed

nqhhdev (approved):

PR #3032, opened 2026-05-04
TW-2705(fix): stale pending events no longer shadow confirmed message…

### LINKED ISSUE #3016: [BUG] Wrong last message displayed

### Describe the bug

Image

You can see in the left sidebar, Tchat is saying that my latest message with Khaled is on 24 April. But this is not true as you can see, I've chat with him today. 

It's not the first time we have this kind of issue, cf https://github.com/linagora/twake-on-matrix/issues/1862 for instance 

### Pre-Bug State

.

### To Reproduce

.

### Expected results

.

### Context

Chrome / 2.21.27

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
Fixes #3016 
## Root cause

`room.lastEvent` the SDK's in-memory field updated on each sync can hold a non-confirmed event (sending or error). The chat list preview had two separate sources of truth: `room.lastEvent` (in-memory, sync) and the async DB scan result (`previewResult`). Reconciliation was done inside the widget (`ChatListItemSubtitle`) without proper timestamp comparison, so a non-confirmed event in `room.lastEvent` could silently override a more recent confirmed message from the DB.

## Solution

**Single source of truth in the domain layer**: `lastEventAvailableInPreview()` now reads `this.lastEvent` as a first candidate alongside state events and DB events, removing the need for any reconciliation logic in the widget.

**Timestamp-aware priority**: two buckets (`bestSynced` / `bestPending`) are maintained throughout the scan. Confirmed events (`status.isSent`) always take priority. A pending/error event only wins if it is genuinely more recent (e.g. a message just sent with no network). This is resolved at the very end of the scan via `_newestEvent(bestSynced, bestPending)`.

**No premature early returns on pending candidates**: early returns are guarded by `bestSynced != null` only, ensuring the DB scan always runs when the best in-memory candidate is unconfirmed.

**No flicker on room updates**: `ChatListSortRooms` previously nullified `_previewByRoomId[roomId]` immediately on `onUpdate`, causing a blank preview during the async recomputation. It now adds the room to a `_pendingRefresh` set instead, keeping the last known preview visible until the new result is ready.

## Impact description

Applies only to the chat list preview. No change to the timeline view or message sending flow.

## Test recommendations

- Inject or mock a `SENDING` event with `origin_server_ts` set 7 days in the past, then verify the confirmed last
  message is shown instead.
- Send a message with network disabled and verify the pending message appears
  immediately in the preview.
- Re-enable network, let the message confirm, and verify the confirmed message is
  still shown.
- Open a room and return to the chat list: verify no flicker occurs on the
  preview.
- 4 regression tests added in
  `test/domain/model/room/last_event_available_in_preview_test.dart`.

## Other
Needs #3028 to be complete. By itself, this PR does not fix the behavior where icons are not shown on hover due to a chat list difference.

## Resolved
- Web:

https://github.com/user-attachments/assets/de4d0e74-51be-4662-8030-e78dc4a91796

- Mobile:

https://github.com/user-attachments/assets/76820a23-6ed9-41cf-9ebe-c051f14e51c6

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected message-preview selection to prefer confirmed/sent messages over pending/unsent ones and handle incomplete states consistently.

* **Performance**
  * Added preview caching with targeted refresh tracking to avoid redundant recomputation and reduce database scans.

* **Tests**
  * Added test coverage for pending vs confirmed message precedence and fallback behaviors.

### DIFF
--- a/lib/domain/model/room/room_extension.dart
+++ b/lib/domain/model/room/room_extension.dart
@@ -203,7 +203,12 @@ extension RoomExtension on Room {
           }
         }
       }
-      if (bestSynced != null) return RoomPreviewFound(bestSynced);
+      if (bestSynced != null) {
+        final best = bestPending != null
+            ? _newestEvent(bestSynced, bestPending)
+            : bestSynced;
+        return RoomPreviewFound(best);
+      }
 
       final dbEvents = await client.database.getEventList(
         this,
--- a/lib/domain/model/room/room_extension.dart
+++ b/lib/domain/model/room/room_extension.dart
@@ -170,22 +170,40 @@ extension RoomExtension on Room {
   Future<RoomPreviewResult> lastEventAvailableInPreview() async {
     const previewLimit = 30;
     try {
+      Event? bestSynced;
+      Event? bestPending;
+
+      final syncCandidate = lastEvent;
+      if (syncCandidate != null &&
+          EventVisibilityResolver.isEligibleForChatListPreviewSync(
+            syncCandidate,
+          )) {
+        if (syncCandidate.status.isSent) {
+          bestSynced = syncCandidate;
+        } else {
+          bestPending = syncCandidate;
+        }
+      }
+
+      if (bestSynced != null) return RoomPreviewFound(bestSynced);
+
       final statePreviewCandidates = client.roomPreviewLastEvents
           .map(getState)
           .whereType<Event>()
           .toList();
-      Event? best;
       for (final e in statePreviewCandidates) {
         if (await EventVisibilityResolver.isEligibleForChatListPreview(
           this,
           e,
         )) {
-          best = _newestEvent(best, e);
+          if (e.status.isSent) {
+            bestSynced = _newestEvent(bestSynced, e);
+          } else {
+            bestPending = _newestEvent(bestPending, e);
+          }
         }
       }
-      if (best != null) {
-        return RoomPreviewFound(best);
-      }
+      if (bestSynced != null) return RoomPreviewFound(bestSynced);
 
       final dbEvents = await client.database.getEventList(
         this,
@@ -198,11 +216,21 @@ extension RoomExtension on Room {
           this,
           e,
         )) {
-          best = _newestEvent(best, e);
+          if (e.status.isSent) {
+            bestSynced = _newestEvent(bestSynced, e);
+          } else {
+            bestPending = _newestEvent(bestPending, e);
+          }
         }
       }
-      if (best != null) {
-        return RoomPreviewFound(best);
+      // Prefer the most recent confirmed event. A pending/failed event wins
+      // only when it is genuinely newer (e.g. a message just sent without
+      // network).
+      final dbBest = bestPending != null
+          ? _newestEvent(bestSynced, bestPending)
+          : bestSynced;
+      if (dbBest != null) {
+        return RoomPreviewFound(dbBest);
       }
       return roomFullyScanned
           ? const RoomPreviewEmpty()
--- a/lib/pages/chat_list/chat_list_sort_rooms.dart
+++ b/lib/pages/chat_list/chat_list_sort_rooms.dart
@@ -33,6 +33,11 @@ class _ChatListSortRoomsState extends State<ChatListSortRooms> {
   Map<String, StreamSubscription?> _roomSubscriptions = {};
   Future<List<Room>>? _sortFuture;
 
+  /// Rooms whose preview is stale and must be recomputed on the next sort.
+  /// We keep the last known value in [_previewByRoomId] until the new result
+  /// is ready, so the UI never flickers to an empty preview mid-scan.
+  final Set<String> _pendingRefresh = {};
+
   /// Monotonically increasing counter to invalidate stale sort futures.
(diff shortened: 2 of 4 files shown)


lib/domain/model/room/room_extension.dart
```diff
@@ -198,11 +216,21 @@ extension RoomExtension on Room {
           this,
           e,
         )) {
-          best = _newestEvent(best, e);
+          if (e.status.isSent) {
+            bestSynced = _newestEvent(bestSynced, e);
+          } else {
+            bestPending = _newestEvent(bestPending, e);
+          }
```
nqhhdev, line 228: `EventStatus.error` falls into the `else` branch alongside `EventStatus.sending`. so a failed message can end up in `bestPending` and potentially surface as the chat-list view. Should we explicitly exclude error events here?

lib/pages/chat_list/chat_list_sort_rooms.dart
```diff
@@ -33,6 +33,11 @@ class _ChatListSortRoomsState extends State<ChatListSortRooms> {
   Map<String, StreamSubscription?> _roomSubscriptions = {};
   Future<List<Room>>? _sortFuture;
 
+  /// Rooms whose preview is stale and must be recomputed on the next sort.
+  /// We keep the last known value in [_previewByRoomId] until the new result
+  /// is ready, so the UI never flickers to an empty preview mid-scan.
+  final Set<String> _pendingRefresh = {};
```
nqhhdev, line 39: When a room is removed in didUpdateWidget, we cancel its sub and drop it from `_previewByRoomId`, but `_pendingRefresh` is never pruned. Should we add `_pendingRefresh.remove` alongside the sub?

test/domain/model/room/last_event_available_in_preview_test.dart
nqhhdev, line 1: The _message() helper doesn't set an explicit status. If the SDK default is not EventStatus.synced, then isSent returns false and the confirmed-vs-pending ordering tests would pass for the wrong reason. Should we pin status: EventStatus.synced explicitly?

lib/domain/model/room/room_extension.dart
jbrnzl: As discussed internaly, we will keep this behavior for now and see how we can notify users they have old messages in error.

lib/pages/chat_list/chat_list_sort_rooms.dart
jbrnzl: You are right @nqhhdev. I added _pendingRefresh.remove(room.id) does it seems ok for you ?

test/domain/model/room/last_event_available_in_preview_test.dart
jbrnzl: @nqhhdev From what I see in matrix-6.0.0/lib/src/event.dart line 63: `static const EventStatus defaultStatus = EventStatus.synced;` so for me default is synced ? I may be wrong please correct me

lib/pages/chat_list/chat_list_sort_rooms.dart
nqhhdev: Yes and test again

pm-McFly (approved):

lib/pages/chat_list/chat_list_sort_rooms.dart
nqhhdev, line 166: `_pendingRefresh` is never cleared in dispose() — `_roomSubscriptions` gets .clear() but `_pendingRefresh` is left as-is. Low risk since the widget is gone, but it's inconsistent with how other state is cleaned up. Worth adding `_pendingRefresh.clear()` before super.dispose() for symmetry.

WDYT @jbrnzl

lib/domain/model/room/room_extension.dart
```diff
@@ -170,22 +170,40 @@ extension RoomExtension on Room {
...
+            syncCandidate,
+          )) {
+        if (syncCandidate.status.isSent) {
+          bestSynced = syncCandidate;
+        } else {
+          bestPending = syncCandidate;
+        }
+      }
+
+      if (bestSynced != null) return RoomPreviewFound(bestSynced);
```
nqhhdev, line 188: Suggest either hoisting bestPending to function scope so it survives into the final fallback, or restructuring the early-exit guards so bestPending is always considered before returning a non-Found result.

test/domain/model/room/last_event_available_in_preview_test.dart
nqhhdev: Checked, correct

lib/domain/model/room/room_extension.dart
jbrnzl: bestPending is already in function scope so it survives to the DB fallback, but you're right that after we are dropping it: a synced state event could shadow a newer pending lastEvent. I will fix by applying the same _newestEvent(bestSynced, bestPending) logic there as in the DB phase

nqhhdev (approved):

Issue #3027, opened 2026-04-30
[BUG] When you're in a chat and a new message arrives, the arrow with the counter of new messages only appears after you scroll


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Given that you're in a chat and a new message arrives to this chat, the arrow that indicates that there are new messages only appears after you scroll the chat up or down

### Pre-Bug State

N/A

### To Reproduce

1. Open a chat and scroll to the very bottom of the chat history 
2. Receive a new message to this chat 
3. You can't tell that there's been a new message, because there's no arrow with a counter. It only appears after you scroll a bit 
NB: the counter appears fine if you somewhere in the middle of chat history (i'd guess it's because the arrow to the bottom is already there) 

https://github.com/user-attachments/assets/458dbe9e-af03-417a-afb3-7ac350b0f960

### Expected results

If I'm in a chat and I receive new messages, I see the arrow to the bottom with the counter of unread messages

### Context

- Device: tested on Redmi Note 12 Pro, Honor
- OS: tested on Android, Windows
- Browser: tested on Chrome, Firefox
- App Version: 2.22.0
- Server: staging 
- Account information: dubois@stg.lin-saas.com
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

tprudentova:
Orrrrr we can apply a different logic. For example, in telegram there are 2 different behaviours: 

1. You're at the very bottom of chat history. In this case, when a new message arrives, it sends you to the new bottom of chat history (you jump to the bottom the way it happened in TC before the fix)
2. You're anywhere except the very bottom of chat history. In this case when a new message arrives, you don't jump to the bottom but instead you see an arrow with the counter 

If we can apply this, it is preferrable :) 

https://github.com/user-attachments/assets/f15209e2-6e80-4380-b02c-b5f0b3c2be5b

PR #3033, opened 2026-05-05
TW-3027: Enable auto-scrolling to new messages when at the bottom of the chat

nqhhdev:
## Ticket
**Related issue**

- #3027 

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/60fe275e-b2a5-4153-b306-996ffb05d515

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-scroll now only triggers when appropriate, avoiding unwanted jumps when new messages arrive.
  * Improved detection of new incoming events and user scroll position: if the user was pinned near the bottom the view advances and the scroll-down button hides; otherwise the scroll-down button is shown to indicate unseen messages.


jbrnzl:
I am testing this one

jbrnzl (approved):
Tested and it's working well 💪🏻

lib/utils/matrix_sdk_extensions/event_list_extension.dart
```diff
@@ -83,6 +87,10 @@ extension EventListExtension on List<Event> {
         top: const [],
         bottom: List.from(newEvents),
         shouldScrollToBottom: false,
+        hasNewEvents:
+            newEvents.isNotEmpty &&
+            (oldEvents.isEmpty ||
+                newEvents.first.eventId != oldEvents.first.eventId),
```
pm-McFly, line 93: @nqhhdev kinda wonder, why should we use this new boolean although there is already `shouldScrollToBottom`?

nqhhdev: Great question and tks for your contribution, let me explain about `shouldScrollToBottom` vs `hasNewEvents`

- `shouldScrollToBottom` — "force scroll now" — only true for your own messages arriving at the top of the list. Unconditionally calls _scrollToBottom().                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                        
- `hasNewEvents` — "something new arrived" — true for any sender. Defers to a post-frame callback to snapshot wasAtBottom first, then either auto-scrolls or shows the "scroll down" button.

- Why both? shouldScrollToBottom is a sync decision (before layout). The "others' message + already at bottom" case needs addPostFrameCallback to read scroll position after the new message grows maxScrollExtent — that's what hasNewEvents enables.

pm-McFly: got it! thx for the explanation :)
maybe I am nitpicking if saying that the naming is thus a bit confusing - but that's really no biggy for today!

LGTM

pm-McFly (approved):

PR #3041, opened 2026-05-12
TW-3037(fix): send read receipts for hidden events

### LINKED ISSUE #3037: [BUG] New message displayed, but no?

### Describe the bug

This time, I've a channel where we display that I've an unread message. Not only on the channels list, but also within the channel itself. 

https://github.com/user-attachments/assets/277d0c42-9e00-4c08-80c8-384f43fa4a80

### Pre-Bug State

.

### To Reproduce

.

### Expected results

I think there is nothing to see 

### Context

- Device: iPhone 12 / Chrome
- App Version: Web 2.21.17 / iPhone : 2.21.17

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
## Ticket
Fixes #3037

## Root cause
When a room member deletes a message, the homeserver does not decrement
`notificationCount` in subsequent sync responses. The app correctly hides
redacted events from the timeline, but the notification badge keeps reading
the stale server-side count. Because hidden events (redactions, state changes)
never trigger the visibility callback that places a read marker, the badge
becomes impossible to dismiss: there is nothing visible to read.

## Solution
In `_processReadMarker` (`chat.dart`), when the user opens a room and a read
marker is about to be sent for an event:

1. Look up `room.lastEvent` (the true latest event, which may be hidden).
2. Confirm that `room.lastEvent` is present in `timeline.events`: this proves
   the user is at the live end of the conversation, not mid-conversation from
   a search jump.
3. Find the most recent visible event in the timeline
   (`EventVisibilityResolver.isVisibleInTimeline`).
4. If the event being marked **is** that last visible event, advance
   `currentEventId` to `room.lastEvent` before calling `setReadMarker`.

The homeserver then recomputes `notificationCount` to 0 and the phantom badge
disappears on the next sync.

## Known limitation
If a user preference hides deleted messages, the chat-list badge may temporarily show a higher unread count than visible messages. The badge is correctly reset when the chat is opened.

## Test recommendations
1. **Simple case**: receive a message (badge = 1), do not open the room,
   have the sender delete it. Open the room: badge should disappear.
2. **Edit + delete**: receive a message, sender edits it, then deletes it.
   Open the room: badge should disappear.
3. **False-positive guard**: receive a message, do not delete it. Opening the
   room must not clear the badge until the message is actually visible.
4. **Mid-conversation guard**: jump to an old event via search, scroll around.
   No phantom read markers should be sent for recent hidden events.

## Demos
- Before:

https://github.com/user-attachments/assets/66b60ba4-4ace-4370-8f26-c90f2fa0507b

- After:

https://github.com/user-attachments/assets/f0a3a5b6-7b1f-48a8-b4bb-171ca7e508de

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subscription cleanup to ensure all active connections are properly terminated, preventing potential resource leaks.
  * Enhanced read marker positioning in chat to more accurately track message read status.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-on-matrix/pull/3041?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -10,6 +10,7 @@ import 'package:fluffychat/data/memory/mxc_image_cache_manager.dart';
 import 'package:fluffychat/app_state/success.dart';
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/domain/matrix_events/event_type_rules.dart';
+import 'package:fluffychat/domain/matrix_events/event_visibility_resolver.dart';
 import 'package:fluffychat/di/global/get_it_initializer.dart';
 import 'package:fluffychat/domain/app_state/contact/get_contacts_state.dart';
 import 'package:fluffychat/domain/app_state/room/report_content_state.dart';
@@ -730,8 +731,20 @@ class ChatController extends State<Chat>
       final timeline = this.timeline;
       if (timeline == null || timeline.events.isEmpty) return;
 
-      final currentEventId = _pendingReadMarkerEventId;
+      var currentEventId = _pendingReadMarkerEventId;
       _pendingReadMarkerEventId = null;
+
+      // If the event being marked is the last visible event of the room,
+      // advance the read marker to room.lastEvent so that hidden trailing
+      // events don't leave a stale badge.
+      if (timeline.allowNewEvent) {
+        final lastVisibleEventId = timeline.events
+            .firstWhereOrNull(EventVisibilityResolver.isVisibleInTimeline)
+            ?.eventId;
+        if (currentEventId == lastVisibleEventId) {
+          currentEventId = room!.lastEvent!.eventId;
+        }
+      }
 
       Logs().d('Set read marker...', currentEventId);
 
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -666,8 +666,8 @@ class MatrixState extends State<Matrix>
     onLoginStateChanged.remove(name);
     await onNotification[name]?.cancel();
     onNotification.remove(name);
-    await onAutoReadReceipt[name]?.cancel();
-    onAutoReadReceipt.remove(name);
+    await onUiaRequest[name]?.cancel();
+    onUiaRequest.remove(name);
   }
 
   Future<void> initMatrix() async {
@@ -1316,11 +1316,21 @@ class MatrixState extends State<Matrix>
     }
     intentFileStreamSubscription?.cancel();
     intentUriStreamSubscription?.cancel();
-    onRoomKeyRequestSub.values.map((s) => s.cancel());
-    onKeyVerificationRequestSub.values.map((s) => s.cancel());
-    onLoginStateChanged.values.map((s) => s.cancel());
-    onNotification.values.map((s) => s.cancel());
-    onAutoReadReceipt.values.map((s) => s.cancel());
+    for (final s in onRoomKeyRequestSub.values) {
+      s.cancel();
+    }
+    for (final s in onKeyVerificationRequestSub.values) {
+      s.cancel();
+    }
+    for (final s in onLoginStateChanged.values) {
+      s.cancel();
+    }
+    for (final s in onNotification.values) {
+      s.cancel();
+    }
+    for (final s in onUiaRequest.values) {
+      s.cancel();
+    }
     onClientLoginStateChanged.close();
     client.httpClient.close();
     onFocusSub?.cancel();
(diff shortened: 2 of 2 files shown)


jbrnzl:
@pm-McFly Please review this carefully if you have the time 🙏🏻

pm-McFly (changes requested):
> @pm-McFly Please review this carefully if you have the time 🙏🏻

Thanks for the call — I think the suggested approach is interesting, but it would be cleaner to handle this by relying on sync/timeline/filter behavior rather than adding UI-specific logic per client/config.

For now, let’s address the core bug we agreed on:
> filtered events aren’t getting read markers because the Timeline view doesn’t see them.

A simpler solution focused on that would:

- Be easier to maintain
- Avoid client-specific UI workarounds
- Keep relying on the homeserver NotificationCount

Sketch of expected behavior:

1. 08:00 — receive text "Hello" → app shows notification → hs_count = 1
2. 08:05 — user opens chat → mark 1 as read → hs_count = 0
3. 08:07 — receive hidden event (room avatar change) → no notification → hs_count = 0 → hidden event is invisible to timeline, app can safely mark it read
4. 08:15 — receive text "task for today..." → visible & unread → notification → hs_count = 1
5. 08:16 — receive text "errata: tasks for today..." → visible → notification → hs_count = 2
6. 08:17 — receive image → notification → hs_count = 3
7. 08:17 — delete message from 08:15 → notification → hs_count = 4
8. 08:20 — user opens chat → app can safely mark as read up to the deletion event (7.)

**Known limitation**: If a user preference hides deleted messages, the chat-list badge may temporarily show a higher unread count than visible messages — acceptable here because the badge is correctly reset when the chat is opened.

jbrnzl:
> > @pm-McFly Please review this carefully if you have the time 🙏🏻
> 
> Thanks for the call — I think the suggested approach is interesting, but it would be cleaner to handle this by relying on sync/timeline/filter behavior rather than adding UI-specific logic per client/config.
> 
> For now, let’s address the core bug we agreed on:
> 
> > filtered events aren’t getting read markers because the Timeline view doesn’t see them.
> 
> A simpler solution focused on that would:
> 
> * Be easier to maintain
> * Avoid client-specific UI workarounds
> * Keep relying on the homeserver NotificationCount
> 
> Sketch of expected behavior:
> 
> 1. 08:00 — receive text "Hello" → app shows notification → hs_count = 1
> 2. 08:05 — user opens chat → mark 1 as read → hs_count = 0
> 3. 08:07 — receive hidden event (room avatar change) → no notification → hs_count = 0 → hidden event is invisible to timeline, app can safely mark it read
> 4. 08:15 — receive text "task for today..." → visible & unread → notification → hs_count = 1
> 5. 08:16 — receive text "errata: tasks for today..." → visible → notification → hs_count = 2
> 6. 08:17 — receive image → notification → hs_count = 3
> 7. 08:17 — delete message from 08:15 → notification → hs_count = 4
> 8. 08:20 — user opens chat → app can safely mark as read up to the deletion event (7.)
> 
> **Known limitation**: If a user preference hides deleted messages, the chat-list badge may temporarily show a higher unread count than visible messages — acceptable here because the badge is correctly reset when the chat is opened.

Thank you very much, I completely agree. This will be easier to maintain and not introduce such amount of logic in UI. I will make the changes you suggested and ask for your review again once done !

jbrnzl:
@pm-McFly I updated the PR following our discussion. Please I'll be more than happy if you can review it again and share concerns you may still have 🙏🏻

9clg6 (approved):

lib/pages/chat/chat.dart
```diff
@@ -729,9 +730,27 @@ class ChatController extends State<Chat>
...
+          lastRoomEvent.eventId != currentEventId &&
+          timeline.events.any((e) => e.eventId == lastRoomEvent.eventId)) {
+        final lastVisible = timeline.events.firstWhereOrNull(
+          EventVisibilityResolver.isVisibleInTimeline,
+        );
+        if (lastVisible?.eventId == currentEventId) {
+          currentEventId = lastRoomEvent.eventId;
+        }
+      }
+
```
pm-McFly, line 748: Sorry to be much nitpicking on that issue (:

Why do we need such a big condition? the room and the timeline are not aligned in some cases?

can't we go with a simpler like:

```dart
if (timeline.allowNewEvent && currentEventId == lastVisibleEventId) {
  currentEventId = room!.lastEvent!.eventId;
}
```

jbrnzl: @pm-McFly thank you for the careful review !
No they are aligned and `timeline.allowNewEvent` is a nice improvement indeed. I updated accordingly.
The reason for the extra checks and the firstWhereOrNull particularly is to tell us whether currentEventId is the latest visible event, which is what lets us safely advance past hidden trailing events without skipping unread visible ones.
Let me know if you have something else in mind.

pm-McFly (approved):

Issue #3047, opened 2026-05-15
[BUG] Video playing really slowly, as if it was slowed down


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When I play a video on my Android, it's really slow, you can see on the recording that one second of the video takes like six seconds to play

https://github.com/user-attachments/assets/92d666c7-9090-444e-b11f-5964901bd0f2

### Pre-Bug State

N/A

### To Reproduce

1. Send a video (it's a sometimes issue, so a bit tricky to catch...)
2. Play it 
3. The video is played on like 1/6 of speed

### Expected results

All videos are played in normal speed

### Context

- Device: Redmi Note 12 Pro
- OS: Android
- Browser: [e.g. chrome, safari]
- App Version: 2.22.1
- Server: linagora
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

PR #3056, opened 2026-05-19
TW-3047: Optimize video playback performance by enabling hardware acceleration and improving player initialization

nqhhdev:
## Ticket
**Related issue**

- #3047 

## Resolved
- Web:
- Android:
- IOS:

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary repaints in video playback and mobile controls for smoother UI
  * Enabled hardware-accelerated video decoding for better performance
  * Isolated thumbnail and position indicator rendering to lower CPU/GPU usage

* **Updates**
  * Improved video player lifecycle and media opening for more reliable resource handling and stability
  * Mobile video theme/widget now accepts an optional URL property for flexible content sources

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/twake-on-matrix/pull/3056?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/widgets/video_player.dart
+++ b/lib/widgets/video_player.dart
@@ -35,17 +35,15 @@ class _VideoPlayerState extends State<VideoPlayer> {
     configuration: const PlayerConfiguration(logLevel: MPVLogLevel.warn),
   );
 
-  /// Creates a [VideoController] with hardware acceleration explicitly enabled.
+  /// Creates a [VideoController] with explicit hwdec to force MediaCodec on Android.
   ///
-  /// On Android, libmpv's `hwdec=auto` can silently fall back to software
-  /// decoding on some devices (e.g. Redmi Note 12 Pro), causing ~6x slow
-  /// playback. Explicit `enableHardwareAcceleration` ensures mediacodec is
-  /// preferred. Web uses a different renderer and does not need this flag.
+  /// `hwdec=auto-safe` (the default) can silently fall back to software
+  /// decoding on some devices, causing slow playback. `mediacodec-copy` forces
+  /// HW decode via MediaCodec with a CPU-side copy, which has wider device
+  /// compatibility than plain `mediacodec`.
   VideoController _createController(Player p) => VideoController(
     p,
-    configuration: const VideoControllerConfiguration(
-      enableHardwareAcceleration: true,
-    ),
+    configuration: const VideoControllerConfiguration(hwdec: 'mediacodec-copy'),
   );
 
   /// Opens the media source on [player], logging any errors.
--- a/lib/widgets/video_player.dart
+++ b/lib/widgets/video_player.dart
@@ -1,4 +1,6 @@
-import 'package:flutter/foundation.dart';
+import 'dart:async';
+import 'dart:typed_data';
+
 import 'package:flutter/material.dart';
 import 'package:matrix/matrix.dart';
 import 'package:media_kit/media_kit.dart';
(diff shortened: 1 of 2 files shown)


9clg6 (approved):

lib/widgets/video_player.dart
```diff
@@ -30,49 +29,62 @@ class _VideoPlayerState extends State<VideoPlayer> {
...
+  /// Creates a [VideoController] with hardware acceleration explicitly enabled.
+  ///
+  /// On Android, libmpv's `hwdec=auto` can silently fall back to software
+  /// decoding on some devices (e.g. Redmi Note 12 Pro), causing ~6x slow
+  /// playback. Explicit `enableHardwareAcceleration` ensures mediacodec is
+  /// preferred. Web uses a different renderer and does not need this flag.
+  VideoController _createController(Player p) => VideoController(
+    p,
+    configuration: const VideoControllerConfiguration(
+      enableHardwareAcceleration: true,
```
jbrnzl: Hi @nqhhdev I can see in the `VideoControllerConfiguration` constructor that `enableHardwareAcceleration` already defaults to true so I don't see how setting it explicitly here would change the runtime behavior. I might be missing context on why it "ensures mediacodec is preferred".
If the root cause is really the HW decoder selection on Android, wouldn't the right fix be hwdec ? Something like
```
 configuration: const VideoControllerConfiguration(
    hwdec: 'mediacodec',
  ),
```
Could you share how you validated the fix on the Redmi Note 12 Pro or equivalent ? Just trying to make sure we're patching the actual root cause. I may be wrong, just a few interrogations.

nqhhdev: Thanks @jbrnzl, my explanation: 

- enableHardwareAcceleration: true is already the default → maps to hwdec=auto-safe, not mediacodec. The comment is misleading. => updated
- The actual fix should be VideoControllerConfiguration(hwdec: 'mediacodec-copy') — hwdec is directly exposed in v1.3.1. => Updated
- Validated on Android device (not the exact Redmi Note 12 Pro). Cannot fully confirm the fix targets the root cause without testing on the affected device.

cc: @jbrnzl

jbrnzl: Thank you @nqhhdev !

lib/widgets/video_player.dart
```diff
@@ -84,11 +94,10 @@ class _VideoPlayerState extends State<VideoPlayer> {
...
     return Scaffold(
-      body: Video(
-        fill: Colors.black,
-        pauseUponEnteringBackgroundMode: true,
-        resumeUponEnteringForegroundMode: true,
-        controller: videoController,
+      // RepaintBoundary isolates the video surface so that controls
+      // rebuilding above it do not trigger a repaint of the video layer.
+      body: RepaintBoundary(
+        child: Video(fill: Colors.black, controller: videoController),
```
jbrnzl: A last question, sorry @nqhhdev why are we deleting 
```
pauseUponEnteringBackgroundMode: true,
resumeUponEnteringForegroundMode: true,
```
? Not necessary ?

jbrnzl (changes requested):

lib/widgets/video_player.dart
nqhhdev: pauseUponEnteringBackgroundMode: true — default is true => removed
resumeUponEnteringForegroundMode: true — default is false => reverted

nqhhdev: FYI @jbrnzl

jbrnzl (approved):

jbrnzl (approved):

PR #3085, opened 2026-05-27
TW-3048(feat): visio link integration

### LINKED ISSUE #3048: In-App Call Integration - Part 1 - Client Side Link Generation + Redirection
# About

The first version of Visio integration relying on a web based LiveKit client and client side generated link slug.

## DoD

- [ ] App fetches LiveKit config from well-known
- [ ] Call button is visible only if config is found
- [ ] When clicking the button, a message is sent with the concatenation of `base_url` from config and a visio slug generated.
- [ ] The "call" message is detected and a special UI is used for that bubble

# User story

In the top bar of every chat there is a camera button. 

When user clicks on this button, a unique link to a Visio meeting is generated, and an event is sent to the chat that looks like a bubble with two buttons: ‘Copy link’ and ‘Join video call’. 

If there’s been an error in link generation, user sees an error message ‘Failed to start the call’

The text includes the display name of the person that generated the link and the text 'Has started a video call'

Clicking the button ‘Copy link’ copies the link to clipboard, there’s a toast message 

Clicking the button ‘Join video call’ opens the Visio link in new tab 

Users can open context menu for this bubble. The context menu includes options: Forward, Report, Pin, Select, Delete (if role permissions allow)

The event is displayed in the chat list as 'DisplayName has started a video call' 

The event is displayed in notifications as 'DisplayName has started a video call' 

Image

# Design 

https://www.figma.com/design/uLw3iWnOEGaRlJnfPz4Pv0/Twake--Matrix-?node-id=41095-17230&t=gmx3CoXDnZ0IZjLh-1

jbrnzl:
Fixes #3048 

# Description

In the top bar of every chat there is a new camera button: when user clicks on this button, a unique link to a Visio meeting is generated, and an event is sent to the chat that looks like a bubble with two buttons: ‘Copy link’ and ‘Join video call’.

### Tasks

- [ ] Get the configuration from the well-known
- [ ] Display camera button only if config is found
- [ ] Generate visio slug following format `abc-defg-hij`
- [ ] Send a `m.text` message with an additional `callUrlKey` field with the visio link (visio link is also set in the message body)
- [ ] Display a new UI for the "call message"

### Well-Known Config

Because this step one will be followed by deeper LiveKit on Matrix integration, we follow the configuration format from day one.
So the well-known will present a `org.matrix.msc4143.rtc_foci` **array** containing one or many transport configs.

Our app must use the `livekit_base_url` key to build the visio URL to be sent.

```json
"org.matrix.msc4143.rtc_foci": [
  {
    "type": "livekit",
    "livekit_base_url": "https://meet.twake.app"
  }
]
```

If multiple entries are valid the app can use any of them (e.g. first found).

# Test recommendations

- Start a call from a DM: camera icon visible, message sent, bubble renders with "Copy link" + "Join call" : both actions work (copy shows snackbar, join opens in browser).
- Same in group rooms and spaces.
- Paste a `https://meet.twake.app/abc-defg-hij` shaped URL manually (the one in your env):  should render as a video-call card.
- Paste a meet URL with an invalid slug (`https://meet.twake.app/foo`):  should render as a plain text link, not a card.
- Receive the message from another client (Element web): other client shows plain text + link, our app shows the rich card.

# Pre-merge

- [x] https://github.com/linagora/linagora-design-flutter/pull/71 should be merged

# Demo

https://jbournazel-drive.twake.linagora.com/public?sharecode=5XELrPDkf4vG#/


assets/l10n/intl_en.arb
```diff
@@ -3208,5 +3208,20 @@
       "successCount": {},
       "totalCount": {}
     }
+  },
+  "videoCallStartedTitle": "Has started a video call",
```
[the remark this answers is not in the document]
jbrnzl: This wording is intentional we already have the sender's display name shown right above the message bubble

lib/utils/voip/video_call_helper.dart
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  static String generateUrl() {
+    final random = Random.secure();
+    String letters(int length) => List.generate(
+      length,
+      (_) => _slugAlphabet[random.nextInt(_slugAlphabet.length)],
+    ).join();
+    final slug = '${letters(3)}-${letters(4)}-${letters(3)}';
+    return '${AppConfig.videoCallBaseUrl}/$slug';
```
9clg6, line 32: Why the URL is not generated server-side ?

lib/presentation/mixins/start_video_call_mixin.dart
```diff
@@ -0,0 +1,16 @@
+import 'dart:async';
+
+import 'package:fluffychat/generated/l10n/app_localizations.dart';
+import 'package:fluffychat/pages/chat/events/video_call_message.dart';
+import 'package:flutter/widgets.dart';
+import 'package:matrix/matrix.dart';
+
+mixin StartVideoCallMixin {
+  void startVideoCall({required Room? room, required BuildContext context}) {
```
9clg6: BuildContext not necessary in parameters + context shouldn't be passed into business logic

lib/presentation/mixins/start_video_call_mixin.dart
```diff
@@ -0,0 +1,16 @@
+import 'dart:async';
+
+import 'package:fluffychat/generated/l10n/app_localizations.dart';
+import 'package:fluffychat/pages/chat/events/video_call_message.dart';
+import 'package:flutter/widgets.dart';
+import 'package:matrix/matrix.dart';
+
+mixin StartVideoCallMixin {
```
9clg6: This controller already mixes in 20+ mixins — each one adds a responsibility that could live in its own service. A stateless single-method mixin like StartVideoCallMixin is the clearest example: it's just a function call to room.sendTextEvent(...), it doesn't need access to this or widget state. I'd prefer a service/helper here rather than growing the mixin list further. (I know this is a broader pattern issue, but let's at least not make it worse.)

lib/utils/voip/video_call_helper.dart
jbrnzl: Thank you for this review !! Honestly this is the v1 of the video call feature so I wanted to keep it simple at first. The link generation is just a slug appended to a base URL: no network call, latency, auth/error handling, and no extra backend endpoint to maintain for that simple thing. Let me know if you still have concerns about it.

lib/presentation/mixins/start_video_call_mixin.dart
jbrnzl: I couldn't agree more. I updated let me know if it is better ?

lib/pages/chat/events/video_call_content.dart
```diff
@@ -0,0 +1,73 @@
...
+              horizontal: LinagoraSpacing.base,
+            ),
+            child: OverflowBar(
+              spacing: LinagoraSpacing.base,
+              overflowSpacing: LinagoraSpacing.base,
+              overflowAlignment: OverflowBarAlignment.start,
+              children: [
+                LinagoraButton(
+                  label: l10n.videoCallCopyLink,
+                  icon: Icons.open_in_new,
```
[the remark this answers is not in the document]
jbrnzl: This choice is validated by the design and team. No need to address

lib/presentation/mixins/start_video_call_mixin.dart
```diff
@@ -0,0 +1,16 @@
...
+  void startVideoCall({required Room? room, required BuildContext context}) {
+    if (room == null) return;
+    final url = VideoCallMessage.generateUrl();
+    unawaited(
+      room.sendTextEvent('${L10n.of(context)!.videoCallStartedTitle} $url'),
+    );
```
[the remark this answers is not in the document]
jbrnzl: Adressed

lib/utils/voip/video_call_helper.dart
Crash--: At the end, it'll not be a call to generate a URL, it'll be a call to the platform with the intent "I want to create a video call" and the platform will resolve this intent to something and give you back the link to use. 

So I'm more than OK with this first version 👍

lib/config/app_config.dart
```diff
@@ -69,6 +69,8 @@ abstract class AppConfig {
   static String twakeChatGooglePlay =
       'https://play.google.com/store/apps/details?id=app.twake.android.chat';
 
+  static String videoCallBaseUrl = 'https://meet.linagora.com';
```
Crash--: Nein. This variable should come from the .env or something dynamically generated by the environment. 

Because ok for meet.linagora.com if you're linagorian user, but if you're a twake user it'll be meet.twake.app, if you're another... xxx.yyy.fr

lib/pages/chat/chat_view.dart
```diff
@@ -160,6 +161,18 @@ class ChatView extends StatelessWidget with MessageContentMixin {
                             onPressed: controller.toggleSearch,
                             icon: const Icon(Icons.search),
                           ),
+                          if (controller.canSendMessages)
```
Crash--, line 164: I think you should check if the MEET_URL (or something like that) is defined in the .env file. If no, you should not display the button.

lib/config/app_config.dart
Crash--: cc @pm-McFly

lib/pages/chat/chat_view.dart
jbrnzl: @Crash-- Shouldn't we put meet.twake.app as the default MEET_URL value if not defined by the env ?

lib/utils/voip/video_call_helper.dart
9clg6: Okay !

lib/pages/chat/events/video_call_message.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:async';
+import 'dart:math';
+
+import 'package:fluffychat/config/app_config.dart';
+import 'package:matrix/matrix.dart';
+
+class VideoCallMessage {
```
9clg6: VideoCallMessage is a utility class located in the events/ folder, but the other files in that folder are widgets, which VideoCallMessage is not. It would be better suited to a file like “lib/utils/voip/video_call_helper.dart” and should be renamed VideoCallHelper.
What do you think?

lib/config/app_config.dart
pm-McFly: We've checked that point with @jbrnzl yup!
The "hardcoded" or default value that will be bundled in the app for now would be as for the others default: https://meet.twake.app/ (e.g. likewise, the default "hardcoded" value for the homeserver is https://matrix.twake.app/)

For v1.1 the url will come from well-known, until we clear-out the gutter for v2 that should be the MX Native LiveKit calls

lib/pages/chat/events/video_call_message.dart
```diff
@@ -0,0 +1,34 @@
...
+  const VideoCallMessage._();
+
+  static const _slugAlphabet = 'abcdefghijklmnopqrstuvwxyz';
+
+  static final RegExp _urlRegExp = RegExp(
+    '${RegExp.escape(AppConfig.videoCallBaseUrl)}/[a-z]{3}-[a-z]{4}-[a-z]{3}(?=\\s|\$)',
```
nqhhdev: The regex is built once at class-load time from AppConfig.videoCallBaseUrl. Two consequences:
- if the config value changes at runtime, the regex won't update.
- the pattern only matches the xxx-xxxx-xxx slug format this app generates—a user manually sharing any other Jitsi room URL (e.g. /my-custom-room) will see plain text, not the call card. Is that intentional scoping?

lib/utils/voip/video_call_helper.dart
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  static void start({required Room? room, required String startedTitle}) {
+    if (room == null) return;
+    final url = generateUrl();
+    unawaited(room.sendTextEvent('$startedTitle $url'));
+  }
```
nqhhdev, line 45: - Silently drops any send error (network failure, power level rejection, etc.). The user taps the button and gets no feedback if it fails. At minimum the error should be logged; ideally the caller should receive a callback or the error should surface via a snackbar.

lib/pages/chat/events/video_call_message.dart
jbrnzl: Sounds way better ! Thank you, I updated !

lib/pages/chat/events/video_call_message.dart
jbrnzl: Hello @nqhhdev thank you for the review ! I think it is not a big issue with the regex current behavior. The only edge-case is the config being edited while the app is running. For me it is acceptable. Maybe we can ask @pm-McFly to confirm ?
Regarding what the pattern is matching, I can confirm this is for now intentional:

- Message contains AppConfig.videoCallBaseUrl + xxx-xxxx-xxx slug => video card
- Message contains AppConfig.videoCallBaseUrl + anything else => Raw message content

nqhhdev: @pm-McFly Help us plsss

lib/utils/voip/video_call_helper.dart
jbrnzl: thanks @nqhhdev I think it is OK because the matrix sdk already handle everything for us (stop me if I'm wrong). We insert an event in the timeline, if network is not available yet we will have a sending state, and if failed to be sent we will have the feedback in the event itself (error icon+retry). Like the others messages:
Capture d’écran 2026-06-01 à 11 27 54
Capture d’écran 2026-06-01 à 11 30 32
Regarding the power level we have `controller.canSendMessages` Do you think it is enought ?

lib/pages/chat/chat_view.dart
Crash--: I don't think so. Or if we do that we should add a conf to disable the button.

Crash--: how it is solved? I see anything?

jbrnzl: https://github.com/linagora/twake-on-matrix/pull/3085#discussion_r3332546798
Ok, I got confused because we said that `meet.twake.app` will be the default

Crash--: In some env we’ll not have meet. So please even in V1 consider something like this aka « being able to disable video call ». And not relying on the well known directly is a mistake. I don’t understand why we have made this choice. It doesn’t add a lot of complexity. It doesn’t add a new concept. It just reuse what we already have.

lib/utils/voip/video_call_helper.dart
nqhhdev: Nice 👌

nqhhdev:
- [x] Should check edited message with the Visio link.
 
Screenshot 2026-06-02 at 15 59 16

lib/pages/chat/chat_view.dart
pm-McFly: So, as checked together the official livekit integration tends to use either:
- GET `${homeserver}/_matrix/client/v1/rtc/transports`
- well-known entry: `org.matrix.msc4143.rtc_foci`

We tried to make the GET route work, but no results yet...
Probably not yet implemented as the MSC is not validated yet.

Though, matrix.org is providing the Element Calls on LiveKit and looking at their WK the entry is set - so we can assume that lead is correct to use :)

The WK `org.matrix.msc4143.rtc_foci` is defined as: https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md#discovery-of-rtc-transports

Example from matrix.org:
```json
"org.matrix.msc4143.rtc_foci": [
  {
    "type": "livekit",
    "livekit_service_url": "https://livekit-jwt.call.matrix.org"
  }
]
```

We are not yet to rely on a JWT LiveKit Service and we should not advertise it to not brick other clients.
So we will add to our WK the following entry:

```json
"org.matrix.msc4143.rtc_foci": [
  {
    "type": "livekit",
    "livekit_base_url": "https://meet.twake.app"
  }
]
```

The v1 of our integration will thus append the visio slug `abc-defg-hij` to the `livekit_base_url`, but the corner stone would already be there to live on.

@jbrnzl aye? cc/fyi. @Crash--

lib/pages/chat/events/video_call_message.dart
pm-McFly: @nqhhdev, @jbrnzl a new direction has been found fo the message format, maybe we can resolve that discussion?

jbrnzl: @pm-McFly Sure, we can close it 👍🏻

jbrnzl:
> * [x]  Should check edited message with the Visio link.
> 
> Screenshot 2026-06-02 at 15 59 16

I asked @tprudentova to check the video demo of the last behavior and she is OK with it, so we are good on this point :)

lib/pages/chat/chat_view.dart
Crash--: I don't care about the native integration, we're not yet there. It'll be the v2. 

But I need, for v1, to have a way to configure / disable the livekit_url. So why can't we just expose the livekit_url from the well-known and if not present, then we do not display the button.

nqhhdev:
> > * [x]   Should check edited message with the Visio link.
> > 
> > Screenshot 2026-06-02 at 15 59 16
> 
> I asked @tprudentova to check the video demo of the last behavior and she is OK with it, so we are good on this point :)

Resolve conflicts first, pls

lib/config/app_config.dart
pm-McFly: Well-Known details: https://github.com/linagora/twake-on-matrix/pull/3085#discussion_r3342043490

lib/pages/chat/chat_view.dart
pm-McFly: > But I need, for v1, to have a way to configure / disable the livekit_url. So why can't we just expose the livekit_url from the well-known and if not present, then we do not display the button.

Yup exactly!
The JSON above is the well-known (WK) expected content. We keep the official format, but instead of `livekit_service_url` we use `livekit_base_url`.

jbrnzl:
Follow-up: #3101

Issue #3078, opened 2026-05-26
[BUG] [iOS] Can't paste link into composer


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Action 'Paste' not displayed in the composer when you copy a link from the browser 

https://github.com/user-attachments/assets/e0127699-f3b6-4932-a4d2-30078711bcf6

### Pre-Bug State

N/A

### To Reproduce

1. Copy a link using the Copy function of the browser 
2. Go to TC 
3. Long-tap to call action menu 
4. Action Paste isn't there

### Expected results

There's a paste action if the link is copied to clipboard

### Context

- Device: iPhone
- OS: iOS
- Browser: [e.g. chrome, safari]
- App Version: 2.22.2
- Server: linagora, staging
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
Closed by #3102

PR #3102, opened 2026-06-04
TW-3078: enhance clipboard paste functionality for iOS to support URLs

nqhhdev:
## Ticket
**Related issue**

- #3078 

Issue: On iOS, copying a URL from Safari/Chrome and long-tapping the composer showed no "Paste" button. If paste was triggered anyway, the app crashed.                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                
  Why it broke across 3 places:                                                               
                                                                                                                                                                                                                                                                                                                                                                                
  1. context_menu_builder_ios_paste_without_permission.dart — SystemContextMenu (iOS 16+) only shows "Paste" when the clipboard holds public.plain-text. A browser URL is stored as public.url, so the native menu silently omitted the button. AdaptiveTextSelectionToolbar had the same blind spot since it checks clipboard status via Clipboard.getData('text/plain') which 
  also misses public.url. Fixed by probing contextMenuButtonItems — if Paste is absent, inject it manually. On iOS 16+ SystemContextMenu is still used when Paste is already present to preserve the no-permission-alert benefit; only falls back to AdaptiveTextSelectionToolbar when we had to inject Paste ourselves.                                                      
                                                                                                                                                                                                                                                                                                                                                                                
  2. clipboard.dart — TwakeClipboard.pasteText() only tried Formats.plainText. When the clipboard held a URL, getValue returned null and the code threw an error instead of trying the next format. Fixed by falling back to Formats.uri which super_clipboard already supports for public.url on iOS, converting the result to a string. Returns null gracefully if neither    
  format is available.                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                
  3. text_editting_controller_extension.dart — After inserting pasted text, the cursor was set to end + pastedText.length using the old end offset from before the text was modified. For a range selection this could exceed the new text length causing a crash. Fixed by unifying all paste cases into replaceRange(start, end, pastedText) with the cursor always placed at 
  start + pastedText.length, which is always valid.

## Resolved
- Web:
- Android:
- IOS:


https://github.com/user-attachments/assets/a03a7bc4-f02e-4c8e-920c-afcdc7e62423



## Summary by CodeRabbit

* **Bug Fixes**
  * Restored/injected a visible Paste option in the iOS context menu when missing.
  * Clipboard paste now falls back to URL data when plain text is unavailable.

* **Refactor**
  * Simplified and clarified text-paste insertion flow for more predictable cursor placement and reliability.

### DIFF
--- a/lib/presentation/extensions/text_editting_controller_extension.dart
+++ b/lib/presentation/extensions/text_editting_controller_extension.dart
@@ -10,21 +10,17 @@ extension TextEdittingControllerExtension on TextEditingController {
     final pastedText = await TwakeClipboard.instance.pasteText(
       clipboardReader: clipboardReader,
     );
-    if (pastedText != null) {
-      if (start == -1 || end == -1) {
-        text = pastedText + text;
-        selection = TextSelection.collapsed(offset: text.length);
-        return;
-      }
-      if (start == end) {
-        final startText = text.substring(0, start);
-        final trailingText = text.substring(end, text.length);
-        text = startText + pastedText + trailingText;
-      } else {
-        text = text.replaceRange(start, end, pastedText);
-      }
-      selection = TextSelection.collapsed(offset: end + pastedText.length);
+    if (pastedText == null) return;
+
+    if (start == -1 || end == -1) {
+      text = pastedText + text;
+      selection = TextSelection.collapsed(offset: pastedText.length);
+      return;
     }
+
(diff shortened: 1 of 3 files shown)


9clg6 (approved):

jbrnzl (approved):
I tried and it works ! Well done 💪🏻

Issue #3099, opened 2026-06-03
Contacts Views - Unify All Contacts Views Content


pm-McFly:
# Description

The contact list is showed at different places across the app, but it's content is sparsed:
- Contacts page
- DM creation
- Group creation
- Adding member to a room

Sometimes including local phone addressbook entries, sometimes showing only ToM entries, etc.

The idea of the present issue is to discuss a manner for unifying all these views and lists.
Maybe one widget List can be re-used and filled with information from a contacts store? Maybe we must compute it each time depending on the context?

Open topic yet for discussion

Issue #3098, opened 2026-06-03
Contact Page - UI Revamp


pm-McFly:
# Description

As briefly discussed in #3080 the Contacts page UI has to be updated to correctly reflect different cases where it is empty:

- No ToM Server available; *i.e page cannot be used*
  - How to hint the user?
- ToM Server available + No entries in Addressbook Endpoint + Invitation **not** allowed
  - Contact can be created manually
    - How to hint the manual creation?
  - Contact can be found using Contacts Sync if "Contacts" permission is granted
    - How to ask user for Contacts permission?
- ToM Server available + No entries in Addressbook Endpoint + Invitation allowed
  - Contact can be found using Contacts Sync if "Contacts" permission is granted
    - How to ask user for Contacts permission?
  - Contacts in phone address book can be invited
    - How to hint user that they can invite their own contacts?
- ToM Server available + No entries in Addressbook Endpoint + No contacts in phone
  - How to show the empty contacts page?

pm-McFly:
@9clg6 This issue is mostly there to remind @tprudentova about the UI/UX work remaining on the "Contacts" feature ;)

Issue #3080, opened 2026-05-26
[BUG] Contacts Page Should be Dedicated to ToM Address Book (+ invitations)


pm-McFly:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The "Contacts" page contains all entries from:
1. DMs found by the SDK
2. ToM Address Book endpoint
3. Local Phone Address Book

When in a "company environment" - i.e. using a ToM Server configured to compute the Address Book by merging personnal entries and LDAP provided users - everything is mixed and when a user departs the company suddenly its associated entry jumps to the top of the page if one had a DMs with them. Making the departed user "more visible" although they just left.

As a "Contact" is a concept brought by the Twake plateform, today, the "Contacts" page should reflect that.

### Pre-Bug State

.

### To Reproduce

On a "Company environment" (e.g. B2B SaaS or LINAGORA internal), create a DM with someone outside the company and/or with someone inside the company.

1. Open the "Contacts" page
2. External (i.e. federated) DM users will show at the top of the page
3. Former company member DMs will show at the top as well

### Expected results

- Open the "Contacts" page
- See an ordered list of contacts sorted by Display Name
  - Contacts are only company provided
  - AND, contacts found by synchronization that are automatically added to my ToM Address Book
  - AND, ones that I added manually to my ToM Address Book
 
- If the well-known indicates `app.twake.chat.enable_invitations: true`
  - also list local phone address book entries with the "invite" button

### Context

.

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

QA Env is configured as Company Featured Env. It is able to chat with LINAGORA. When chatting with someone from LNG that DM appears at the TOP of the list although that mxid is not in my contacts (_synched or manually added_).

<img width="463" height="597" alt="Image" src="https://github.com/user-attachments/assets/4ba63828-f96d-4d4a-b919-7ef5aa19e1d6" />

</details>

pm-McFly:
cc: @tprudentova
Can you give me your 2 cts about this, please?

tprudentova:
@pm-McFly yes, I completely agree with only showing the company contacts, contacts found though sync/people you can invite (if applicable) and people added manually

pm-McFly:
Can you check with our design team to have a visual for an empty page that hints "use our Twake augmented environnement to benefit a Contacts page" kinda message? 
You rock 🤘

tprudentova:
[...]

PR #3105, opened 2026-06-04
fix(contacts): remove SDK DM entries from Contacts page

### LINKED ISSUE #3097: Contacts Page - Remove DMs entries
# Description

When accessing the Contacts Page, if one has DMs with people they appear at the very top of the Contact List.

These DMs entries are now to be removed to avoid mixed usages of the Contacts Page; it is not because I speak to one that I want them easily accessible.

The Contact Page must **ONLY** list entries comming from ToM Addressbook endpoint, containing only the user's true contacts.

# DoD

- [ ] No more DMs entries provided by the SDK

9clg6:
## Context

Closes #3097 (sub-issue of #3080).

The Contacts page mixed three sources: SDK direct chats (DMs), the ToM Address Book, and the local phone book. In a company environment, the recent-DMs section pinned federated users or former company members to the **top** of the list — making a departed user *more* visible right after they left.

The Contacts page is a Twake concept dedicated to the **ToM Address Book**. DMs found by the SDK should not be listed there.

## Change

- Add an overridable `enableRecentContacts` getter on `ContactsViewControllerMixin` (default `true`, so DM creation / group creation / add-member views keep their recent contacts).
- Override it to `false` in `ContactsTabController` → `_refreshRecentContacts` is skipped and `presentationRecentContactNotifier` stays empty.
- Remove the now-dead `SliverRecentContacts` widget from the page body.

Because the recent notifier stays empty, `SliverEmptyContacts` (which gates on `_hasRecentData()`) still behaves correctly and the page is never left blank.

## Scope

Only the Contacts page changes. Other contact lists (DM creation, group creation, add member to room) keep recents via the unchanged default getter.

Empty-state UI revamp (#3098) and view unification (#3099) are out of scope.

IMG_5811
IMG_5812
IMG_5814
IMG_5815

## Test

- `flutter analyze` clean.
- `test/mixin/contacts_view_controller_mixin_test.dart` passes (+18).
- Verified on a physical device (iOS release): the Contacts page no longer shows the "Recent" section — the list is the alphabetical ToM Address Book only.

## Summary by CodeRabbit

* **Refactor**
  * Removed the recent contacts section from the contacts page
  * The contacts view now displays only Address Book entries instead of mixing in recently discovered contacts
  * Previously accessed or automatically discovered contacts no longer appear
  * Contacts displayed are limited exclusively to those saved in your Address Book
  * The interface now focuses solely on Address Book entries


nqhhdev:
Thanks for your contribution. I have one concern:

- The Contacts page is meant to reflect the ToM Address Book exclusively. Before this fix, `_refreshRecentContacts` was always called, which mixed in SDK-discovered DMs (direct chat rooms) alongside Address Book entries. This caused departed users or federated contacts to appear at the top of the list because they had an active DM, even though they are no longer in the company directory. So we added an overridable enableRecentContacts getter on `ContactsViewControllerMixin` (defaults to true so DM/group creation views keep showing recents), and `ContactsTabController` overrides it to false. The `SliverRecentContacts` widget is also removed from the page body since it has no data source on this page anymore.

What do you think?

9clg6:
> * The Contacts page is meant to reflect the

Agreed, that's exactly the intent of this PR. Scoping the Contacts page to the Address Book only, while keeping enableRecentContacts overridable so DM/group creation still show recents. Thanks for confirming the approach.

jbrnzl (approved):

nqhhdev (approved):
LGTM

PR #3111, opened 2026-06-08
fix(chat): prefill caption with composer text when pasting an image

9clg6:
## Problem

When you type a message and then **paste an image** (Ctrl+V on web), the caption dialog opens **empty** — the text you were typing is lost from the caption.

TW-2750 (#2751) added "attach media to the text already typed in the composer", but only wired `pendingText` into the **file-picker** and **drag-and-drop** flows. The **paste-image** path (`pasteImage`) was missed.

## Fix

- Thread `pendingText` through `pasteImage` into `SendFileDialog`, matching the other media flows.
- On successful send, clear the composer + remove the saved draft (the typed text now travels as the image caption, so it must not also remain in the composer / be re-sent as a separate message).
- Text is **preserved** if the dialog is cancelled or the paste fails (composer only clears on `done`).

## Scope

Paste-image is web-only (`pasteImagesUsingBytes` is guarded by `PlatformInfos.isWeb`). File-picker and drag-and-drop already worked.

## Test plan

1. Open a chat on web.
2. Type some text in the composer.
3. Paste an image (Ctrl/Cmd+V).
4. ✅ Caption field is pre-filled with the typed text.
5. Send → image sent with caption, composer is empty.
6. Repeat, but **cancel** the dialog → typed text is still in the composer.

image
image

image

## Summary by CodeRabbit

* **New Features**
  * Drafts are now persisted per chat so unsent message text is reliably saved and restored.
* **Bug Fixes**
  * Pasting images preserves current message text during the paste/send flow.
  * Drafts are automatically cleared after successful send or media/camera send completion, preventing stale drafts.
* **Tests**
  * Added tests verifying save/read/remove behavior and per-room draft isolation.

### DIFF
--- a/lib/presentation/mixins/handle_clipboard_action_mixin.dart
+++ b/lib/presentation/mixins/handle_clipboard_action_mixin.dart
@@ -1,9 +1,9 @@
 import 'package:fluffychat/presentation/extensions/text_editting_controller_extension.dart';
 import 'package:fluffychat/presentation/mixins/paste_image_mixin.dart';
 import 'package:fluffychat/utils/clipboard.dart';
+import 'package:fluffychat/utils/room_draft_storage.dart';
 import 'package:flutter/material.dart';
 import 'package:matrix/matrix.dart';
-import 'package:shared_preferences/shared_preferences.dart';
 import 'package:super_clipboard/super_clipboard.dart';
 
 mixin HandleClipboardActionMixin on PasteImageMixin {
@@ -14,6 +14,8 @@ mixin HandleClipboardActionMixin on PasteImageMixin {
   Room? get room;
 
   TextEditingController get sendController;
+
+  final RoomDraftStorage _draftStorage = const RoomDraftStorage();
 
   void registerPasteShortcutListeners() {
     ClipboardEvents.instance?.registerPasteEventListener(_onPasteEvent);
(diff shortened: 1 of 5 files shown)


lib/presentation/mixins/handle_clipboard_action_mixin.dart
```diff
@@ -37,11 +38,18 @@ mixin HandleClipboardActionMixin on PasteImageMixin {
...
     return pasteImage(
       context,
       room!,
       clipboardReader: clipboardReader,
-      onSendFileCallback: onSendFileCallback,
+      pendingText: pendingText,
+      onSendFileCallback: () async {
+        sendController.clear();
+        final prefs = await SharedPreferences.getInstance();
+        await prefs.remove('draft_${room!.id}');
```
nqhhdev: Should separate small func so we can resue it.

9clg6: Done in 51fff241e. Extracted into a RoomDraftStorage class (read/save/remove) so the draft_<roomId> logic is reused here and across chat.dart instead of being duplicated.

jbrnzl (approved):
Nice!

nqhhdev (approved):
LGTM

PR #3104, opened 2026-06-04
feat: native OS photo picker in chat (iOS PHPicker + Android Photo Picker)

9clg6:
## What

The chat media picker now opens the **OS-native gallery picker** on **both iOS and Android** (WhatsApp-style), replacing the custom `photo_manager` in-app grid on mobile.

- **iOS** → `PHPickerViewController`
- **Android** → the modern **Android Photo Picker** (`ACTION_PICK_IMAGES`)
- System albums + search + multi-select
- **No photo-library permission prompt** — both pickers run out-of-process
- **Web unchanged** — keeps the existing file/web pipeline

The mobile attachment bar is also revamped: a `+` popup menu (Documents / Camera) plus a dedicated gallery button.

## How

- New `SendFilesMixin.pickAndSendMediaNative()` opens `image_picker.pickMultipleMedia()`, maps each `XFile` to a typed `ImageFileInfo`/`VideoFileInfo` (with `customMimeType`), and pushes them through the existing `UploadManager.uploadFileMobile()` mobile pipeline.
- **Android Photo Picker opt-in**: `image_picker` defaults to the legacy `ACTION_GET_CONTENT` (file/SAF explorer) intent. `main.dart` now sets `ImagePickerAndroid.useAndroidPhotoPicker = true` (Android-only, one-time at startup) so Android gets the modern Photo Picker UI instead of the document explorer. Added `image_picker_android` + `image_picker_platform_interface` as direct deps (already transitive) to reference the flag.
- The downstream pipeline (thumbnail, image compression, **HEIC→JPG**, encryption, Matrix `m.image`/`m.video` event) is **reused as-is**. Returning typed `ImageFileInfo`/`VideoFileInfo` (not a bare `FileInfo`) is required so `sendFileEventMobile` triggers HEIC conversion + thumbnail, which are gated on the concrete type.
- `pickMultipleMedia()` wrapped in try/catch for `PlatformException`.
- **Web-narrow guard**: the mobile input bar is selected by viewport width (`responsiveUtils.isMobile`), not platform. On a narrow web viewport that mobile branch is still rendered, but the native picker feeds the mobile (`dart:io`) pipeline, which breaks on web. `chat_input_row.dart` now falls back to the single `+` button routed through `onSendFileClick`'s platform-aware path when `!PlatformInfos.isMobile`.

## Tests

`test/presentation/mixins/send_files_mixin_test.dart` — unit tests:
[...]

PR #3116, opened 2026-06-09
refactor(chat): centralize room draft persistence in RoomDraftStorage

9clg6:
## What

Extracts a small `RoomDraftStorage` helper that owns the `draft_<roomId>` SharedPreferences key and the read/save/remove operations.

## Why

The key `'draft_$roomId'` was hand-built at six call sites in `ChatController` (load draft, send, debounced input-change save, media-send cleanup). Centralizing it removes the duplication and keeps the storage format in one place.

## How

- New `lib/utils/room_draft_storage.dart` — `read` / `save` / `remove` keyed by room id.
- `ChatController` routes every draft access through a `RoomDraftStorage` field; removed the now-unused `shared_preferences` import.
- Added `test/utils/room_draft_storage_test.dart` (round-trip, key format, remove, per-room isolation).

## Notes

- **No behavior change.**
- Split out of #3104 (native photo picker) as an unrelated cleanup.

`flutter analyze` clean, tests green.

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -62,6 +62,7 @@ import 'package:fluffychat/presentation/model/forward/forward_argument.dart';
 import 'package:fluffychat/resource/image_paths.dart';
 import 'package:fluffychat/utils/account_bundles.dart';
 import 'package:fluffychat/utils/adaptive_bottom_sheet.dart';
+import 'package:fluffychat/utils/room_draft_storage.dart';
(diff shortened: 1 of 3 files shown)


9clg6:
Closing — duplicate of #3111 (fix/paste-image-caption), which already introduces RoomDraftStorage. I created this before realizing the refactor already had a home there. The only extra here was a unit test for RoomDraftStorage; can add it to #3111 if wanted.

PR #3139, opened 2026-06-12
TW-2968(fix): wrong member count in group chat

### LINKED ISSUE #2968: Wrong member count in group chat
2.11.1, [[linagora.com](http://linagora.com/)](http://linagora.com), IOS 

The counter of members says there’s only 1 member in a chat, though there are a lllot of members there

Image

Update 26/03/2026

Diana got this issue too in the latest version. I think it happens in older chats in which we have this bug with all members displayed as “read-only” (maybe in member count we don’t include read-only users? But I have such chats on my side and the member count is correct…)

Image

Image

jbrnzl:
## Ticket

Fixes #2968 

## Root cause & Context

We previously added a sentry log #2970 member to follow #2968 and I was wondering if we still had the issue or not and we have.
I made some investigation and indeed the member count comes from the room summary, and it looks like servers only send `m.joined_member_count` `m.invited_member_count` in /sync when they change. If one update is missed (gapped sync, app killed mid-sync), the persisted counts stay wrong forever: nothing ever reconciles them with the member list.
I could experiment this case with few rooms/accounts:
photo_2026-06-12_17-20-16 photo_2026-06-12_17-20-24

And we can verify it debugging (the room in the last screenshot with 4 members):
As you can see matrix-dart-sdk is giving us 0 and 1 but the total of the members is 4.

Capture d’écran 2026-06-12 à 15 03 20

## Changes

New `requestParticipantsWithSummaryReconciliation` extension on `Room`: after fetching the participants, if the real counts differ from the summary, the corrected values are re-injected through the sync pipeline (memory + database + UI). Called when opening a chat. Each healing is logged with the existing `Wrong-member-count` Sentry tag so we can follow the fix in production.

##  Concerns

I know this is not a perfect fix as the bug is an edge case where some informations were not stored as I said in the "Root cause" but this is a try fixing the UI client side. We should not rely on self-syncing but this is the only path I found to fix it. Maybe a more complete fix should be handled server side or within the matrix-dart-sdk but I am not sure how exatcly for now. Please feel free to help me make sure we can safely introduce this fix by testing or discussing it.

## Test recommendations

Unit tests cover the reconciliation and its guards. Manually: open a group whose displayed count is wrong, then the details page, the count should match the member list, with a `Wrong-member-count: healing summary` line in the logs.

## Release Notes

* **Bug Fixes**
  * Improved chat room member count accuracy by automatically reconciling stored summary counts against the actual participant membership when requesting members (with reconciliation limited to join/invite scenarios and avoiding incorrect updates in edge cases).

* **Tests**
  * Added/expanded automated tests to verify reconciliation updates are applied only when needed, persisted exactly once when stale, and skipped when counts are already correct or reconciliation shouldn’t run.
  * Updated test client setup to allow optional database injection and streamlined transaction behavior.

### DIFF
--- a/lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
@@ -1,11 +1,10 @@
 import 'package:fluffychat/utils/logging/sentry_tracked_events.dart';
 import 'package:matrix/matrix.dart';
 
-/// Fix stale room summary member counts.
-///
-/// Servers only send `joined_member_count` and `invited_member_count` in
-/// /sync when they change: one missed delta leaves the persisted counts
-/// wrong forever, as nothing ever reconciles them with the member list.
+/// TODO(TW-2968): client-side workaround for a Synapse bug that omits
+/// member-count deltas — https://github.com/element-hq/synapse/issues/19865
+/// Remove this reconciliation once Synapse is fixed (keep the
+/// `Wrong-member-count` Sentry tag to confirm the fix landed).
 extension RoomMemberCountReconciliationExtension on Room {
   /// Only reconciles when [membershipFilter] contains both [Membership.join]
   /// and [Membership.invite]: otherwise the list is not usable.
--- a/lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
@@ -1,0 +1,82 @@
+import 'package:fluffychat/utils/logging/sentry_tracked_events.dart';
+import 'package:matrix/matrix.dart';
+
+/// Fix stale room summary member counts.
+///
+/// Servers only send `joined_member_count` and `invited_member_count` in
+/// /sync when they change: one missed delta leaves the persisted counts
+/// wrong forever, as nothing ever reconciles them with the member list.
+extension RoomMemberCountReconciliationExtension on Room {
+  /// Only reconciles when [membershipFilter] contains both [Membership.join]
+  /// and [Membership.invite]: otherwise the list is not usable.
+  Future<List<User>> requestParticipantsWithSummaryReconciliation([
+    List<Membership> membershipFilter = const [
+      Membership.join,
+      Membership.invite,
+      Membership.knock,
+    ],
+    bool suppressWarning = false,
+    bool? cache,
+  ]) async {
+    final participants = await requestParticipants(
+      membershipFilter,
+      suppressWarning,
+      cache,
+    );
+    if (membershipFilter.contains(Membership.join) &&
+        membershipFilter.contains(Membership.invite)) {
+      await healSummaryMemberCounts(participants);
+    }
+    return participants;
+  }
+
+  /// Updates the summary member counts, in memory and in database, when they
+  /// diverge
+  Future<void> healSummaryMemberCounts(
+    List<User> groundTruthParticipants,
+  ) async {
+    // A synthetic JoinedRoomUpdate on a non-joined or unknown room would
+    // corrupt its membership or restore it in the room list.
+    if (membership != Membership.join || client.getRoomById(id) == null) {
+      return;
+    }
+    final joinedCount = groundTruthParticipants
+        .where((user) => user.membership == Membership.join)
+        .length;
+    final invitedCount = groundTruthParticipants
+        .where((user) => user.membership == Membership.invite)
+        .length;
+    // A joined room always has at least one joined member (the syncing user)
+    if (joinedCount == 0) {
+      return;
+    }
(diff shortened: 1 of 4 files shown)


lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
```diff
@@ -0,0 +1,82 @@
...
+      '$id from (joined=${summary.mJoinedMemberCount}, '
+      'invited=${summary.mInvitedMemberCount}) to (joined=$joinedCount, '
+      'invited=$invitedCount)',
+    );
+    // Feed the corrected counts through the regular sync pipeline, as if the
+    // server had sent them: same merge (heroes preserved), same persistence
+    // and UI notifications as a real summary update.
+    await client.handleSync(
+      SyncUpdate(
+        nextBatch: client.prevBatch ?? '',
```
nqhhdev: If `client.prevBatch` is null, the synthetic SyncUpdate gets nextBatch: ''. The SDK may persist '' as the new sync token, causing the next real /sync to restart from the beginning (full re-sync). Unlikely in practice but a silent data hazard. WDYT?

9clg6:
I think your fix is good. The server only includes m.joined_member_count / m.invited_member_count in /sync when they change, that's spec-compliant, not a server bug. The problem is purely client-side: if we miss that single update (gapped sync, app killed mid-sync), the persisted count stays stale forever, because the server won't resend the summary until the next change. Recomputing/reconciling from the local member list is the right place to fix it.
One thing to double-check: we should only override with the recomputed count when the member list is fully loaded locally, otherwise (lazy-loaded / partial membership) the recount could be lower than reality.

9clg6 (approved):

lib/utils/matrix_sdk_extensions/room_member_count_reconciliation_extension.dart
jbrnzl: Thank you @nqhhdev ! I checked and from what I can see handleSync does not read nextBatch. I think it is only used for real /sync. Btw I changed it to `nextBatch: ''` as in other places in the project. Don't hesitate if you have any other concerns

jbrnzl:
> I think your fix is good. The server only includes m.joined_member_count / m.invited_member_count in /sync when they change, that's spec-compliant, not a server bug. The problem is purely client-side: if we miss that single update (gapped sync, app killed mid-sync), the persisted count stays stale forever, because the server won't resend the summary until the next change. Recomputing/reconciling from the local member list is the right place to fix it. One thing to double-check: we should only override with the recomputed count when the member list is fully loaded locally, otherwise (lazy-loaded / partial membership) the recount could be lower than reality.

Thank your for the review ! Yes, I can confirm it is well handled in requestParticipants (with getMembersByRoom) 🙏🏻

krille-chan:
@jbrnzl this looks actually like you are working around a bug in synapse (or a severe bug in the sdk?) 🤔 while it is the responsibility of the sync loop to always give you an updated count. Do you know how an "missed delta" can happen there and how to reproduce it? Would like to fix the root cause

jbrnzl:
> @jbrnzl this looks actually like you are working around a bug in synapse (or a severe bug in the sdk?) 🤔 while it is the responsibility of the sync loop to always give you an updated count. Do you know how an "missed delta" can happen there and how to reproduce it? Would like to fix the root cause

@krille-chan I think the easier to reproduce this issue is to create a new room from user `@a:hs1` and invite `@b:hs2`.
On `@b:hs2` screen you will see a wrong member count.
Here is the exemple:

https://github.com/user-attachments/assets/2dff9964-76e7-4d84-888d-c7c89d9ee999

Reproductible on FluffyChat and most of matrix clients I think as they behave the same:

photo_2026-06-17_14-22-21

Member count seems to never updates until there is a new `member event` in the room

krille-chan:
@jbrnzl oh wow very good catch. I was able to reproduce it but also could confirm that this is not a bug in the sdk but just coming down from synapse. I have opened a bug report at synapse here: https://github.com/element-hq/synapse/issues/19865

jbrnzl:
I think we can merge this fix event if it seems like a synapse issue. We can still keep track of "wrongMemberCount" on sentry, and see later if we can remove this fix, or if we have an other issue with the member count when the synapse issue is resolved.
What do you think @9clg6 @nqhhdev ?

jbrnzl:
> @jbrnzl oh wow very good catch. I was able to reproduce it but also could confirm that this is not a bug in the sdk but just coming down from synapse. I have opened a bug report at synapse here: [element-hq/synapse#19865](https://github.com/element-hq/synapse/issues/19865)

Thank you very much !

9clg6:
@jbrnzl  Yes we can merge this workaround, tagged with a TODO waiting to a solution from Synapse.

9clg6 (approved):

nqhhdev (approved):

9clg6 (approved):

Issue #3114, opened 2026-06-09
[BUG] Black heart displayed instead of a red heart


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When someone sends a red heart (using native emoji keyboard), it is displayed as black

### Pre-Bug State

.

### To Reproduce

1. Send a red heart or leave it as an emoji 
2. It's displayed as black

### Expected results

The red heart is red

### Context

- Browser: [e.g. chrome, safari]
- App Version: [e.g. 22]
- Server: [e.g. production, example.com]
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
Fixed by https://github.com/linagora/twake-on-matrix/pull/3160

PR #3141, opened 2026-06-15
TW-3114: Partially remove Inter font family from text styles

nqhhdev:
## Ticket
**Related issue**

- #3114

## Resolved

- Changed the font files (Inter-*.ttf) because the original Inter fonts contained emoji glyphs. Even after removing fontFamily: 'Inter' from the code, the font was still registered in pubspec.yaml — so Flutter on web could still pick it up and use it to render emoji, blocking color emoji. Patching the files removes those emoji glyphs from Inter so it can no longer interfere.

- The new fallback fonts (Noto*) were added because after removing Inter, non-Latin scripts (Arabic, Thai, Korean, Chinese, etc.) had no font to fall back to on web — mobile OS provides these natively, but web doesn't. So we bundle them explicitly to ensure those scripts still render correctly.

https://github.com/user-attachments/assets/622591e4-8f77-435b-99b6-f507e986bf89

https://github.com/user-attachments/assets/48c9664f-e6b1-4d2d-8414-da1e02700e3b

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search functionality to consistently handle emoji variation selectors and properly handle null or empty search terms.

* **Chores**
  * Refined font family handling across UI components for improved consistency.
  * Updated Flutter web engine initialization.


assets/google_fonts/fallback/NotoEmoji-Regular.ttf
9clg6, line 1: Are we really using all these fonts ? Can we limit to which we are really using ?

lib/config/themes.dart
```diff
@@ -22,76 +54,51 @@ abstract class TwakeThemes {
...
-      fontWeight: FontWeight.w500,
-      letterSpacing: 0.15,
-    ),
-    titleSmall: TextStyle(
-      fontFamily: 'Inter',
-      fontWeight: FontWeight.w500,
-      letterSpacing: 0.1,
-    ),
+  static const TextTheme fallbackTextTheme = TextTheme(
+    bodyLarge: TextStyle(fontWeight: FontWeight.w500, letterSpacing: -0.15),
```
9clg6, line 27: short-hand could be cool here; but not a must-have

9clg6 (approved):

assets/google_fonts/fallback/NotoEmoji-Regular.ttf
nqhhdev: Removed unused font.

pm-McFly:
Somehting I don't understand is:
- you said the issue is coming from Inter having glyphs for emojis like the heart and thus flutter is using those ones instead of falling back on another font like for the turtle emoji (for example)
- you then removed some stuff from Inter to fix it: ok
- you had to add Roboto font to support non latin glyphs due to missing fallbacks. <- new issue caused by stripping Inter files: ko

Then, what if we place Noto Emoji above Inter in the pub file so web tries to load emojis first and fall back for all chars (latin, tamil, or chinese, etc.) to Inter.
That way no need to change Inter ttf files or to add roboto files. No?

---

And regarding that issue for animated emojis and using Noto Animated Emojis. If that font is added above Noto Emojis:

1. Noto Animated Emojis
2. Noto Emojis
3. Inter

Would allow us to use animation only on bubbles where we only have emojis and use non animated ones on others.

Or is there something I am missing?

pm-McFly:
And otherwise if Inter does not support certain charset, we have to inform our design/product team so we choose another font to avoid having fallbacks that are not compatible with our design system letter spacing and font height size

nqhhdev:
> Somehting I don't understand is:
> 
> * you said the issue is coming from Inter having glyphs for emojis like the heart and thus flutter is using those ones instead of falling back on another font like for the turtle emoji (for example)
> * you then removed some stuff from Inter to fix it: ok
> * you had to add Roboto font to support non latin glyphs due to missing fallbacks. <- new issue caused by stripping Inter files: ko
> 
> Then, what if we place Noto Emoji above Inter in the pub file so web tries to load emojis first and fall back for all chars (latin, tamil, or chinese, etc.) to Inter. That way no need to change Inter ttf files or to add roboto files. No?
> 
> And regarding that issue for animated emojis and using Noto Animated Emojis. If that font is added above Noto Emojis:
> 
> 1. Noto Animated Emojis
> 2. Noto Emojis
> 3. Inter
> 
> Would allow us to use animation only on bubbles where we only have emojis and use non animated ones on others.
> 
> Or is there something I am missing?

Thank for your response, btw flutter's font fallback only triggers when the primary font has no glyph at all since Inter has monochrome emoji glyphs, it always wins regardless of font order. Patching Inter (stripping those glyphs) is the workaround we went with: it fixes the issue while keeping Inter so letterSpacing metrics stay intact.

nqhhdev:
> And otherwise if Inter does not support certain charset, we have to inform our design/product team so we choose another font to avoid having fallbacks that are not compatible with our design system letter spacing and font height size

Yes, of course, I agreed with you @pm-McFly, worth aligning with design team on a base font that doesn't conflict with emoji codepoints so no patching is needed long-term

nqhhdev:
> > Somehting I don't understand is:
> > 
> > * you said the issue is coming from Inter having glyphs for emojis like the heart and thus flutter is using those ones instead of falling back on another font like for the turtle emoji (for example)
> > * you then removed some stuff from Inter to fix it: ok
> > * you had to add Roboto font to support non latin glyphs due to missing fallbacks. <- new issue caused by stripping Inter files: ko
> > 
> > Then, what if we place Noto Emoji above Inter in the pub file so web tries to load emojis first and fall back for all chars (latin, tamil, or chinese, etc.) to Inter. That way no need to change Inter ttf files or to add roboto files. No?
> > And regarding that issue for animated emojis and using Noto Animated Emojis. If that font is added above Noto Emojis:
> > 
> > 1. Noto Animated Emojis
> > 2. Noto Emojis
> > 3. Inter
> > 
> > Would allow us to use animation only on bubbles where we only have emojis and use non animated ones on others.
> > Or is there something I am missing?
> 
> Thank for your response, btw flutter's font fallback only triggers when the primary font has no glyph at all since Inter has monochrome emoji glyphs, it always wins regardless of font order. Patching Inter (stripping those glyphs) is the workaround we went with: it fixes the issue while keeping Inter so letterSpacing metrics stay intact.

WDYT guys: @pm-McFly @9clg6 @jbrnzl

jbrnzl:
> > > Somehting I don't understand is:
> > > 
> > > * you said the issue is coming from Inter having glyphs for emojis like the heart and thus flutter is using those ones instead of falling back on another font like for the turtle emoji (for example)
> > > * you then removed some stuff from Inter to fix it: ok
> > > * you had to add Roboto font to support non latin glyphs due to missing fallbacks. <- new issue caused by stripping Inter files: ko
> > > 
> > > Then, what if we place Noto Emoji above Inter in the pub file so web tries to load emojis first and fall back for all chars (latin, tamil, or chinese, etc.) to Inter. That way no need to change Inter ttf files or to add roboto files. No?
> > > And regarding that issue for animated emojis and using Noto Animated Emojis. If that font is added above Noto Emojis:
> > > 
> > > 1. Noto Animated Emojis
> > > 2. Noto Emojis
> > > 3. Inter
> > > 
> > > Would allow us to use animation only on bubbles where we only have emojis and use non animated ones on others.
> > > Or is there something I am missing?
> > 
> > 
> > Thank for your response, btw flutter's font fallback only triggers when the primary font has no glyph at all since Inter has monochrome emoji glyphs, it always wins regardless of font order. Patching Inter (stripping those glyphs) is the workaround we went with: it fixes the issue while keeping Inter so letterSpacing metrics stay intact.
> 
> WDYT guys: @pm-McFly @9clg6 @jbrnzl

I think it is the best way to do it indeed. Better to keep Inter as primary (for semantic reasons), and then cut the glyph we don't want on it and rely on an other font as a fallback . (Maybe in the future we can have a full Twake font 😉 with everything we want already included)

Btw regarding Noto Animated Emoji from what I've read it does not provides a classic .ttf but animations instead. So it will not be as easy as a font fallback cascade like this anyway:
1. Noto Animated Emojis
2. Noto Emojis
3. Inter

Two questions/concerns:
- I do not understand why do we need all these changes, not only patching Inter font files
- If we patch Inter font files I think we need to add a script to reproduce the stripping.

pm-McFly:
> Two questions/concerns:
> - I do not understand why do we need all these changes, not only patching Inter font files
> - If we patch Inter font files I think we need to add a script to reproduce the stripping.

Valid points! For both

Also, when comparing to TMail I saw it also features inter and almost all roboto+noto fonts..
Maybe we can get two birds in one shot by getting rid of Inter for all
If the font does not match our expectations, maybe we are better to go with another 🤔

pm-McFly:
Because we are changing inter fonts, we must comply to its license.
here are the todos i have identified:

# **1. Modify and Rename the Font Asset**

* [x] **Strip the problematic glyphs** from the original Inter font files.
* [x] **Rename the physical font files** to something custom (e.g., `Twake-Font.ttf`).
* **Why:** "Inter" is a Reserved Font Name (RFN) under the SIL OFL. Stripping glyphs creates a "Derivative Work," meaning we are legally prohibited from continuing to use the original name.

* [x] **Update internal font metadata** (Family Name, PostScript name, etc.) using a font editor to remove all references to "Inter".
* **Why:** The system and end-users cannot see the font presented as "Inter" anywhere.

# **2. Ensure License Compliance (OFL)**

* [x] **Retain the SIL OFL v1.1 license** strictly for the new font files.
* **Why:** The OFL is a "copyleft" license. Even though our app is AGPL-v3, the font itself cannot be relicensed.

* [x] **Include the original copyright notice and OFL text** directly alongside the newly named font files in our repository tree.
* **Why:** This is a strict requirement for distributing modified OFL fonts.

# **3. Update Repository Documentation**

* [x] **Update `README.md` and/or `CREDITS.txt**` to explicitly state the dual-licensing nature of the repository.
* **Why:** To prevent license confusion. We must clarify that the application code is governed by **AGPL-v3**, while the bundled custom font component is governed by **SIL OFL v1.1** (which is officially supported by the Free Software Foundation).

jbrnzl:
@nqhhdev Once created, can we add to this PR the script to generate Twake-Font.ttf ? May be usefull

nqhhdev:
> @nqhhdev Once created, can we add to this PR the script to generate Twake-Font.ttf ? May be usefull

updated

scripts/generate-twake-inter-fonts.sh
```diff
@@ -0,0 +1,75 @@
...
+import sys, re
+from fontTools.ttLib import TTFont
+
+src, dst = sys.argv[1], sys.argv[2]
+font = TTFont(src)
+
+for record in font["name"].names:
+    value = record.toUnicode()
+    # Replace "InterVariable" first, then bare "Inter"
+    updated = value.replace("InterVariable", "TwakeInter")
```
pm-McFly, line 76: Replacing both `Inter` and `InterVariable` to `TwakeInter` is not creating naming conflict when loading the files?

Variable fonts have a different behavior reagrding weights (bold, thin, light, etc.) so I wonder how the app is able to correctly call the according font type.

README.md
```diff
@@ -327,6 +327,8 @@ docker run -d -p 6868:6868 -e TWAKECHAT_LISTEN_PORT=6868 --name twake-web -v /pa
 
 - <a href="https://github.com/googlefonts/noto-emoji/">Noto Emoji Font</a> for the awesome emojis.
 
+- **TwakeInter font** (bundled in `assets/google_fonts/`): a derivative of the [Inter](https://github.com/rsms/inter) typeface by Rasmus Andersson, modified to remove emoji glyphs that interfered with color emoji rendering on Flutter web. "Inter" is a Reserved Font Name under the SIL Open Font License. This derivative has been renamed to "TwakeInter" in compliance with that requirement. The font component is licensed under **SIL OFL v1.1** (see `assets/google_fonts/OFL.txt`), independent of the application's AGPL-v3 license.
```
pm-McFly: perfect, thx!
But license is `assets/google_fonts/LICENSE.txt` not `OFL` ;)

lib/config/themes.dart
```diff
@@ -22,74 +23,22 @@ abstract class TwakeThemes {
...
+    labelSmall: TextStyle(fontWeight: FontWeight.w500, letterSpacing: 0.5),
+    displayLarge: TextStyle(fontWeight: FontWeight.w700),
+    displayMedium: TextStyle(fontWeight: FontWeight.w600, letterSpacing: 0.4),
+    displaySmall: TextStyle(fontWeight: FontWeight.w600, letterSpacing: 0.4),
+    headlineMedium: TextStyle(fontWeight: FontWeight.w600),
+    headlineSmall: TextStyle(fontWeight: FontWeight.w600, letterSpacing: 0.4),
+    headlineLarge: TextStyle(fontWeight: FontWeight.w600, fontSize: 32),
+    titleLarge: TextStyle(fontWeight: FontWeight.w600, letterSpacing: -0.15),
+    titleMedium: TextStyle(fontWeight: FontWeight.w500, letterSpacing: 0.15),
+    titleSmall: TextStyle(fontWeight: FontWeight.w500, letterSpacing: 0.1),
```
pm-McFly, line 41: Why did the letter spacing changed?
Example `bodyLarge` went from `-0.15` to `0.1`

pm-McFly (commented):
Nice catch on that font handling and glyph stripping! thx a lot

README.md
nqhhdev: Updated

lib/config/themes.dart
nqhhdev: - bodyLarge is letterSpacing: -0.15 ← unchanged from main                                                                                                                                                                                                                                                                                             
- titleSmall is letterSpacing: 0.1 ← also unchanged

scripts/generate-twake-inter-fonts.sh
nqhhdev: The script renames font metadata but outputs only one file (TwakeInter.ttf from InterVariable.ttf). The 4 static TTF files (TwakeInter-Regular/Medium/SemiBold/Bold.ttf) were separately added and are referenced by weight in pubspec.yaml. The variable font is mapped as the default (weight 0). Worth clarifying in a reply. Wdyt

lib/config/themes.dart
pm-McFly: my bad! misread the diff!

scripts/generate-twake-inter-fonts.sh
```diff
@@ -0,0 +1,75 @@
...
+    updated = re.sub(r'\bInter\b', 'TwakeInter', updated)
+    if updated != value:
+        record.string = updated.encode(
+            "utf-16-be" if record.isUnicode() else "latin-1",
+            errors="replace",
+        )
+
+font.save(dst)
```
pm-McFly, line 84: Is this part of the script stripping the glyphs we don't want? like the Inter glyph for "❤️"

scripts/generate-twake-inter-fonts.sh
pm-McFly: No need thx! It was a curiosity question ;)

scripts/generate-twake-inter-fonts.sh
nqhhdev: No, this script only renames the font metadata (family name, PostScript name, etc.) from "Inter" to "TwakeInter". The emoji glyph stripping was done separately on the bundled TTF files before committing. The script handles the rename step only.  Do you want to add it into the script?

pm-McFly: Yes, as the script downloads the font version we want, that will be our update door ;)
having an all in one script would be easier I think

nqhhdev: Added, help me review again pls

pm-McFly (approved):

scripts/generate-twake-inter-fonts.sh
pm-McFly: LGTM

Crash--:
Do we need to push this font to the flutter design system? Do we have the same issue on Twake-Mail?

pm-McFly:
> Do we need to push this font to the flutter design system? Do we have the same issue on Twake-Mail?

Now that we have identified a way to fix, I have shared to @hoangdat this PR, he'll look into it with his team ;)

My take is to place the font defintion into the DS at some point yes.
I feel like it is closely bound to Typography anyway.

How is it handled in the React Apps?

jbrnzl:
> Do we need to push this font to the flutter design system? Do we have the same issue on Twake-Mail?

I think it can be usefull indeed, it looks like Twake Mail is suffering the same issue on some places:
image

@nqhhdev wdyt ? If you need help or want to look at it with me tomorrow, please go, I'll be happy to do so

Crash--:
If the fix needs to live in both apps, then it should not be done here, but on the Design system lib. 

Also few words (cf quoted text), can we use emoji-data.txt instead of hardcoding our list? (https://unicode.org/Public/emoji/1.0/emoji-data.txt) 


```
 One concern about the glyph-stripping step that I think will bite us later. The script removes glyphs only from two hand-picked ranges:

  EMOJI_RANGES = [
      (0x2194, 0x2BFF),
      (0x1F000, 0x1FBFF),
  ]

  The problem is that emoji codepoints are not confined to those two ranges — Unicode scatters them across several blocks. So any emoji-capable character that (a) Inter
  has a glyph for and (b) falls outside these ranges will still render as a monochrome Inter glyph — i.e. the exact same "black heart" bug, just for a different
  character.

  I checked the actual cmap of Inter-Regular.ttf against known emoji codepoints, and here are the ones that survive the current patch:

  ┌──────────────────┬───────────┬──────────────────────────┐
  │ Char (with VS16) │ Codepoint │  In Inter, not stripped  │
  ├──────────────────┼───────────┼──────────────────────────┤
  │ ©️               │ U+00A9    │ yes → renders monochrome │
  ├──────────────────┼───────────┼──────────────────────────┤
  │ ®️               │ U+00AE    │ yes → renders monochrome │
  ├──────────────────┼───────────┼──────────────────────────┤
  │ ‼️               │ U+203C    │ yes → renders monochrome │
  ├──────────────────┼───────────┼──────────────────────────┤
  │ ⁉️               │ U+2049    │ yes → renders monochrome │
  ├──────────────────┼───────────┼──────────────────────────┤
  │ ™️               │ U+2122    │ yes → renders monochrome │
  └──────────────────┴───────────┴──────────────────────────┘

  Plus the keycap bases (#, *, 0–9), which Inter also covers — but those we must not strip, since they need to stay as normal text glyphs (they only become emoji when
  followed by the combining keycap U+20E3).

  These are low-frequency, so it's not urgent. But it's the same failure mode, so we'll likely get follow-up "X shows up black" reports unless we close the whole class
  now.

  Suggestion: drive the strip list from Unicode data instead of hand-picked ranges

  Unicode publishes an authoritative file called emoji-data.txt that lists every codepoint with the Emoji property — i.e. the canonical "is this an emoji" list, updated
  with each Unicode release. Instead of guessing ranges, we can compute:

  ▎ strip set = (Inter's cmap) ∩ (codepoints with Emoji=Yes) − keep-list (0–9, #, *)
```

Crash--:
> How is it handled in the React Apps?

On a React/DOM app, Inter as the UI font doesn't cause this bug because the browser handles it. Browsers do per-character font matching and honor emoji presentation: when a character carries the VS16 selector (❤️, which is what native keyboards send), the browser ignores Inter for that character and falls back to the OS color emoji font (Apple Color Emoji / Segoe UI Emoji / Noto Color Emoji), even though Inter has a glyph for it. Bare ❤ without VS16 stays monochrome, but that's the Unicode-correct text-presentation default and matches native apps anyway.

  That browser fallback is exactly what Flutter's CanvasKit renderer doesn't reproduce: CanvasKit paints glyphs itself and only knows the fonts we register, so the bundled Inter "wins" the codepoint and we get the black heart. That's why this font-patching is a CanvasKit-specific workaround a React app (or Flutter's old HTML renderer) never needs it.

Crash--:
And I just tested in on Twake Drive 

Capture d’écran 2026-06-17 à 17 12 35

But 
Capture d’écran 2026-06-17 à 17 12 57


Because here, we're forcing the color so we erase the red color. I'll ask @Bobpodvalnyi what to do in this case. 

Edit: In fact this is not because of forcing a color, it's because we have a component that split filename and it seems that we're removing some characters...

pm-McFly:
FYI, this PR with in place fix is being reworked with a design system integration of the fonts and typography declaration:
- https://github.com/linagora/linagora-design-flutter/issues/78
- https://github.com/linagora/linagora-design-flutter/pull/77
- https://github.com/linagora/twake-on-matrix/pull/3160

Keeping as backup just in case

jbrnzl:
I close this PR as some new are fixing the issue https://github.com/linagora/twake-on-matrix/pull/3160 https://github.com/linagora/linagora-design-flutter/pull/77 https://github.com/linagora/twake-on-matrix/pull/3172

PR #3142, opened 2026-06-15
fix(web): guard ChatWebScrollbar against multi-attached ScrollController

9clg6:
## Problem
Entering message multi-select mode rebuilds the web chat layout, briefly attaching the same `ScrollController` to two `Scrollable`s. Reading `ScrollController.position` then trips the framework assertion `_positions.length == 1`.

## Fix
All position access goes through a single `_attachedPosition` getter that returns `null` unless exactly one scroll view is attached. `null` is treated as "no scrollable content" (zero-height thumb, no-op scroll).

## Context
Surfaced by the web Patrol migration (#3128), which only swallowed the assertion in tests. This fixes the root cause.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat scrollbar stability by making its thumb and metrics only activate when a single scrollable area is reliably attached.
  * Prevented potential framework assertion crashes during transient multi-scroll attachment scenarios.
  * Updated thumb dragging, track tapping, and wheel/trackpad scrolling to ignore input when scrolling isn’t available.
  * Refined scrollbar content detection to avoid incorrect thumb sizing or display.

* **Tests**
  * Added widget tests covering thumb visibility when no scrollable is attached, multi-attachment safety, and correct thumb rendering for a single attached scrollable.

### DIFF
--- a/lib/pages/chat/chat_web_scrollbar.dart
+++ b/lib/pages/chat/chat_web_scrollbar.dart
@@ -50,16 +50,31 @@ class _ChatWebScrollbarState extends State<ChatWebScrollbar> {
     if (mounted) setState(() {});
   }
 
+  /// The single attached [ScrollPosition], or `null` when the controller has
+  /// zero or more than one attached scroll view.
+  ///
+  /// During web layout rebuilds (e.g. entering message multi-select mode) the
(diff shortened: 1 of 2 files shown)


nqhhdev (approved):
LGTM

jbrnzl (approved):

PR #3151, opened 2026-06-16
TW-3073(fix): preload hero users so display names show on cold start

### LINKED ISSUE #3073: [BUG] Display Name not displayed correctly for a while (until it's loaded)

### Describe the bug

Sometimes the displayname isn't fetched when you load the app, you see the username instead

https://github.com/user-attachments/assets/5ba96191-7016-41fd-a1d8-3402754280fc

### Pre-Bug State

N/A

### To Reproduce

1. Start TC app 
2. Go to the chat list 
3. Some display names aren't fetched, usernames are displayed instead. If you wait for a while, they are finally fetched

### Expected results

The Display Name/the chosen contact name is displayed immediately after starting the app

### Context

- Device: Redmi Note 12 Pro
- OS: Android, iOS
- App Version: 2.22.2
- Server: linagora
- Account information:
- Others

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
## Ticket
Fixes #3073

## Root cause
On a cold start, on DMs or unnamed groups the hero member profiles aren't loaded into memory yet and `getLocalizedDisplayname` falls back to a placeholder derived from the matrix ID.

As getLocalizedDisplayname from matrix-dart-sdk documentation says:

> Please note, that necessary room members are lazy loaded. To be sure that you have the room members, call and await `Room.loadHeroUsers()` before.

The bug in action: 

https://github.com/user-attachments/assets/db8987ca-3aa3-4529-a718-ead0414b94cd

## Solution
`ChatListItem` becomes a `StatefulWidget` that loads the room's hero users once (only when the room has no name).

## Test recommendations
Cold start with DMs or unnamed groups: names show directly instead of placeholders

## Demos
No specific demo added, on purpose, the displayName are now correctly fetched from cold start.

## Summary by CodeRabbit

* **Refactor**
  * Updated the chat list item rendering to load hero-user info only when the room has an empty name, and to compute the displayed room name within the updated UI flow for improved consistency.
  * Prevents unnecessary hero-user reloads while keeping the list accurate as room context changes.

* **Tests**
  * Added widget tests verifying hero-user loading runs for nameless rooms and is skipped for named rooms.

### DIFF
--- a/lib/pages/chat_list/chat_list_item.dart
+++ b/lib/pages/chat_list/chat_list_item.dart
@@ -17,7 +17,7 @@ import 'package:matrix/matrix.dart';
 
 enum ArchivedRoomAction { delete, rejoin }
 
-class ChatListItem extends StatefulWidget {
+class ChatListItem extends StatelessWidget with ChatListItemMixin {
   final Room room;
   final bool activeChat;
   final bool isSelectedItem;
@@ -43,37 +43,8 @@ class ChatListItem extends StatefulWidget {
     super.key,
   });
 
-  @override
-  State<ChatListItem> createState() => _ChatListItemState();
-}
-
-class _ChatListItemState extends State<ChatListItem> with ChatListItemMixin {
-  Future<List<User>>? _heroUsersFuture;
-
-  Room get room => widget.room;
-
-  bool get _isGroupChat => !room.isDirectChat;
-
-  @override
-  void initState() {
-    super.initState();
-    _loadHeroUsersIfNeeded();
-  }
-
-  @override
-  void didUpdateWidget(covariant ChatListItem oldWidget) {
-    super.didUpdateWidget(oldWidget);
-    if (oldWidget.room.id != room.id) {
-      _loadHeroUsersIfNeeded();
-    }
-  }
-
-  void _loadHeroUsersIfNeeded() {
-    _heroUsersFuture = room.name.isEmpty ? room.loadHeroUsers() : null;
-  }
-
   void clickAction(BuildContext context) async {
-    if (widget.onTap != null) return widget.onTap!();
+    if (onTap != null) return onTap!();
     switch (room.membership) {
       case Membership.ban:
         TwakeSnackBar.show(
@@ -115,18 +86,20 @@ class _ChatListItemState extends State<ChatListItem> with ChatListItemMixin {
     }
   }
 
+  bool get _isGroupChat => !room.isDirectChat;
+
   @override
   Widget build(BuildContext context) {
     return Padding(
       padding: ChatListItemStyle.padding,
       child: TwakeInkWell(
-        isSelected: widget.activeChat,
+        isSelected: activeChat,
         onTap: () => clickAction(context),
-        onSecondaryTapDown: widget.onSecondaryTapDown,
-        onLongPress: widget.onLongPress,
+        onSecondaryTapDown: onSecondaryTapDown,
+        onLongPress: onLongPress,
         child: TwakeListItem(
-          child: FutureBuilder<List<User>>(
-            future: _heroUsersFuture,
+          child: FutureBuilder(
+            future: room.name.isEmpty ? room.loadHeroUsers() : null,
             builder: (context, _) {
               final displayName = room.getLocalizedDisplayname(
                 MatrixLocals(L10n.of(context)!),
@@ -137,8 +110,7 @@ class _ChatListItemState extends State<ChatListItem> with ChatListItemMixin {
                 child: Row(
                   mainAxisAlignment: MainAxisAlignment.center,
                   children: [
-                    if (widget.isEnableSelectMode)
-                      widget.checkBoxWidget ?? const SizedBox(),
+                    if (isEnableSelectMode) checkBoxWidget ?? const SizedBox(),
                     Padding(
                       padding: ChatListItemStyle.paddingAvatar,
                       child: Stack(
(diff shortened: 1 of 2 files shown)


lib/pages/chat_list/chat_list_item.dart
```diff
@@ -43,8 +43,27 @@ class ChatListItem extends StatelessWidget with ChatListItemMixin {
...
+
+  bool get _isGroupChat => !room.isDirectChat;
+
+  @override
+  void initState() {
+    super.initState();
+    if (room.name.isEmpty) {
+      _heroUsersFuture = room.loadHeroUsers();
+    }
+  }
```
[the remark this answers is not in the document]
jbrnzl: Addressed

pm-McFly:
Out of nowhere... Do you know or understand why the function is named "Localized" names?

jbrnzl:
> Out of nowhere... Do you know or understand why the function is named "Localized" names?

Sure, I had the same question before I checked, that seems like a weird name to me to. In fact it's called "Localized..." because you can give the method a MatrixLocalizations param so that it localizes the fallbacks like "Group with X, Y..." or "Empty chat" when a room has no explicit name :)

lib/pages/chat_list/chat_list_item.dart
```diff
@@ -43,8 +43,37 @@ class ChatListItem extends StatelessWidget with ChatListItemMixin {
...
+  void initState() {
+    super.initState();
+    _loadHeroUsersIfNeeded();
+  }
+
+  @override
+  void didUpdateWidget(covariant ChatListItem oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.room.id != room.id) {
+      _loadHeroUsersIfNeeded();
```
krille-chan: I think you can actually make this much more easy. `Room.loadHeroUsers()` just does nothing severe if the heroes are already loaded and it is built in mind with being used in a FutureBuilder (where it gets reexecuted on every build). Same in the FluffyChat Chat List. You might wanna safe some performance here but making a StatelessWidget Stateful also costs performance.

However you can skip it if the room.name is empty:

https://github.com/krille-chan/fluffychat/blob/main/lib/pages/chat_list/chat_list_item.dart#L74-L76

jbrnzl: Oh ! Thanks a lot for the review ! You are right that's much cleaner. I overthought it trying to save costs...I updated !

nqhhdev (approved):

lib/pages/chat_list/chat_list_item.dart
```diff
@@ -101,67 +98,78 @@ class ChatListItem extends StatelessWidget with ChatListItemMixin {
...
-                            ),
-                            child: Icon(
-                              Icons.group,
-                              size: ChatListItemStyle.groupIconSize,
-                              color: room.isUnreadOrInvited
-                                  ? LinagoraSysColors.material()
-                                        .onSurfaceVariant
-                                  : LinagoraRefColors.material().tertiary[30],
-                            ),
+          child: FutureBuilder(
```
9clg6, line 101: I'm not a big fan of FutureBuilder, it has a real impact on performances. For example, FutureBuilder re-load at every build, we probably don't want it. 
Instead of 
```dart
future: room.name.isEmpty ? room.loadHeroUsers() : null,
```

could we assign the future outside the build,  as

```dart 
late Future<int> _futureData;

@override
void initState() {
  super.initState();
  _futureData = fetchData();
}

return FutureBuilder<int>(
    future: _futureData,
```

lib/pages/chat_list/chat_list_item.dart
```diff
@@ -101,67 +98,78 @@ class ChatListItem extends StatelessWidget with ChatListItemMixin {
...
+            future: room.name.isEmpty ? room.loadHeroUsers() : null,
+            builder: (context, _) {
```
9clg6, line 103: Please handle connection state as : 

```dart
builder: (context, snapshot) {
  if (snapshot.connectionState == ConnectionState.waiting) {
    return CircularProgressIndicator();
  } else if (snapshot.connectionState == ConnectionState.done) {
    if (snapshot.hasError) {
      return Text('Error: ${snapshot.error}');
    } else {
      return Text('Data: ${snapshot.data}');
    }
  }
  // Missing cases for ConnectionState.none and ConnectionState.active
}
```

Please prefer skeleton loader than a circular progress indicator

lib/pages/chat_list/chat_list_item.dart
9clg6: WDYT ?

lib/pages/chat_list/chat_list_item.dart
9clg6: WDYT ?

9clg6 (changes requested):

lib/pages/chat_list/chat_list_item.dart
jbrnzl: Thanks for the review ! I think we can follow FluffyChat upstream on this one as @krille-chan [already recommended ](https://github.com/linagora/twake-on-matrix/pull/3151#discussion_r3426101091)
With the room.name guard + loadHeroUsers cache I think it's more than an acceptable trade-off.

lib/pages/chat_list/chat_list_item.dart
jbrnzl: @9clg6 My concern is that we already have a valid fallback state so I'm not sure it is really usefull here.
Also if you look carefully at the future, it's not throwing errors: ignoreErrors is true. So it is shaped in a way that is good too as is on a FutureBuilder as FluffyChat upstream. This is my opinion, but I can be wrong, please tell me what you think about it

9clg6 (approved):

Issue #3130, opened 2026-06-12
Create the `TextSearch` Core Module


pm-McFly:
# Description

Introduce a central `TextSearch` module (cf. #3129) that replaces the scattered `.toLowerCase().contains()` search logic across the codebase. The module must be built around a **configurable pipeline** so that future normalization steps (diacritics, transliteration) and match strategies (multi-word, fuzzy) can be added without changing the public API.

# Pipeline architecture
 
```
needle -> normalize once -> normNeedle
 
for each item in haystack (single pass):
  for each fieldExtractor:
    extract field from item -> normalize at compare time -> compare to normNeedle
  if any field matches -> keep the ORIGINAL item in results
```
 
The pipeline has two independent axes:
 
1. **Normalization steps** - applied to the needle exactly once up front, and to each haystack element's extracted field **at compare time**, inside the single pass over `haystack`. Normalization never mutates the original item and is never written back into a separate normalized haystack; on a match, the **original, untouched item** (same type as the input list - `List<T> in -> List<T> out`) goes into the result. A normalization step must be a **general-purpose text transform** - something that makes sense to apply to *any* string, not just one domain's data:
   - Case folding (`toLowerCase`)
   - Diacritic stripping (NFD decomposition - see Issue #3131)
   - Future: transliteration (see https://github.com/linagora/twake-on-matrix/discussions/3138)
2. **Match strategies** (how the normalized needle is compared against each normalized haystack):
   - `exact` - equality
   - `substring` - `contains()` (current default)
   - Future: `multiWord`, `fuzzy` (see https://github.com/linagora/twake-on-matrix/discussions/3138 - **not implemented now**)

## Scope boundary: pipeline normalization vs. field extraction
 
Domain-specific, field-shape transforms - stripping spaces from a phone number, taking the local part of a Matrix ID (`split(':')[0]`), slugifying a display name - are **not** general text transforms. They only make sense for one kind of field.

These stay **caller-side**, expressed as the `fieldExtractors` closures that are already part of the API:
 
```dart
fieldExtractors: [(contact) => contact.phoneNumber.replaceAll(' ', '')]
```
 
This is not a "special case" bolted on - `fieldExtractors` are the designed extension point for "turn this object into the string I want to match against." No second mechanism needed, no pipeline bypass.
 
[...]

Issue #3129, opened 2026-06-12
Search System Improvements


pm-McFly:
# Background & Objective

Currently, list-filtering and text search logic across the Flutter/Dart application rely on brittle, scattered `.toLowerCase().contains()` calls. This epic introduces a central, unified `TextSearch` module (`lib/domain/search/text_search.dart`).

The immediate goal is to establish exact behavioral parity across the app, ensuring no UX regressions, while laying the architectural groundwork for advanced search capabilities like fuzzy matching, multi-word search, and indexing in the future.

# Core Architecture

The search pipeline is built in **Dart** and operates on two independent axes:

1. **Normalization steps:** General-purpose text transforms applied to both the query (needle) and target data (haystack) before matching (e.g., case folding, NFD diacritic stripping). 
2. **Match strategies:** How the normalized needle compares against the normalized haystack (e.g., `exact`, `substring`).

*Note: Domain-specific shape transforms - such as stripping spaces from phone numbers or isolating the local part of a Matrix ID - must be strictly handled via `extractor` rather than pipeline included normalizations.*

## Technical Approach (Single-Pass Loop)

For performance and simplicity, the normalization of the haystack elements happens strictly at **comparison time**.
The pipeline processes the haystack in a single loop (`O(n)`).

If a normalized field extracted from a haystack item matches the normalized input, the **raw item** is added to the results. 

The return type must precisely match the type of the haystack.

## Reference implementation shape
 
Single pass over `haystack`. Normalization runs at compare time on the extracted field, never on the item itself. Return type is `List<T>` - same as the input - containing the original, untouched items.
 
```dart
abstract class NormalizationStep {
  String normalize(String input);
}

// This enum can be extended in the future when adding Fuzzy matching for example
enum SearchMode { exact, substring }

@immutable
class SearchOptions {
  final bool caseSensitive;      // default: false (current behavior)
  final bool diacriticSensitive; // default: false  (changed behavior)
  final SearchMode mode;         // default: SearchMode.substring
  final List<NormalizationStep>? normalize; // optional override list

  const SearchOptions({
    this.caseSensitive = false,
    this.diacriticSensitive = false,
    this.mode = SearchMode.substring,
    this.normalize,
  });
}

List<T> matchAnyField<T>(
  String needle,
  List<T> haystack, {
  required List<String? Function(T)> fieldExtractors,
  SearchOptions options = const SearchOptions(),
}) {
  if (needle.isEmpty) return <T>[];
 
[...]

PR #3156, opened 2026-06-17
TW-3130: Create the SearchEngine Core Module (Part 1)

aparesy:
**Related issue**

- #3130

## Impact description                 

  This PR lays the foundation for a unified, modular client-side search feature. It introduces a `SearchEngine` class under `lib/domain/services/search/`, built around a composable normalization pipeline. Each normalization    
  step is an independent, pluggable unit (implementing `NormalizationStep`), making it straightforward to extend later (e.g. adding diacritic stripping or fuzzy matching) without touching the core engine.                       

  The engine exposes two methods:           
- `matchAnyField<T>` — filters a list by matching a needle against any of the provided field extractors
- `matchesText` — convenience wrapper for single-string matching

  Search behavior is controlled via `SearchOptions`, which currently supports `caseSensitive` and `mode` (`exact` or `substring`). 

## Test recommendations

```
  flutter test test/domain/services/search/          
```

- `match_any_field_test.dart` — covers the core matching contract: substring match, empty needle, no match, original items returned, case-insensitivity, `caseSensitive` flag, multiple field extractors
- `match_text_latin_test.dart` — covers `matchesText` default behavior: found/not found, empty needle, case-insensitive, diacritics preserved by default

Example:                                                                                                                                                                                                                         
```dart                                                                                                                                                                                                                         
test('should match on any of the provided field extractors', () {                                                                                                                                                                
final items = [
{'name': 'Alice', 'email': 'alice@example.com'},                                       
{'name': 'Bob', 'email': 'bob@example.com'},                                                 
];
final result = engine.matchAnyField( 
 'alice',                                                                                                                                                                                                                     
      items,                                                                                                                                                                                                                       
      fieldExtractors: [                                                                                                                                                                                                           
        (m) => m['name'],                                                                                                                                                                                                          
        (m) => m['email'],                                                                                                                                                                                                         
      ],                                                                                                                                                                                                                           
    );                                                                                                                                                                                                                             
    expect(result, [items[0]]);                                                                                                                                                                                                    
  });                                                                                                                                                                                                                              
```

## Resolved
All 12 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added a reusable `SearchEngine` with multi-field searching and single-text matching.
  * Introduced `SearchOptions` to configure case sensitivity and match mode (substring vs exact).
  * Added text normalization steps to support consistent matching behavior.
  * Registered `SearchEngine` for global dependency access.
* **Tests**
  * Added unit/widget tests covering empty queries, exact vs substring matching, case sensitivity behavior, multi-field extractors, and Latin/diacritic handling.


nqhhdev:
I have 2 concerns:

1. Layer placement — lib/domain/ in our project is strictly for business use cases (interactors, repositories, models). `SearchEngine` is a generic, stateless text utility with no Matrix/Twake context. Should this live in lib/utils/ instead? Introducing lib/domain/services/ as a new sub-layer also creates structural drift from our current domain structure. 

2.  DI registration — Our convention is to register all components in GetIt. SearchEngine is currently used via const `SearchEngine()` directly at call sites, bypassing DI. Should it be registered as a singleton, or is it intentionally a plain utility (which would support the lib/utils/ argument)?

aparesy:
Regarding layer placement, this has been discussed in #3129 and this really depends on one's philosophy about architecture so I'm not sure I'm at a good position to impose anything. To be honest, both visions seem valid.

About DI registration, I'd say that the search feature tends to be more on the side of a local utility

jbrnzl:
> I have 2 concerns:
> 
> 1. Layer placement — lib/domain/ in our project is strictly for business use cases (interactors, repositories, models). `SearchEngine` is a generic, stateless text utility with no Matrix/Twake context. Should this live in lib/utils/ instead? Introducing lib/domain/services/ as a new sub-layer also creates structural drift from our current domain structure.
> 2. DI registration — Our convention is to register all components in GetIt. SearchEngine is currently used via const `SearchEngine()` directly at call sites, bypassing DI. Should it be registered as a singleton, or is it intentionally a plain utility (which would support the lib/utils/ argument)?

I agree with you on the fact that it should be placed under lib/utils/

jbrnzl (approved):
Well done!

aparesy:
To be honest an argument could also be made in favor of domain/model or domain/services because this could be seen as "business logic" from what I understand
The closest thing to this search core in domain/utils would be domain/logging and domain/manager that both implement some kind of logic with multiple subfolders (The other directories under domain/utils do not have any subfolders and are quite synthetic), but probably both of these could fit under domain/model or domain/services

aparesy:
Maybe domain/model/search would be a compromise since it doesn't make us create a new folder entirely in domain/ but at the same time it allows us to keep a precise denomination @nqhhdev @pm-McFly ?

pm-McFly:
I agree with @nqhhdev on the DI statement. Search should be a singleton used/passed by any components needing it.

I am not so fond of `utils` kind of folders usually...
But it is already in the archecture so I agree with both @nqhhdev and @jbrnzl that maybe placing the files under `lib/utils/...` is better for this service implementation.

Thx for explaining the folder structure! :)
Is it documented somewhere, btw?

(My bad for misleading suggestion @aparesy)

aparesy:
Alright I'll do a fixup

aparesy:
Looks good now @nqhhdev @jbrnzl :)

jbrnzl (approved):

nqhhdev (approved):
LGTM

aparesy:
Actually should we put it in GetIt now instead ?

aparesy:
Last fixup: added two lines in get_it_initializer to include getIt

jbrnzl (approved):

nqhhdev (approved):

Issue #3131, opened 2026-06-12
Add Diacritic / NFD Normalization in Search


pm-McFly:
# Description

Build on the `TextSearch` core (#3130) by adding a pluggable diacritic-stripping normalization step. This allows callers to opt into accent-insensitive matching (e.g. searching "Zoe" matches "Zoé") without changing the default behavior.
 
This is a **general-purpose** text transform (works on any string, any field), so it correctly belongs in the normalization pipeline per Issue 01's scope rule — unlike phone/Matrix-ID shape transforms, which stay caller-side as `fieldExtractors`.
 
# Current state

- A hand-rolled `removeDiacritics()` extension exists in [`lib/utils/string_extension.dart` (lines 15–27)](https://github.com/linagora/twake-on-matrix/blob/350c1923698dfb10d56204973477aa26ebe8e5dc/lib/utils/string_extension.dart#L15).
- It uses a hard-coded latin-character map and is **never called during search** today.
- The new step should use **Unicode NFD decomposition** for broader coverage (and maybe NFKD?).

# Proposed API changes
 
1. **New normalization step:**
```dart
// lib/domain/search/normalization/diacritic_strip_step.dart
class DiacriticStripStep implements NormalizationStep {
  @override
  String normalize(String input) {
    // NFD decomposition + strip combining marks
    // Dart stdlib does not expose ICU; the package `unidecode` or the existing
    // hand-rolled map may be used. Evaluate during implementation.
  }
}
```
 
2. **`SearchOptions` update:**
```dart
class SearchOptions {
  final bool caseSensitive;      // default: false
  final bool diacriticSensitive; // default: true  ← unchanged default
  final SearchMode mode;
  final List<NormalizationStep>? normalize;
 
  const SearchOptions({
    this.caseSensitive = false,
    this.diacriticSensitive = true, // keeps current behavior
    this.mode = SearchMode.substring,
    this.normalize,
  });
}
```
 
When `diacriticSensitive == false` and the caller does **not** provide a custom `normalize` list, the pipeline auto-injects `DiacriticStripStep` immediately before `CaseFoldingStep`.
When `diacriticSensitive == true`, the step is omitted.
 
## Default value decision
 
| Option | Value | Justification |
|--------|-------|---------------|
| `diacriticSensitive` | `true` (default) | No existing call site strips diacritics during search. Switching the default would silently change behavior for users with accented names/messages. |
 
[...]

PR #3157, opened 2026-06-17
TW-3131: Add Diacritic / NFD Normalization in Search (Part 2)

aparesy:
## Ticket
**Related issue**

- #3131

## Impact description

This PR builds on #3130 by adding NFD-based diacritic normalization to the `SearchEngine`. It introduces a `DiacriticStripStep` that decomposes accented characters using Unicode NFD normalization, then strips combining marks (category `\p{Mn}`), allowing searches like `"nguyen"` to match `"Nguyễn"` or `"cafe"` to match `"café"`.

The `diacriticSensitive` flag is added to `SearchOptions` (defaults to `true`, preserving existing behavior). Setting it to `false` activates the diacritic stripping step in the pipeline.

The existing `removeDiacritics()` method in `string_extension.dart` is marked `@Deprecated` as `DiacriticStripStep` is now the canonical way to handle this.

## Test recommendations

14 tests across two files, runnable with:
```
flutter test test/domain/services/search/
```

- `diacritic_strip_latin_test.dart` — covers Latin diacritics: acute (é→e), grave (à→a), umlaut (ü→u), tilde (ñ→n), cedilla (ç→c), circumflex (ê→e)
- `match_text_vietnamese_test.dart` — covers Vietnamese: ă, â, ê, ô, ơ, ư, full names (Nguyễn, Việt Nam, Hà Nội), stacked tone marks (phở)

## Resolved
All 14 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added diacritic-insensitive text searching controlled via `SearchOptions` (accent differences no longer prevent matches).
  * Supports Unicode normalization for diacritic handling (configurable decomposition).
  * Improved matching accuracy for Latin and Vietnamese text, including composed/stacked diacritics.
* **Deprecations**
  * Deprecated the legacy `removeDiacritics()` helper in favor of the diacritic-stripping approach used by search.
* **Tests**
  * Added/expanded automated coverage for diacritic-insensitive matching (positive and negative cases) for both Latin and Vietnamese inputs.

### DIFF
--- a/lib/utils/search/search_options.dart
+++ b/lib/utils/search/search_options.dart
@@ -1,12 +1,20 @@
+import 'steps/diacritic_strip_step.dart';
+
+export 'steps/diacritic_strip_step.dart' show UnicodeDecomposition;
+
 // Can be extended in the future, e.g. with fuzzy matching.
 enum SearchMode { exact, substring }
 
 class SearchOptions {
   final bool caseSensitive;
+  final bool diacriticSensitive;
+  final UnicodeDecomposition diacriticDecomposition;
   final SearchMode mode;
 
   const SearchOptions({
     this.caseSensitive = false,
+    this.diacriticSensitive = true,
+    this.diacriticDecomposition = UnicodeDecomposition.nfd,
     this.mode = SearchMode.substring,
   });
 }
--- a/test/utils/search/match_text_vietnamese_test.dart
+++ b/test/utils/search/match_text_vietnamese_test.dart
@@ -1,0 +1,71 @@
+// Stress-tests diacritic stripping with Vietnamese: stacked marks (tone + vowel modification)
+// that NFD decomposes into multiple combining characters, all stripped to the base letter.
+// Also verifies that stripping never collapses distinct base letters into each other.
+import 'package:fluffychat/utils/search/search_options.dart';
+import 'package:fluffychat/utils/search/search_engine.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+const _engine = SearchEngine();
+const _insensitive = SearchOptions(diacriticSensitive: false);
+
+bool _match(String needle, String haystack) =>
+    _engine.matchesText(needle, haystack, options: _insensitive);
+
+void main() {
+  group('matchesText vietnamese — diacriticSensitive: false', () {
+    test('should match ă and â (a variants)', () {
+      expect(_match('ban', 'băn'), isTrue);
+      expect(_match('ban', 'bân'), isTrue);
+    });
+
+    test('should match ê (e with circumflex)', () {
+      expect(_match('kep', 'kẹp'), isTrue);
+    });
+
+    test('should match ô and ơ (o variants)', () {
+      expect(_match('bo', 'bô'), isTrue);
+      expect(_match('bo', 'bơ'), isTrue);
+    });
+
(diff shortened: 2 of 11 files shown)


aparesy:
In the last fixup, I moved the files to lib/utils as a follow-up of the latest modifications in #3156. I also added some more tests to be more exhaustive and added the possibility to do an NFKD decomposition as an alternative for the existing NFD decomposition.

The issue describes integration tests as a requisite, but since these modifications do not change the concrete behavior of the application, I don't really know how I could add pertinent integration tests

aparesy:
Quick update because of modifications in #3156, nothing changed here

lib/utils/search/search_options.dart
```diff
@@ -1,12 +1,20 @@
...
 
 class SearchOptions {
   final bool caseSensitive;
+  final bool diacriticSensitive;
+  final UnicodeDecomposition diacriticDecomposition;
   final SearchMode mode;
 
   const SearchOptions({
     this.caseSensitive = false,
+    this.diacriticSensitive = true,
```
nqhhdev, line 16: `diacriticSensitive` defaults to true (sensitive), but `caseSensitive` defaults to false (insensitive). Is this asymmetry intentional? Since the whole feature is about enabling diacritic-insensitive search, I'd expect the default to be false so callers don't have to opt in every time. Or is true intentional to avoid breaking existing behavior?

aparesy: Yes the idea is to keep parity with the old behavior by default as mentioned in the original issue #3131 :
class SearchOptions {
  final bool caseSensitive;      // default: false
  final bool diacriticSensitive; // default: true  ← unchanged default
  final SearchMode mode;
  final List<NormalizationStep>? normalize;
 
  const SearchOptions({
    this.caseSensitive = false,
    this.diacriticSensitive = true, // keeps current behavior
    this.mode = SearchMode.substring,
    this.normalize,
  });
}

test/utils/search/match_text_vietnamese_test.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:fluffychat/utils/search/search_engine.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+const _engine = SearchEngine();
+const _insensitive = SearchOptions(diacriticSensitive: false);
+
+bool _match(String needle, String haystack) =>
+    _engine.matchesText(needle, haystack, options: _insensitive);
+
+void main() {
```
jbrnzl, line 14: Just a little question, does it support stripped letters ? For example 'Đ' in "Đạt" ? :)

aparesy: For most letters with a strip or whatever small tweak, the decomposition actually supports them, but AFAIK the stripped D is the exception to the rule as it is considered an entirely different letter. I have tested it and this confirms that they are two different letters ; however, while testing more vietnamese diacritics, it seems that the decomposition doesn't work perfectly yet with them

Screenshot From 2026-06-23 16-16-30

jbrnzl (approved):
LGTM

nqhhdev (approved):
LGMT

aparesy:
I just rebased because some commits were pointing at the wrong branch

jbrnzl (approved):

9clg6 (approved):

Issue #3106, opened 2026-06-04
[BUG] If you don't have TC running in the background, you can't jump to the proper notification


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

If TC isn't running in the background and you jump to a notification, you're redirected to the list of chats instead of the particular chat

https://github.com/user-attachments/assets/7c73ed35-0fb6-4a13-af89-426d15b3bf4a

### Pre-Bug State

N/A

### To Reproduce

1. Kill Twake Chat if you had it running 
2. Receive a notification 
3. Tap on the notification 
4. You're in chat list instead of the chat where the notification was from

### Expected results

When you jump to a notification, you're redirected to that particular message 

### Context

- Device: Redmi Note 12 Pro
- OS: Android
- Browser: [e.g. chrome, safari]
- App Version: 2.22.4
- Server: linagora
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

PR #3167, opened 2026-06-22
TW-3106: Fix cannot jump to the proper notification on android

nqhhdev:
## Ticket
**Related issue**

- #3106

## Root cause

-   Android only. Affects users who fully kill the app then tap a push notification — they land on the chat list instead of the specific room.

## Solution

-  Android only. Affects users who fully kill the app then tap a push notification — they land on the chat list instead of the specific room.

## Impact description

- Fixed Android cold-start notification routing — users who kill the app and tap a push notification now land directly in the correct chat room.

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/66a89808-8528-45ee-911e-306f9066d9a5

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a cold-start race that could cause missed or misrouted push notifications.
  * Improved cold-start notification launch handling with clearer decision logging during notification-based app launches.

* **Chores**
  * Expanded push-related Sentry tracking to cover additional notification and payload parsing failures.
  * Enhanced push setup debug logging to help diagnose notification routing issues.

### DIFF
--- a/lib/utils/push_helper.dart
+++ b/lib/utils/push_helper.dart
@@ -95,7 +95,6 @@ Future<void> _tryPushHelper(
       iOS: DarwinInitializationSettings(),
     ),
     onDidReceiveNotificationResponse: onSelectNotification,
-    //onDidReceiveBackgroundNotificationResponse: onSelectNotification,
   );
 
   client ??= (await ClientManager.getClients(initialize: false)).first;
--- a/lib/utils/background_push.dart
+++ b/lib/utils/background_push.dart
@@ -337,6 +337,11 @@ class BackgroundPush {
     final instance = BackgroundPush.clientOnly(client);
     instance._matrixState = matrixState;
     instance.onFcmError = onFcmError;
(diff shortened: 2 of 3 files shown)


lib/utils/push_helper.dart
```diff
@@ -95,7 +95,6 @@ Future<void> _tryPushHelper(
       iOS: DarwinInitializationSettings(),
     ),
     onDidReceiveNotificationResponse: onSelectNotification,
-    //onDidReceiveBackgroundNotificationResponse: onSelectNotification,
```
9clg6, line 98: Why did you deleted this ?

nqhhdev: - that line was already commented out before this PR because onDidReceiveBackgroundNotificationResponse requires a `@pragma('vm:entry-point')` top-level function — a closure won't work. The fix uses getNotificationAppLaunchDetails() instead.

jbrnzl (approved):

9clg6 (approved):

PR #3168, opened 2026-06-22
fix(linkify): prevent filenames from opening new tab in message caption

### LINKED ISSUE #3124: [BUG] Attached files are considered as link?

### Describe the bug

When I click on the file name, it opens a web browser instead of downloading the file or I don't know what to expect. But at least not opening a new tab on a non working URL 

https://github.com/user-attachments/assets/42c271b9-7859-4a11-adc2-1052f62a5d40

### Pre-Bug State

.

### To Reproduce

.

### Expected results

.

### Context

Chrome / LNG

### Screenshots or Video

_No response_

### Logs

_No response_

### Additional information

- Frequency
- Others

9clg6:
Closes https://github.com/linagora/twake-on-matrix/issues/3124

## Problem

The `linkfy_text` package detects filenames like `test.md` as URLs
(`.md` = Moldova ccTLD). This registers a tap handler that calls
`UrlLauncher.launchUrl()`, opening a new browser tab instead of doing nothing.

## Solution

Extracted `isFilenameUrl()` in `lib/utils/extension/url_extension.dart`.
A string is classified as a filename (not a URL) when **all** of these hold:
- no `http://` / `https://` scheme
- no `/`, `?`, or `#` (structural URL indicators)
- the **last** dot-delimited segment matches a known file extension set
  (multi-dot names like `my.report.pdf` and `archive.tar.gz` supported)

`TwakeLinkPreview` calls `isFilenameUrl()` in both `linkBuilder`
(prevents rendering as a clickable span) and `onTapDownLink` (skips
`UrlLauncher` call). Result: filename text in captions is plain, non-clickable.

## Trade-off

**Extension blocklist vs. TLD allowlist** — blocklist covers common formats
exhaustively; may miss rare custom extensions. The alternative (whitelist
known web TLDs) would be more principled but requires a much larger list
and still risks false negatives on new TLDs.

We classify on the **last** segment only (no dot-count restriction), so
multi-dot filenames work. The only residual ambiguity is a real multi-dot
domain whose final label collides with a file extension that is also a TLD
(e.g. `blog.example.md`, `.sh`, `.rs`) — these become non-clickable. Such
strings are effectively never typed in a caption, so we accept it in favor
of consistent filename handling. `google.com`, `github.io`, `wikipedia.org`
are unaffected.

## Unit tests — `isFilenameUrl` (28/28)

| Input | Expected | Result |
|---|---|---|
| `null` | `false` | ✅ |
| `""` | `false` | ✅ |
| `"https://example.com/test.md"` | `false` (explicit https) | ✅ |
| `"http://example.com/test.md"` | `false` (explicit http) | ✅ |
| `"https://test.md"` | `false` (explicit https, no path) | ✅ |
| `"google.com"` | `false` (bare .com domain) | ✅ |
| `"github.io"` | `false` (bare .io domain) | ✅ |
| `"wikipedia.org"` | `false` (bare .org domain) | ✅ |
| `"example.com/page"` | `false` (bare domain with path) | ✅ |
| `"test.md"` | `true` (markdown filename) | ✅ |
| `"document.pdf"` | `true` | ✅ |
| `"report.xlsx"` | `true` | ✅ |
| `"archive.zip"` | `true` | ✅ |
| `"image.png"` | `true` | ✅ |
| `"main.dart"` | `true` | ✅ |
| `"script.py"` | `true` | ✅ |
| `"video.mp4"` | `true` | ✅ |
| `"TEST.MD"` | `true` (case-insensitive) | ✅ |
| `"Report.Pdf"` | `true` (mixed case) | ✅ |
| `"my.report.pdf"` | `true` (multi-dot filename) | ✅ |
| `"v1.0.zip"` | `true` (multi-dot with version) | ✅ |
| `"archive.tar.gz"` | `true` (compound extension) | ✅ |
| `"sub.test.md"` | `true` (multi-dot, known extension) | ✅ |
| `"test.md/path"` | `false` (has path separator) | ✅ |
| `"test.md?query=1"` | `false` (has query string) | ✅ |
| `"test.md#section"` | `false` (has fragment) | ✅ |
| `"noextension"` | `false` (no dot) | ✅ |
| `"test.xyz123"` | `false` (unknown extension) | ✅ |

## Summary by CodeRabbit

* **New Features**
  * Improved link handling so file-like text is recognized separately from web URLs.
  * Added support for identifying filename-style links, including multi-dot names like `archive.tar.gz`.
* **Bug Fixes**
  * Prevented filename-like links from being treated as tappable web links.
  * Reduced false positives for strings with explicit schemes, paths, queries, or fragments.
* **Tests**
  * Added a test suite to validate filename-URL detection across common cases and edge conditions.

### DIFF
--- a/lib/utils/extension/url_extension.dart
+++ b/lib/utils/extension/url_extension.dart
@@ -1,0 +1,50 @@
+const _fileExtensions = {
+  // Documents
+  'pdf', 'doc', 'docx', 'xls', 'xlsx', 'ppt', 'pptx',
+  'odt', 'ods', 'odp', 'txt', 'csv', 'tsv', 'rtf',
+  // Markup / data
+  'md', 'json', 'xml', 'yaml', 'yml', 'toml', 'ini', 'cfg', 'conf',
+  // Images
+  'png', 'jpg', 'jpeg', 'gif', 'svg', 'webp', 'bmp', 'ico', 'tiff',
+  // Video / audio
+  'mp4', 'mkv', 'avi', 'mov', 'webm', 'mp3', 'wav', 'ogg', 'flac', 'aac',
+  // Archives
+  'zip', 'tar', 'gz', 'bz2', 'xz', '7z', 'rar', 'pkg', 'dmg', 'iso',
+  'apk', 'ipa', 'exe', 'msi',
+  // Code
+  'py', 'js', 'ts', 'jsx', 'tsx', 'dart', 'java', 'kt', 'swift',
+  'go', 'rs', 'c', 'cpp', 'h', 'cs', 'php', 'rb', 'sh', 'bash',
+  'zsh', 'fish', 'ps1', 'sql', 'html', 'css', 'scss',
+  // Fonts
+  'ttf', 'otf', 'woff', 'woff2',
+  // Misc
+  'log', 'lock', 'env',
+};
+
+bool _hasExplicitScheme(String url) =>
+    url.startsWith('http://') || url.startsWith('https://');
+
+bool _hasUrlStructuralIndicator(String url) =>
+    url.contains('/') || url.contains('?') || url.contains('#');
+
+bool _hasKnownFileExtension(String url) {
+  final lastDot = url.lastIndexOf('.');
+  if (lastDot == -1) return false;
+  return _fileExtensions.contains(url.substring(lastDot + 1).toLowerCase());
+}
+
+/// Returns true if [url] looks like a filename falsely detected as a URL
+/// by a linkify parser (e.g. "test.md" parsed as Moldova TLD).
+///
+/// Criteria:
+/// - no explicit http/https scheme
+/// - no path, query string, or fragment (structural URL indicators)
+/// - last dot-delimited segment matches a known file extension
+///   (multi-dot names like `my.report.pdf` or `archive.tar.gz` supported)
+bool isFilenameUrl(String? url) {
+  if (url == null) return false;
+  if (url.isEmpty) return false;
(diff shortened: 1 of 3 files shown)


lib/utils/extension/url_extension.dart
```diff
@@ -0,0 +1,51 @@
...
+  'log', 'lock', 'env',
+};
+
+bool _hasExplicitScheme(String url) =>
+    url.startsWith('http://') || url.startsWith('https://');
+
+bool _hasUrlStructuralIndicator(String url) =>
+    url.contains('/') || url.contains('?') || url.contains('#');
+
+bool _hasSingleKnownFileExtension(String url) {
```
nqhhdev: This rejects filenames with multiple dots (e.g. my.report.pdf, v1.0.zip, archive.tar.gz). Is that intentional, or should we support multi-dot filenames?

Image

9clg6: Good catch — fixed. Now matching only the last dot-delimited segment instead of requiring a single dot, so `my.report.pdf`, `v1.0.zip` and `archive.tar.gz` are handled. Added tests for these cases.

android/gradle.properties
jbrnzl, line 1: It this change intentional ?

jbrnzl: Oh ok I see other files, I think little mistake while rebasing @9clg6

9clg6: It was not rebased yet 😄

nqhhdev (approved):

jbrnzl (approved):

PR #3169, opened 2026-06-22
fix(media): use actual video dimensions in sending state preview

### LINKED ISSUE #3153: [BUG] Streched media preview on sending state

### Describe the bug

When you send a video in the timeline, the preview is streched until it is actually sent
https://github.com/user-attachments/assets/9e1181a3-ae28-4a1a-8882-25650673262c

### Pre-Bug State

NA

### To Reproduce

- Go to a room
- Send a video

### Expected results

The preview is exactly the same in sending state or sent state

### Context

- Device: Web, mobile
- App Version: 2.22.5

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

9clg6:
## Summary
- Refactor `SendingVideoWidget` to decode the placeholder thumbnail's actual pixel dimensions via `ui.instantiateImageCodec`, then recompute the display size with the correct aspect ratio
- Use `videoBubbleHeight` (matching `EventVideoPlayer`) instead of `imageBubbleHeight` for consistent sizing between sending and sent states
- Inline `VideoWidget` into `SendingVideoWidget` (no longer needed as a separate widget)

## Root cause
During video upload, the sending preview used default fallback dimensions (500×420 web / 256×320 mobile) because the event's `info` map has no `w`/`h` yet (especially on web where video metadata isn't extracted before the fake event). This caused the bubble to appear stretched until upload completed and the server-returned event provided real dimensions.

Now the widget reads the thumbnail's real dimensions once it loads and rebuilds with the correct aspect ratio.

## Test plan
- [x] Send a portrait video on web — preview matches sent state
- [x] Send a landscape video — verify correct aspect ratio
- [x] Send a square video — verify correct aspect ratio
- [x] Test on mobile (iOS/Android)

image
image
image

Fixes #3153

## Summary by CodeRabbit

* **Improvements**
  * Enhanced outgoing video messages with more reliable, self-contained thumbnail placeholder loading and rendering.
  * Thumbnails now display only after the placeholder image bytes are successfully decoded, with sizing aligned to the resolved display dimensions.

### DIFF
--- a/lib/pages/chat/events/sending_video_widget.dart
+++ b/lib/pages/chat/events/sending_video_widget.dart
@@ -1,3 +1,6 @@
+import 'dart:typed_data';
+import 'dart:ui' as ui;
+
 import 'package:fluffychat/di/global/get_it_initializer.dart';
 import 'package:fluffychat/pages/chat/events/message/message_style.dart';
 import 'package:fluffychat/pages/chat/events/message_content_style.dart';
@@ -6,6 +9,7 @@ import 'package:fluffychat/presentation/model/chat/upload_file_ui_state.dart';
 import 'package:fluffychat/presentation/model/file/display_image_info.dart';
 import 'package:fluffychat/utils/manager/upload_manager/upload_manager.dart';
 import 'package:fluffychat/utils/extension/build_context_extension.dart';
+import 'package:fluffychat/utils/extension/image_size_extension.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/event_extension.dart';
 import 'package:fluffychat/widgets/mixins/upload_file_mixin.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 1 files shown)


jbrnzl:
@9clg6 Nice, I tested and in most cases it resolves the issue, it's great ! 👍🏻 

But there is still some cases where the preview is streched as you can see:

https://github.com/user-attachments/assets/1e9029bf-a725-41ba-9572-e2a1912a2733


Do you know what we can do ?

The video to reproduce the issue (maybe it's not worth handling that case)
https://github.com/user-attachments/assets/77decab7-e927-4b4a-a9ee-e8a95b1811d4

9clg6:
> @9clg6 Nice, I tested and in most cases it resolves the issue, it's great ! 👍🏻
> 
> But there is still some cases where the preview is streched as you can see:
> 
>  Enregistrement.de.l.ecran.2026-06-23.a.11.51.58.mov 
> Do you know what we can do ?
> 
> The video to reproduce the issue (maybe it's not worth handling that case) https://github.com/user-attachments/assets/77decab7-e927-4b4a-a9ee-e8a95b1811d4

Thanks for you feedback, but I think it's an edge-case that doesn't need to be covered, because the risk of regression is real, and ratio effort/risk/gain is not interesting.

WDYT ?

jbrnzl:
> > @9clg6 Nice, I tested and in most cases it resolves the issue, it's great ! 👍🏻
> > But there is still some cases where the preview is streched as you can see:
> > Enregistrement.de.l.ecran.2026-06-23.a.11.51.58.mov
> > Do you know what we can do ?
> > The video to reproduce the issue (maybe it's not worth handling that case) https://github.com/user-attachments/assets/77decab7-e927-4b4a-a9ee-e8a95b1811d4
> 
> Thanks for you feedback, but I think it's an edge-case that doesn't need to be covered, because the risk of regression is real, and ratio effort/risk/gain is not interesting.
> 
> WDYT ?

I agree 👍🏻 We will see if this specific case is encountered by users or not but it's specific, so we can go without

jbrnzl (approved):
I tested it 👍🏻

nqhhdev (approved):
LGTM

PR #3159, opened 2026-06-18
TW-3132: Rewire Room/Chat Search Call Sites to TextSearch (Part 3)

aparesy:
## Ticket

- #3132

## Impact description
The purpose of this PR is to implement the search engine discussed in #3129 for concrete Room/chat searches. This way, when a user searches for a room in one of the several search bars whose behavior was modified, the search isn't diacritic-sensitive anymore and the match is done against substrings, and not complete words.
The four dart files impacted are all related to different room/chat searches
Note: As we can see in the demonstrations below, the highlightings of search matches can be faulty in some cases, but I suggest tackling this in another issue.

## Test recommendations
Some unit tests were implemented in the PRs related to #3130 and #3131. Below I'll indicate which widgets are concretely impacted.

## Resolved
- Web:
- `lib/utils/extension/presentation_search_extension.dart`: This file tackles the combined contacts + rooms results in the main search bar. It is notably used in SearchContactsAndChatsController.

https://github.com/user-attachments/assets/b1f8c83d-6a1b-4cab-953d-5402f9ab2d11

- `lib/domain/model/room/room_list_extension.dart`: Has a similar purpose for rooms in main search bar. Also used in SearchContactsAndChatsController.

https://github.com/user-attachments/assets/60464dbe-0399-40d2-8773-e89d2858c9a7

- `lib/pages/chat/input_bar/input_bar.dart`: In a chat room, type # to trigger room alias autocomplete.

https://github.com/user-attachments/assets/078435fa-5a11-4c39-b958-33b65586fc52

- `lib/presentation/mixins/search_recent_chat_mixin.dart`: In share or forward dialog, the room picker search. Used in Share screen (pages/share/share.dart) and Forward screen (pages/forward/forward.dart)

Example for forward: 
https://github.com/user-attachments/assets/d3dce98b-08e2-4b1e-bf63-ddf864be5c8c

Example for share: 

https://github.com/user-attachments/assets/b6e33bfe-3e18-4ffe-91f0-afcca668f766

## Summary by CodeRabbit

* **New Features**
  * Diacritic-insensitive search now more reliably matches accented variants across chat names, aliases, contact info, and room “#…” suggestions.

* **Bug Fixes**
  * Recent-chat and contact keyword matching updated so queries return correct results when users type different diacritic forms.
  * Room “#…” suggestions improved for Unicode input and matching against canonical aliases and display names.

* **Tests**
  * Expanded search scenarios to cover accented-title matching and updated test setup for shared search behavior.

Issue #3137, opened 2026-06-12
Rewire InputBar & StickerPicker Search to `TextSearch`


pm-McFly:
# Description

Replace `.toLowerCase().contains()` patterns in the InputBar autocomplete (commands, emotes, emojis, mentions, rooms) and the StickerPicker filter with `TextSearch` calls (`matchesText` / `matchAnyField`), preserving existing behavior.

# Options to pass
 
Use:

- case-insensitive,
- diacriticSensitive: false,
- substring mode

# Implementation checklist

- [ ] Identify call sites
- [ ] Ensure all existing tests pass; add parity tests.

aparesy:
Resolved in #3184

Issue #3136, opened 2026-06-12
Rewire Stories Header Search to `TextSearch`


pm-McFly:
# Description

Replace the `.toLowerCase().contains()` filter in the Stories header with `TextSearch`, preserving existing behavior.

# Options to pass
 
Use:

- case-insensitive,
- diacriticSensitive: false,
- substring mode

# Implementation checklist

- [ ] Identify call sites
- [ ] Ensure all existing tests pass; add parity tests.

Issue #3135, opened 2026-06-12
Rewire Chat Details Search Call Sites to `TextSearch`


pm-McFly:
# Description

Replace the identical `.toLowerCase().contains()` search patterns across all Chat Details member-filtering screens with `TextSearch` calls. Four call sites share the exact same pattern.

# Options to pass
 
Use:

- case-insensitive,
- diacriticSensitive: false,
- substring mode

# Implementation checklist

- [ ] Identify call sites
- [ ] Ensure all existing tests pass; add parity tests.

aparesy:
Addressed in #3180

PR #3173, opened 2026-06-23
TW-3133 rewire contacts search call sites to text search (Part 4)

aparesy:
## Ticket
**Related issue**

- #3133

## Impact description
This PR implements the rewiring of contacts search call sites according to the specifications in #3133. There are three modified code files, and three files related to testing. It has been deemed more pertinent to add unit tests rather than integration tests in this case because the components modified are mostly separate from the application-side logic.

## Test recommendations

To run the tests related to this PR: `flutter test test/domain/extensions/contact/ test/utils/presentation_search_extension_test.dart`

## Resolved
- Web:

- `contact_extension.dart`: modifies the main contacts search

https://github.com/user-attachments/assets/18a21294-9568-4659-ac42-aa9064dd2c5e

- `presentation_search_extension.dart`: was already modified in #3159, added small changes to the phone number and email matching, mainly to update it and make the email search case and diacritic insensitive

Screenshot From 2026-06-23 13-53-03

- `address_book_extension.dart`: It seems like `searchAddressBooks` isn't called anywhere so I could not test it, but since the changes are extremely similar to the changes done to the other files in this PR and in other rewiring PRs (#3132 #3134 #3135 #3136 #3137), it is very likely that there is no issue with the address book search

## Summary by CodeRabbit

* **Improvements**
  * Refined contact search using a unified matcher across contact names, IDs, email addresses, and phone numbers (including multiple emails/phone numbers), with better case/diacritic tolerance and more accurate empty-field handling.
  * Improved recent chat search and room suggestions using consistent matching across localized names and room aliases, with more reliable result limiting.
  * Enhanced handling for special characters and emoji, with diacritic folding applied to the search term as well.
* **Tests**
  * Expanded coverage for keyword matching scenarios across contacts, chats, and presentation search.


aparesy:
Fixup for some tests adjustments (added `const`, initialized GetIt in `test/domain/model/address_book_test.dart`

9clg6 (dismissed):

jbrnzl:
> address_book_extension.dart: It seems like searchAddressBooks isn't called anywhere so I could not test it, but since the changes are extremely similar to the changes done to the other files in this PR and in other rewiring PRs (https://github.com/linagora/twake-on-matrix/issues/3132 https://github.com/linagora/twake-on-matrix/issues/3134 https://github.com/linagora/twake-on-matrix/issues/3135 https://github.com/linagora/twake-on-matrix/issues/3136 https://github.com/linagora/twake-on-matrix/issues/3137), it is very likely that there is no issue with the address book search

@aparesy Hi !
I think we can delete `searchAddressBooks` and `IterableAddressBookExtension` then. It looks like they were forgotten in a previous refactor/cleanup :)
In the same file, `combineDuplicateAddressBooks` and `AddressBookListExtension` are unused too.
Either we include these changes in this PR, or we open a separate one, what do you think ?

aparesy:
Hi, yes I agree that it would be cleaner to not make those modifications. Personally I'd argue that it's justified to delete the old functions related to address book search because we are doing a complete rewiring and it would be cleaner to only leave the improved search. I'd say that we could also delete the two other functions if they are not used anyways

aparesy:
Okay I also removed some tests

jbrnzl (dismissed):

jbrnzl (approved):

nqhhdev (approved):
LGTM

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -329,50 +331,32 @@ extension SetContactExtension on Set<Contact> {
...
+                (phone) =>
+                    engine.matchesText(
+                      keyword.replaceAll(' ', ''),
+                      phone.number.replaceAll(' ', ''),
+                      options: options,
+                    ) ||
+                    matches(phone.matrixId),
+              ) ==
+              true,
+    );
```
pm-McFly: You are over engineering your own code by only relying on the `matchesText` member function.
While it brings a cool realy return conditional `||` matching, it brings a lot of complexity at reading and up-keep.

You can leverage `matchAnyField` to do the same job, in a simpler way:

<details>
<summary>building the contact `haystack` before extraction </summary>

```suggestion
    return where(
      (contact) => engine.matchAnyField<String?>(
        keyword,
        [
          contact.displayName,
          contact.id,
          ...?contact.emails?.map((e) => e.address),
          ...?contact.phoneNumbers?.map((p) => p.number),
        ],
        fieldExtractors: [(s) => s],
        options: options,
      ).isNotEmpty
    );
```

</details>

But before applying consider the general review comment that suggests an API change to SearchEngine

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -329,50 +331,32 @@ extension SetContactExtension on Set<Contact> {
...
-          emailContains ||
-          emailMatrixIdContains ||
-          phoneMatrixIdContains ||
-          contact.id.contains(keyword);
-    });
-    return contactsMatched;
+                (phone) =>
+                    engine.matchesText(
+                      keyword.replaceAll(' ', ''),
+                      phone.number.replaceAll(' ', ''),
```
pm-McFly:
```suggestion
                      keyword.replaceAll(RegExp(r'\s+'), ''),
                      phone.number.replaceAll(RegExp(r'\s+'), ''),
```

It cacthes invisible spacer chars better and as chunks instead of only `whitespaces` one by one.

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -329,50 +331,32 @@ extension SetContactExtension on Set<Contact> {
...
+                      options: options,
+                    ) ||
+                    matches(phone.matrixId),
```
pm-McFly: isn't `phone.matrixId` same as `contact.id`?

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -329,50 +331,32 @@ extension SetContactExtension on Set<Contact> {
...
+
+    bool matches(String? field) =>
+        engine.matchesText(keyword, field ?? '', options: options);
+
+    return where(
+      (contact) =>
+          matches(contact.displayName) ||
+          matches(contact.id) ||
+          contact.emails?.any(
+                (email) => matches(email.address) || matches(email.matrixId),
```
pm-McFly: isn't `email.matrixId` same as `contact.id`?

pm-McFly (changes requested):
Sometimes we have to handle data structures that have fields with variable length. I left on purpose a spec hole in the [Reference implementation shape](https://github.com/linagora/twake-on-matrix/issues/3129#issue-4649798520) a while back, but hinted it in the [Test recommandations](https://github.com/linagora/twake-on-matrix/pull/3156#issue-4681723561)

Here `phoneNumbers` is a list for example. Although, `displayName` or `id` are plain strings. And the initially suggested `SearchEngine` implementation proposed `fieldExtractors` being `List<String? Function(T)>` and thus leaving the length variation handling on the caller site.

A solution to a more maintable code is to push that logic in the `SearchEngine`:
- Tested once: unit tests of `SearchEngine`
- Caller only defines which fields exist, not which special treatment is needed
- One element sized lists brings "almost no overhead" and most of it's complexity is resolved to O(1)

A suggested change is the following:

```dart
class SearchEngine {
  const SearchEngine();

  List<NormalizationStep> _buildPipeline(SearchOptions options) => [
    if (!options.diacriticSensitive)
      DiacriticStripStep(decomposition: options.diacriticDecomposition),
    if (!options.caseSensitive) const LowerCaseStep(),
  ];

  /// Internal helper to generate normalization and matching closures.
  ({String Function(String) apply, bool Function(String) matches}) _matcher(
    String needle,
    SearchOptions options,
  ) {
    final pipeline = _buildPipeline(options);
    String apply(String v) => pipeline.fold(v, (acc, s) => s.normalize(acc));
    final normNeedle = apply(needle);
    
    bool matches(String v) => switch (options.mode) {
      SearchMode.exact => v == normNeedle,
      SearchMode.substring => v.contains(normNeedle),
    };
    
    return (apply: apply, matches: matches);
  }

  /// Returns `true` if `needle` matches at least one item in `haystack`.
  bool anyMatch<T>(
    String needle,
    List<T> haystack, {
    List<Iterable<String> Function(T)>? fieldExtractors,
    SearchOptions options = const SearchOptions(),
  }) {
    final extractors = fieldExtractors ?? [(T s) => [s.toString()]];
    final matcher = _matcher(needle, options);
    
    return haystack.any((item) =>
      extractors.any((extract) => 
        extract(item).any((v) => matcher.matches(matcher.apply(v)))
      )
    );
  }

  /// Returns a `List<T>` of all items in `haystack` matching `needle`.
  List<T> match<T>(
    String needle,
    List<T> haystack, {
    List<Iterable<String> Function(T)>? fieldExtractors,
    SearchOptions options = const SearchOptions(),
  }) {
    final extractors = fieldExtractors ?? [(T s) => [s.toString()]];
    final matcher = _matcher(needle, options);
    
    return haystack.where((item) =>
      extractors.any((extract) => 
        extract(item).any((v) => matcher.matches(matcher.apply(v)))
      )
    ).toList();
  }

  // --- Original Base Utility Methods ---

  List<T> matchAnyField<T>(
    String needle,
    List<T> haystack, {
    required List<String? Function(T)> fieldExtractors,
    SearchOptions options = const SearchOptions(),
  }) {
    final matcher = _matcher(needle, options);

    return haystack.where((item) {
      return fieldExtractors.any(
        (extract) => matcher.matches(matcher.apply(extract(item) ?? '')),
      );
    }).toList();
  }

  /// Convenience utility for a simple text-to-text comparison.
  bool matchesText(String needle, String haystack, {SearchOptions options = const SearchOptions()}) =>
      anyMatch(needle, [haystack], options: options);
}
```

lib/domain/model/extensions/contact/contact_extension.dart
aparesy: yes definitely agreed on that

lib/domain/model/extensions/contact/contact_extension.dart
aparesy: It seems that contact.id is a separate identifier and does not always correspond to a Matrix Id, it is for example defined in `lib/data/datasource_impl/contact/phonebook_contact_datasource_impl.dart` where it is taken from the phone's address book, or, if that didn't work, uses a hash of the email/phone as a fallback.
This is also coherent with the fact that the commit that first adds these phone.matrixId/email.matrixId is this: https://github.com/linagora/twake-on-matrix/commit/ed5783ab9fa435660f62fbe0566a4a5034fd0269

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -329,34 +329,48 @@ extension SetContactExtension on Set<Contact> {
...
+          fieldExtractors: [
+            (Contact c) =>
+                c.phoneNumbers?.map(
+                  (p) => p.number.replaceAll(_phoneWhitespace, ''),
+                ) ??
+                const [],
+          ],
+          options: options,
+        )
+        .toSet();
```
pm-McFly: I see why you are doing this, but tbh that's will be too much to implement now...
It add strip whitespaces for the phonenumber set and for the "needle" you got from the user query.
But bc the user will sometimes use "+33..." or "06..." or maybe "Timéo" or whatever else, we can skip that kind of specific logic for now.

I mean, when we are matching against emails or mxid we don't format anything, so leave it for phone as well and that will be handled later when doing:
- Partial matching
- Fuzzy matching
- Filters
- etc.

Apply the same logic everywhere first, we will deal with special cases at time.

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -232,28 +232,30 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
...
-            state != null &&
-            ((alias is String && matchesAlias(alias)) ||
-                (altAlias is List &&
-                    altAlias.any((l) => l is String && matchesAlias(l))));
-        final nameMatches = engine.matchesText(
+        final aliasCandidates = [
+          if (alias is String) alias.split(':')[0],
+          if (altAlias is List)
+            ...altAlias.whereType<String>().map((a) => a.split(':')[0]),
+        ];
```
pm-McFly: I know we had the split before, but I don't think it is a smart move somehow...
Room aliases are human friendly names that like matrixid can be re-used over the network.

So you can encounter:
- `#myroom:linagora.com`
- `#myroom:twake.app`

And if user is looking for `myroom` which result do you present here?
Is the matrix address behind it correctly mapped to the result?

I think we can - just like for phone number - apply the most basic system now (exact match on the full alias), and see later with more advanced matching

The room search by name will be enough at day-1, so users will unlikely use aliases for searching today

test/domain/extensions/contact/contact_extension_test.dart
```diff
@@ -453,6 +453,39 @@ void main() {
...
+          Contact(
+            id: 'contact1',
+            phoneNumbers: {
+              PhoneNumber(number: '+33\u{00A0}6\u{00A0}12\u{00A0}34'),
+            },
+          ),
+        ];
+        expect(contacts.searchContacts('+3361234').length, 1);
+      },
+    );
```
pm-McFly: (This comment also applies to other test scenarii) Also test when the needle contains weird chars.

1. `a = à`
2. `à = a`

You are testing 1. but never 2.

pm-McFly (changes requested):

lib/domain/model/extensions/contact/contact_extension.dart
Crash--: > when we are matching against emails or mxid we don't format anything

We should at least trim()

aparesy: I feel like trim would be a good fit for a new normalization step that we could add right after being done with the rewiring, as it is general enough to be added everywhere

lib/pages/chat/input_bar/input_bar.dart
aparesy: Yes I agree, it is a lot simpler this way

test/domain/extensions/contact/contact_extension_test.dart
aparesy: Got it :+1:

aparesy:
I cleaned up the commits and fixed the integration tests, assuming that #3246 will be merged anyways. This simplifies the changes brought by this PR a little bit

jbrnzl:
@aparesy What do you think about this code rabbit feedback ? It seems legit to me
> lib/utils/search/search_engine.dart (1)> `76-89`: _🚀 Performance & Scalability_ | _🔵 Trivial_ | _⚡ Quick win_
> **Make `anyMatch` short-circuit instead of materializing matches.**
> `anyMatch` currently builds the full matched list via `match(...).isNotEmpty`. Use `haystack.any(...)` directly so large searches stop at the first hit.

aparesy:
> @aparesy What do you think about this code rabbit feedback ? It seems legit to me
> 
> > lib/utils/search/search_engine.dart (1)> `76-89`: _🚀 Performance & Scalability_ | _🔵 Trivial_ | _⚡ Quick win_
> > **Make `anyMatch` short-circuit instead of materializing matches.**
> > `anyMatch` currently builds the full matched list via `match(...).isNotEmpty`. Use `haystack.any(...)` directly so large searches stop at the first hit.

Yep makes sense to me, I added it

9clg6 (approved):

lib/domain/model/extensions/contact/contact_extension.dart
```diff
@@ -333,56 +331,23 @@ extension SetContactExtension on Set<Contact> {
...
-      return plainTextContains ||
-          phoneNumberContains ||
-          emailContains ||
-          emailMatrixIdContains ||
-          phoneMatrixIdContains;
-    });
-    return contactsMatched;
-  }
+    if (keyword.isEmpty) return this;
+    final engine = getIt.get<SearchEngine>();
```
pm-McFly, line 335: @9clg6 are the `getIt` calls in this PR correctly bound to Riverpod?

pm-McFly (approved):

lib/domain/model/extensions/contact/contact_extension.dart
aparesy: Ah when we discussed about it last week I think we said that we could rewire them later in a batch? Not sure though

pm-McFly: Okay seen together, follow up PR to use GetIt wrapped in Riverpod

Issue #3176, opened 2026-06-24
[BUG] Wrong online status


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The user is currenly typing and sending messages, but his status is 'online 6 min ago'

https://github.com/user-attachments/assets/e9083969-8c42-456c-ad58-0f26c779835e

### Pre-Bug State

N/A

### To Reproduce

It's a sometimes issue, but the online/last seen status isn't always displayed correctly 
In the video user is typing, but when user stops typing, the status turns to 'online 6 min ago', though the user is currently online and is sending messages 

### Expected results

The online status is changed correctly

### Context

- Device: Macbook Air
- OS: MacOS
- Browser: Safari
- App Version: 2.22.5
- Server: [e.g. production, example.com]
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
Closed by #3176

PR #3183, opened 2026-06-25
TW-3176: implement listenSyncPresence for deduplicating user presence updates

nqhhdev:
## Ticket
**Related issue**

- #3176 

- The old code picked one presence per sync — the one with the most recent lastActiveTimestamp across all users. So if Bob and Carol both had presence events in the same sync, Carol's more recent timestamp would win and Bob's update was silently dropped. The UI never received Bob's presence change and stayed stuck at the stale "online xx min ago".- 

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/4445c7b4-bb03-4da0-91c4-dab468379d70

## Summary by CodeRabbit

* **Bug Fixes**
  * Presence synchronization now deduplicates updates per user within the same sync and shows the most recently active status for each user.
  * Improved timestamp handling, including correct behavior when timestamps are missing or equal.
* **Tests**
  * Added new unit test utilities and a comprehensive test suite covering single/multi-user syncs, de-duplication correctness, missing-timestamp fallbacks, and independence across sequential syncs.

### DIFF
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -252,20 +252,25 @@ class MatrixState extends State<Matrix>
   void _listenSyncPresence(Client client) {
     _presenceSubscription?.cancel();
     _presenceSubscription = client.onSync.stream.listen((sync) {
-      CachedPresence? lastActivePresence;
+      // Deduplicate per user: keep only the most recent presence per userid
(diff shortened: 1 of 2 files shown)


9clg6 (approved):

jbrnzl (approved):
Nice fix 👍🏻

PR #3184, opened 2026-06-25
TW-3137: rewire input bar sticker picker search to text search (Part 7)

aparesy:
## Ticket
- #3137

## Impact description
Last rewiring of the series, impacting the input bar. The sticker picker is also affected, but no way to see it since it seems like it is not used yet. In `input_bar.dart`, I changed most `.contains()` with a call to the searchEngine, with the exception of the emoji search as it most of the time the user will not mistakenly input a diacritic to search for an emoji. Also, the search becomes slow when trying to wire the emoji search to the search engine as the haystack has thousands of entries, which demonstrates the potential need for enhancement of `searchEngine` in the future. 

## Resolved
- Web:

https://github.com/user-attachments/assets/0d591292-ad8b-49fb-ac24-e60406285575

## Summary by CodeRabbit

- **New Features**
  - Enhanced chat suggestion matching for commands, emojis/emotes, mentions, and rooms with more accurate detection and improved Unicode support.
  - Upgraded sticker picker search to use smarter filtering across sticker keys and descriptions.
- **Bug Fixes**
  - Improved matching robustness by reducing accent/diacritic sensitivity.
  - Added Unicode-aware behavior for more consistent emoji/emote and mention recognition across diverse character sets.

### DIFF
--- a/lib/pages/chat/input_bar/input_bar.dart
+++ b/lib/pages/chat/input_bar/input_bar.dart
@@ -23,7 +23,6 @@ import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:linagora_design_flutter/linagora_design_flutter.dart';
 import 'package:matrix/matrix.dart';
 import 'package:scroll_to_index/scroll_to_index.dart';
-import 'package:slugify/slugify.dart';
 
 class InputBar extends StatefulWidget {
   final Room? room;
@@ -96,12 +95,14 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
     );
     final List<Map<String, String?>> ret = <Map<String, String?>>[];
     const maxResults = 30;
+    final engine = getIt.get<SearchEngine>();
+    const opts = SearchOptions(diacriticSensitive: false);
 
     final commandMatch = RegExp(r'^/(\w*)$').firstMatch(searchText);
     if (commandMatch != null && widget.room != null) {
       final commandSearch = commandMatch[1]!.toLowerCase();
       for (final command in widget.room!.client.commands.keys) {
-        if (command.contains(commandSearch)) {
+        if (engine.matchesText(commandSearch, command, options: opts)) {
           ret.add({'type': 'command', 'name': command});
         }
 
@@ -109,7 +110,8 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
       }
     }
     final emojiMatch = RegExp(
-      r'(?:\s|^):(?:([-\w]+)~)?([-\w]+)$',
+      r'(?:\s|^):(?:([-\w\p{L}]+)~)?([-\w\p{L}]+)$',
+      unicode: true,
     ).firstMatch(searchText);
     if (emojiMatch != null && widget.room != null) {
       final packSearch = emojiMatch[1];
@@ -118,7 +120,7 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
       if (packSearch == null || packSearch.isEmpty) {
         for (final pack in emotePacks.entries) {
           for (final emote in pack.value.images.entries) {
-            if (emote.key.toLowerCase().contains(emoteSearch)) {
+            if (engine.matchesText(emoteSearch, emote.key, options: opts)) {
               ret.add({
                 'type': 'emote',
                 'name': emote.key,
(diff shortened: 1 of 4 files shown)


9clg6 (dismissed):

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -96,20 +95,23 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
...
           ret.add({'type': 'command', 'name': command});
         }
 
         if (ret.length > maxResults) return ret;
       }
     }
     final emojiMatch = RegExp(
-      r'(?:\s|^):(?:([-\w]+)~)?([-\w]+)$',
+      r'(?:\s|^):(?:([-\w\p{L}]+)~)?([-\w\p{L}]+)$',
+      unicode: true,
```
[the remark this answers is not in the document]
aparesy: True, will address that

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -197,22 +199,27 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
...
+        if (engine.matchesText(
+              userSearch,
+              user.displayName ?? '',
+              options: opts,
+            ) ||
+            engine.matchesText(
+              userSearch,
+              user.id.split(':')[0],
+              options: opts,
+            )) {
```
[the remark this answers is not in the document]
aparesy: The main difference here is for space-separated needles, but since we are in the context of username matching (e.g. searching `@joseph`) I feel like the slugify is not that necessary so this proposed change is a little counter-intuitive

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -23,7 +23,6 @@ import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:linagora_design_flutter/linagora_design_flutter.dart';
 import 'package:matrix/matrix.dart';
 import 'package:scroll_to_index/scroll_to_index.dart';
-import 'package:slugify/slugify.dart';
```
jbrnzl, line 26: I didn't know about this package ! From what I see here it is the only use in the app so if we are really deleting it, we can also remove the dependency in the pubspec.yaml :)

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -96,20 +95,23 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
...
     const maxResults = 30;
+    final engine = getIt.get<SearchEngine>();
+    const opts = SearchOptions(diacriticSensitive: false);
 
     final commandMatch = RegExp(r'^/(\w*)$').firstMatch(searchText);
     if (commandMatch != null && widget.room != null) {
       final commandSearch = commandMatch[1]!.toLowerCase();
       for (final command in widget.room!.client.commands.keys) {
-        if (command.contains(commandSearch)) {
+        if (engine.matchesText(commandSearch, command, options: opts)) {
```
jbrnzl: What about passing commandMatch[1] directly here ? The caseSensitive param is false by default so .toLowerCase() seems redundant ?

lib/pages/chat/input_bar/input_bar.dart
```diff
@@ -197,22 +199,27 @@ class _InputBarState extends State<InputBar> with PasteImageMixin {
...
     // and captures the username (excluding the @ symbol) for further use
-    const userMentionsRegex = r'(?:\s|^)@([-\w]*)$';
-
-    final userMatch = RegExp(userMentionsRegex).firstMatch(searchText);
+    final userMatch = RegExp(
+      r'(?:\s|^)@([-\w\p{L}]*)$',
+      unicode: true,
+    ).firstMatch(searchText);
     if (userMatch != null && widget.room != null) {
       final userSearch = userMatch[1]!.toLowerCase();
```
jbrnzl: maybe this .toLowerCase() too

lib/pages/chat/input_bar/input_bar.dart
aparesy: Yep will do!

lib/pages/chat/input_bar/input_bar.dart
aparesy: Oh yeah that's an oversight on my part, sorry

lib/pages/chat/input_bar/input_bar.dart
aparesy: Yes, same change

jbrnzl (approved):
> Also, the search becomes slow when trying to wire the emoji search to the search engine as the haystack has thousands of entries, which demonstrates the potential need for enhancement of searchEngine in the future.

Maybe we can create a follow-up issue/discussion for this topic to not loose track ?
Maybe make the matcher engine API public can be easy and help a lot ?

aparesy:
There is already a discussion thread dedicated to the search feature at #3138, but yes I will probably open a discussion to discuss specifically the performance of it in the next few weeks as it is definitely one of the priorities
For the second point, I think this is a good idea, could help us pinpoint the exact expectations as well

9clg6 (approved):

Issue #3191, opened 2026-06-29
Disable the ability to hide deleted messages


tprudentova:
Current behaviour: in the Settings -> Chat user can choose to either hide or display deleted messages. Hiding them causes some issues with counters 

Expected behaviour: the toggle "Hide deleted messages" is removed from Settings -> Chat, deleted messages are displayed in chat list for all users as a bubble "Deleted message" 

<img width="1392" height="818" alt="Image" src="https://github.com/user-attachments/assets/fe38d95e-bf22-4b8f-808f-ce7824d16612" />

jbrnzl:
Closed by #3191

PR #3193, opened 2026-06-30
TW-3191: always display deleted messages

### LINKED ISSUE #3037: [BUG] New message displayed, but no?

jbrnzl:
## Ticket
- Implements #3191
- Fixes #3037

It fixes #3037 because the "phantom" unread badges were caused by the "hide deleted messages" setting: the room unread count comes directly from Synapse, so when the unread event was a hidden deleted message, the user couldn't see it to mark it read and the badge never cleared and should never have even appeared in the first place but this is an other thing.

## New behaviour
Now you still have the notification + unread badge and when you open the room deleted messages are now always visible (this last part is what #3191 is about).

## Changes
- Removed the "Hide deleted messages" toggle from Settings → Chat.
- Removed the `hideRedactedEvents` setting end-to-end (config flag, storage, l10n).
- Deleted messages now always render as a "Deleted message" bubble for everyone.

## More infos
What if in the future we still have a notification badge count but nothing is shown in UI ?
- First, that notification badge should disappear when you enter the room as per #3185 changes.
- And it will probably be because we don't yet support this type of event (ie location sharing,...) and that you enabled "hide unknown events".

## Summary by CodeRabbit

* **Bug Fixes**
  * Redacted messages are now shown consistently in the app, including pinned content and timeline visibility.
  * Removed the “hide redacted events” option from settings and configuration samples, simplifying available privacy controls.
  * Updated translations to match the current set of settings and labels.


Crash--:
Did we think about adding this message deletion logic to Synapse? If the unread counter comes from Synapse, maybe we could fix it there instead.

9clg6 (approved):

jbrnzl:
> Did we think about adding this message deletion logic to Synapse? If the unread counter comes from Synapse, maybe we could fix it there instead.

Yes, I think this was discussed in #3037. I had suggested a kind of server-side filter:

> Store user preference to show/hide events in a server so that notificationCount can be filtered accordingly to the user preferences.

but @pm-McFly pointed out it is not very viable

> Not possible without modifying the synapse server, and I feel it like a last option change...
>
> The only "server-side" option I see is setting a push rule on the m.room.redaction events but that would silent the "User has deleted a message" event, and I fear unexpected side effects.

So the decision was to stop hiding deleted messages so that what we render stays consistent with Synapse counts. Maybe there is other things to explore but I am not really convinced that "synapse keeping a message counted after it's redacted" behavior is something we would like to change (not sure of all the consequences/ side effects)

Crash--:
> Not possible without modifying the synapse server, and I feel it like a last option change...

This statement should be moderated. If we need to modify synapse to simplify stuff, let's do it in a smart way. 

I'm not saying that this is a bad choice to always display "deleted message" (whatapps do that and other but Telegram hide this stuff), but it's a bad choice to say "we don't touch synapse". We can.

pm-McFly:
> This statement should be moderated. If we need to modify synapse to simplify stuff, let's do it in a smart way. [ ... ] it's a bad choice to say "we don't touch synapse". We can.

I place this option as a last resort as it mean updating synapse in a way that will almost certainly not be merged (in the specs or in the upstream) and would obliged us to maintain a fork just for that simple need.

The Specs specify that notifications counting is computed based on the push rules: https://spec.matrix.org/v1.18/client-server-api/#receiving-notifications

> Servers MUST include the number of unread notifications in a client’s [/sync](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3sync) stream, and MUST update it as it changes. Notifications are determined by the push rules which apply to an event.

With an additional note that if a message is encrypted, depending on the push rule, clients are expected to resolve push rules on their as well.

I think that a better option is then to either:

- forbid hidding the deleted message: _in challenge currently_
- re-compute the notification counter locally ; imo a mess to do: _not relevant enough_
- setting push rules on `m.room.redaction` when "hide deleted messages" toggle is on and see what happens: _maybe best alternative to always see those messages_

tprudentova:
For now we display deleted messages to everyone and remove the setting to hide them. Later, if we get a lot of complaints, we can change the behaviour, but for now this flow is good, verified with Diana

nqhhdev (approved):
LGTM

PR #3194, opened 2026-07-01
TW-3065: Warn user when not able to retry sending

### LINKED ISSUE #3065: [BUG] Sometimes it's not possible to use the Retry button

### Describe the bug

Users report that the Retry button for messages and videos sonetimes doesn't do anything. My working theory is that it doesn't do anything if the network is still unstable, but I can't reproduce on my side even with throtting

Benoit reported the issue yesterday at 19:44 Paris time, maybe you can check Sentry? 

Image

### Pre-Bug State

Need to have bad/absent network

### To Reproduce

1. Send a message/media with an error 
2. Click the Retry button 
3. Nothing happens, the message isn't resent

### Expected results

When user clicks Retry, the message is attempted to resend despite the network conditions

### Context

- App Version: [e.g. 22]
- Server: [e.g. production, example.com]
- Account information:
- Others

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

nqhhdev:
## Ticket
**Related issue**

- Fixes #3065

### Root Cause:

- `UploadFileInfo` stores file bytes (matrixFile) and bytes live in RAM only, never serialized. On web, F5 wipes RAM. The sending event persists in local Matrix DB with status = sending/error but bytes are gone — upload can never complete or retry. 
- Note: Mobile is unaffected — fileInfo (file path on disk) IS serialized and restored correctly.  

### Steps to Reproduce :
1. Go offline (or have poor connection)
2. Send a video/image/file → upload fails
3. F5 (page refresh)
4. Come back online → open chat
5. Result: Stuck event visible in timeline, retry button does nothing  

### Solution:

On retry, if data is lost, show a snackbar

## Demo

https://github.com/user-attachments/assets/e660297e-f68e-40e3-a808-13b6db05f8f4

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -3261,5 +3261,10 @@
   "resetCompleteDescription": "Your new encryption identity is ready. Previously encrypted messages are no longer available.",
   "verificationFailedTitle": "Verification failed",
   "verificationFailedDescription": "The request was cancelled or rejected. Please try again.",
-  "retryFailedDescription": "Automatic retry didn't work. Try another way to verify this device."
+  "retryFailedDescription": "Automatic retry didn't work. Try another way to verify this device.",
+  "messageCannotBeResent": "This message can't be resent because its data is unavailable",
+  "@messageCannotBeResent": {
+    "type": "text",
+    "placeholders": {}
+  }
 }
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1269,76 +1269,6 @@ class ChatController extends State<Chat>
     }
   }
 
-  /// After timeline loads, remove stuck sending file events that can never
-  /// complete because their file data is gone (web page refresh wipes RAM).
-  ///
-  /// Root cause: on web, `matrixFile` bytes live in RAM only
-  /// (`@JsonKey(includeToJson: false)`). After a page refresh the bytes are
-  /// gone but the event stays in the local DB as `sending/error` forever.
-  /// On mobile, `fileInfo` (a file path) IS serialized and can be restored,
-  /// so mobile events are never affected.
-  Future<void> _removeSendingFileEventsWithLostData() async {
-    final room = this.room;
-    final events = timeline?.events;
-    if (events == null || room == null) return;
-
-    final uploadManager = getIt.get<UploadManager>();
-
-    // Collect first, then remove — cancelSend() triggers handleSync/onUpdate
-    // which would mutate the list mid-iteration.
-    final stuckEvents = <Event>[];
-    for (final event in events) {
-      if (!event.status.isSending && event.status != EventStatus.error) {
-        continue;
-      }
-      if (event.messageType != MessageTypes.Video &&
-          event.messageType != MessageTypes.Image &&
-          event.messageType != MessageTypes.File) {
-        continue;
-      }
-
-      // Only remove events tracked by our upload system (have upload_info in
-      // unsigned). Events without upload_info are from before this feature or
-      // from other clients — leave them alone.
-      final hasUploadInfo =
-          event.unsigned?['upload_info'] is Map<String, dynamic>;
-      if (!hasUploadInfo) continue;
-
-      // getUploadFileInfo checks in-memory map first, then falls back to
-      // restoring from unsigned JSON (works for mobile FileInfo paths).
-      // If either fileInfo or matrixFile is available the upload can proceed.
-      final uploadInfo = await uploadManager.getUploadFileInfo(
-        event.eventId,
-        room: room,
-      );
-      if (uploadInfo != null &&
-          (uploadInfo.fileInfo != null || uploadInfo.matrixFile != null)) {
-        continue;
-      }
-
-      // upload_info exists but no usable bytes → stuck web event, safe to remove.
-      stuckEvents.add(event);
-    }
-
-    for (final event in stuckEvents) {
-      try {
-        Logs().w(
-          'Chat::_removeSendingFileEventsWithLostData(): '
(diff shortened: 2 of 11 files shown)


nqhhdev:
Fixed: Show uploading content for events with error status


https://github.com/user-attachments/assets/44027c49-55df-4dfb-b5de-edf34df8fb30

nqhhdev:
Improve upload progress display for files


https://github.com/user-attachments/assets/86c63c0a-be74-4e37-b284-a5036bf8dc70

9clg6 (approved):

jbrnzl:
Oh wow, I am a little bit surprised, not about the implementation which seems great but more about the mechanism that silently deletes a message that is in error state after refresh, not sure it gives a good user experience and it feels like we will have feedbacks from users not knowing why messages they sent are disappearing (because maybe they did not saw the error state and they switched to an other conversation, and maybe they want to be warned or have a trace of what happened) 
@tprudentova can you please watch the different videos and confirm or not this behavior is acceptable from a product perspective ?

nqhhdev:
> Oh wow, I am a little bit surprised, not about the implementation which seems great but more about the mechanism that silently deletes a message that is in error state after refresh, not sure it gives a good user experience and it feels like we will have feedbacks from users not knowing why messages they sent are disappearing (because maybe they did not saw the error state and they switched to an other conversation, and maybe they want to be warned or have a trace of what happened) 
> 
> @tprudentova can you please watch the different videos and confirm or not this behavior is acceptable from a product perspective ?

I only remove the event when bytes are null.  Otherwise, I keep it unchanged @jbrnzl

jbrnzl:
> > Oh wow, I am a little bit surprised, not about the implementation which seems great but more about the mechanism that silently deletes a message that is in error state after refresh, not sure it gives a good user experience and it feels like we will have feedbacks from users not knowing why messages they sent are disappearing (because maybe they did not saw the error state and they switched to an other conversation, and maybe they want to be warned or have a trace of what happened)
> > @tprudentova can you please watch the different videos and confirm or not this behavior is acceptable from a product perspective ?
> 
> I only remove the event when bytes are null. Otherwise, I keep it unchanged @jbrnzl

I understand but it may still miss some user feedback
- When the page is refreshed and the message is deleted
and/or
- When the message is in error and we will probably not be able to retry sending it later

Or maybe the flow can be to leave the current errored message and if user retries and it's not possible because we miss some data, we tell and suggest him to delete this errored message ?

Maybe these are bad suggestions but to me it's weird that we delete some errored messages without letting the user know.

tprudentova:
> > > Oh wow, I am a little bit surprised, not about the implementation which seems great but more about the mechanism that silently deletes a message that is in error state after refresh, not sure it gives a good user experience and it feels like we will have feedbacks from users not knowing why messages they sent are disappearing (because maybe they did not saw the error state and they switched to an other conversation, and maybe they want to be warned or have a trace of what happened)
> > > @tprudentova can you please watch the different videos and confirm or not this behavior is acceptable from a product perspective ?
> > 
> > 
> > I only remove the event when bytes are null. Otherwise, I keep it unchanged @jbrnzl
> 
> I understand but it may still miss some user feedback
> 
> * When the page is refreshed and the message is deleted
>   and/or
> * When the message is in error and we will probably not be able to retry sending it later
> 
> Or maybe the flow can be to leave the current errored message and if user retries and it's not possible because we miss some data, we tell and suggest him to delete this errored message ?
> 
> Maybe these are bad suggestions but to me it's weird that we delete some errored messages without letting the user know.

I also have a feeling that deleting errored messages is not a good idea, even if we can't resend it... Maybe for now we can display them with a (!) status but without a retry button? If we can retry we display Retry, if we can't, we don't

nqhhdev:
@tprudentova If the user can't retry and display (!) only, we still have a question: why can't we retry again? Do we have other solutions?

nqhhdev:
https://github.com/linagora/twake-on-matrix/pull/3219

can we merge this PR first to close 2 issues?

cc: @9clg6 @jbrnzl @tprudentova

tprudentova:
> @tprudentova If the user can't retry and display (!) only, we still have a question: why can't we retry again? Do we have other solutions?

Another solution is to display some sort of error pop-up when user tries to resend but we can't resend. Like 'This message can't be resent because its data is unavailable'

nqhhdev:
@tprudentova Could you help me discuss with design team to create a new one?

tprudentova:
> @tprudentova Could you help me discuss with design team to create a new one?

Maybe we can use the error message component that we have rn? Just with different text 

This one I mean 
image

jbrnzl:
> > @tprudentova Could you help me discuss with design team to create a new one?
> 
> Maybe we can use the error message component that we have rn? Just with different text
> 
> This one I mean image

Any updates ?

tprudentova:
Here is the component that we can use 
https://www.figma.com/design/Cu4NB44bcURKEy40DEfwWi/Design-system-1.0?node-id=58560-43982&p=f&t=zm2UERL9S2l5rN7c-0
With text 'This message can't be resent because its data is unavailable' 
image
Maybe while we're at it we can replace this toast message with a component everywhere?

jbrnzl:
> Maybe while we're at it we can replace this toast message with a component everywhere?

Thank you for the update ! We already re-use the same snackbar in twake-on-matrix almost everywhere, it just need to be moved to the DS. We will do it in a follow-up ticket to not overload this PR now :)

jbrnzl:
Can you please check @9clg6 @nqhhdev I updated this PR with the modifications asked by @tprudentova

jbrnzl (approved):

jbrnzl (approved):

PR #3229, opened 2026-07-07
Fix sync error header visibility and Sentry reporting

### LINKED ISSUE #3201: [BUG] Waiting for response

### Describe the bug

On twake chat, lot of time, I see this "Waiting for Response" header meaning that the app considers myself as offline. But I'm not offline, I've internet connectivity. 

Generally, I need to kill and relaunch the app to be back online. 

I didn't find a way to reproduce it 100% of the time :( 

Image

### Pre-Bug State

.

### To Reproduce

.

### Expected results

.

### Context

iOS 26 / iPhone 17. 

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

9clg6:
## Summary

Fixes #3201. Replaces #3206.

- Hide the connection status header after a first successful sync, even when the Matrix SDK emits `SyncStatus.error`.
- Report Matrix sync errors to Sentry with lightweight context and throttling.
- Redact sync error payloads before sending them to Sentry.

## Validation

- `flutter analyze lib/widgets/connection_status_header.dart lib/widgets/matrix.dart`
- GitHub checks from previous PR were green.

## Note

This keeps non-fatal sync errors quiet in the UI once data already exists, while preserving technical signal in Sentry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection status visibility so the header updates consistently when sync data is available.
  * Corrected error messaging so sync errors display the proper localized “something went wrong” text.
  * Fixed sync error handling by throttling repeated error reports for the same client/error type to reduce duplicate alerts and improve reliability.
  * Enhanced error reporting context to include key sync details for more actionable diagnostics.

### DIFF
--- a/lib/widgets/connection_status_header.dart
+++ b/lib/widgets/connection_status_header.dart
@@ -35,10 +35,7 @@ class ConnectionStatusHeaderState extends State<ConnectionStatusHeader> {
       if (!mounted) return;
 
       final client = Matrix.of(context).client;
-      final hide =
-          client.onSync.value != null &&
-          update.status != SyncStatus.error &&
-          client.prevBatch != null;
+      final hide = client.onSync.value != null && client.prevBatch != null;
 
       if (_lastStatus?.status != update.status || _lastHideState != hide) {
         setState(() {
@@ -63,10 +60,7 @@ class ConnectionStatusHeaderState extends State<ConnectionStatusHeader> {
         _lastStatus ??
         client.onSyncStatus.value ??
         const SyncStatusUpdate(SyncStatus.waitingForResponse);
-    final hide =
-        client.onSync.value != null &&
-        status.status != SyncStatus.error &&
-        client.prevBatch != null;
+    final hide = client.onSync.value != null && client.prevBatch != null;
 
     return AnimatedSwitcher(
       duration: const Duration(milliseconds: 200),
--- a/lib/widgets/connection_status_header.dart
+++ b/lib/widgets/connection_status_header.dart
@@ -89,7 +89,7 @@ extension on SyncStatusUpdate {
       case SyncStatus.waitingForResponse:
         return l10n!.waitingForResponse;
       case SyncStatus.error:
-        return error?.exception?.toString() ?? l10n!.oopsSomethingWentWrong;
+        return l10n!.oopsSomethingWentWrong;
       case SyncStatus.processing:
       case SyncStatus.cleaningUp:
       case SyncStatus.finished:
--- a/lib/widgets/connection_status_header.dart
+++ b/lib/widgets/connection_status_header.dart
@@ -89,7 +89,7 @@ extension on SyncStatusUpdate {
       case SyncStatus.waitingForResponse:
         return l10n!.waitingForResponse;
       case SyncStatus.error:
-        return l10n!.oopsSomethingWentWrong;
+        return error?.exception?.toString() ?? l10n!.oopsSomethingWentWrong;
       case SyncStatus.processing:
       case SyncStatus.cleaningUp:
       case SyncStatus.finished:
--- a/lib/widgets/connection_status_header.dart
+++ b/lib/widgets/connection_status_header.dart
@@ -89,7 +89,7 @@ extension on SyncStatusUpdate {
       case SyncStatus.waitingForResponse:
         return l10n!.waitingForResponse;
       case SyncStatus.error:
-        return l10n!.waitingForResponse;
+        return l10n!.oopsSomethingWentWrong;
       case SyncStatus.processing:
       case SyncStatus.cleaningUp:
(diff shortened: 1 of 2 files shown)


9clg6:
> "Hello @9clg6 ! Nice PR ! I just have one concern: sometimes I have this "Waiting for response" text, and I am not able to send new messages (they are stuck on sending state in the rooms). When I force close and reopen everything is instantly back as normal and it's not a connectivity issue on my side so maybe it can be useful to keep the "Waiting for response" wording so that user know something is wrong (and try to do things like restarting the app or checking for his connectivity) and keep track with the sentry log you added ?" @jbrnzl 

Thanks for the feedback. The goal here is to avoid showing "Waiting for response" for transient sync errors after the app already has synced data, because in that case the banner can be misleading.

For the case you mention, where messages stay stuck until the app is restarted, the Sentry reporting should help us capture the related sync errors with context and investigate the real root cause. If the client is truly blocked, we should fix that separately rather than keeping a generic banner that may also appear for non-actionable errors.

jbrnzl:
> > "Hello @9clg6 ! Nice PR ! I just have one concern: sometimes I have this "Waiting for response" text, and I am not able to send new messages (they are stuck on sending state in the rooms). When I force close and reopen everything is instantly back as normal and it's not a connectivity issue on my side so maybe it can be useful to keep the "Waiting for response" wording so that user know something is wrong (and try to do things like restarting the app or checking for his connectivity) and keep track with the sentry log you added ?" @jbrnzl
> 
> Thanks for the feedback. The goal here is to avoid showing "Waiting for response" for transient sync errors after the app already has synced data, because in that case the banner can be misleading.
>
> For the case you mention, where messages stay stuck until the app is restarted, the Sentry reporting should help us capture the related sync errors with context and investigate the real root cause. If the client is truly blocked, we should fix that separately rather than keeping a generic banner that may also appear for non-actionable errors.

I agree on everything we said, but the thing is that even if we investigate future sentry logs about why the client is "blocked" until this time I think it is better to have a misleading generic error wording to at least give user a chance understand that something is wrong (until we have better). Currently if we try to send a message and it fails we do not have a specific error message too, just the retry button, which will not work if client is "blocked", so it can be misleading too in my opinion because what's happening is at an higher level, not only about sending messages. I am not against these changes at all, this is just another point of view. And I'm probably nitpicking 😅

Crash--:
> I think it is better to have a misleading generic error wording to at least give user a chance understand that something is wrong (until we have better)

+1

9clg6:
> > > "Hello @9clg6 ! Nice PR ! I just have one concern: sometimes I have this "Waiting for response" text, and I am not able to send new messages (they are stuck on sending state in the rooms). When I force close and reopen everything is instantly back as normal and it's not a connectivity issue on my side so maybe it can be useful to keep the "Waiting for response" wording so that user know something is wrong (and try to do things like restarting the app or checking for his connectivity) and keep track with the sentry log you added ?" @jbrnzl
> > 
> > 
> > Thanks for the feedback. The goal here is to avoid showing "Waiting for response" for transient sync errors after the app already has synced data, because in that case the banner can be misleading.
> > For the case you mention, where messages stay stuck until the app is restarted, the Sentry reporting should help us capture the related sync errors with context and investigate the real root cause. If the client is truly blocked, we should fix that separately rather than keeping a generic banner that may also appear for non-actionable errors.
> 
> I agree on everything we said, but the thing is that even if we investigate future sentry logs about why the client is "blocked" until this time I think it is better to have a misleading generic error wording to at least give user a chance understand that something is wrong (until we have better). Currently if we try to send a message and it fails we do not have a specific error message too, just the retry button, which will not work if client is "blocked", so it can be misleading too in my opinion because what's happening is at an higher level, not only about sending messages. I am not against these changes at all, this is just another point of view. And I'm probably nitpicking 😅

I agree but the problem is that we don't have enough place to put the error so I placed "Oops, something went wrong…" and we will ask to users who encounter error to show us their logs

lib/widgets/connection_status_header.dart
```diff
@@ -95,7 +89,7 @@ extension on SyncStatusUpdate {
       case SyncStatus.waitingForResponse:
         return l10n!.waitingForResponse;
       case SyncStatus.error:
-        return l10n!.waitingForResponse;
+        return error?.exception?.toString() ?? l10n!.oopsSomethingWentWrong;
```
jbrnzl: Can we have only l10n!.oopsSomethingWentWrong here ? @9clg6

9clg6: Done

jbrnzl (approved):

nqhhdev (approved):
LGTM

PR #3233, opened 2026-07-07
refactor: move contacts invitation to Riverpod view model

9clg6:
## Summary
- Move the contacts invitation presentation logic into a Riverpod view model and state.
- Keep the existing screen behavior while isolating state handling from the widget.
- Add focused tests for the invitation providers and view model.

## Validation
- flutter test test/pages/contacts_tab/contacts_invitation_view_model_test.dart test/pages/contacts_tab/providers/contacts_invitation_providers_test.dart
- flutter analyze lib/pages/contacts_tab/contacts_invitation.dart lib/pages/contacts_tab/contacts_invitation_view.dart lib/pages/contacts_tab/contacts_invitation_view_model.dart lib/pages/contacts_tab/providers/contacts_invitation_providers.dart
- Ran on iOS device Mangue and validated startup/navigation for this branch.

https://github.com/user-attachments/assets/7487cc44-c388-41e2-bcef-6b1e44010825

## Summary by CodeRabbit

* **New Features**
  * Added a new Riverpod-based contact invitation experience for selecting contacts, sending invitations, and sharing invitation links.
  * Added automatic default selection of the first available phone (or first email) and improved selection behavior (including locking rules).
  * Added a dedicated page UI for link sharing and invitation sending with proper loading/success/failure rendering.

* **Bug Fixes**
  * Improved invitation error diagnostics using richer Dio exception details while keeping response contents safely summarized/redacted.

* **Tests**
  * Added unit tests for the invitation view model, Riverpod providers, and safe response-data summary behavior.


lib/pages/contacts_tab/providers/contacts_invitation_providers.dart
nqhhdev, line 1: This file just wrap GetIt for 3 interactors with no added logic. Since these are already GetIt singletons, this extra Riverpod layer doesn't buy us anything. Do we need this provider layer at all, or can the view provider call GetIt directly?

pubspec.yaml
nqhhdev: I don't see you using this lib for others, why did we need it?

lib/pages/contacts_tab/contacts_invitation_state.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:fluffychat/app_state/failure.dart';
+import 'package:fluffychat/app_state/success.dart';
+import 'package:fluffychat/domain/app_state/invitation/generate_invitation_link_state.dart';
+import 'package:fluffychat/domain/app_state/invitation/send_invitation_state.dart';
+import 'package:fluffychat/presentation/model/contact/presentation_contact.dart';
+
+class ContactsInvitationState extends Equatable {
+  final PresentationThirdPartyContact? selectedContact;
+  final Either<Failure, Success> sendInvitationState;
+  final Either<Failure, Success> generateInvitationLinkState;
```
nqhhdev: This stores Either, which duplicates state management. Should we drop Either here and either split into explicit (ex: loading, success, failed) state per action. Wdyt

nqhhdev:
Also add demo

lib/pages/contacts_tab/providers/contacts_invitation_providers.g.dart
jbrnzl, line 1: Generated files need to be included ?

lib/domain/usecase/invitation/generate_invitation_link_interactor.dart
```diff
@@ -32,7 +32,18 @@ class GenerateInvitationLinkInteractor {
...
       );
     } catch (e) {
-      Logs().e('GenerateInvitationLinkInteractor::execute', e);
+      if (e is DioException) {
+        Logs().e(
+          'GenerateInvitationLinkInteractor::execute DioException '
+          'status=${e.response?.statusCode} '
+          'type=${e.type} '
+          'path=${e.requestOptions.path} '
+          'data=${e.response?.data}',
```
jbrnzl: Not sure but should we use _safeResponseDataSummary here too like [here](https://github.com/linagora/twake-on-matrix/pull/3233/changes#diff-5da5dc954a6c3747c8627ef3d0690c11627a6c948a6256a4d4b961bffd6cfba4R44) ?

lib/pages/contacts_tab/providers/contacts_invitation_providers.dart
9clg6: Yes, this provider layer is intentional. During the GetIt-to-Riverpod migration, it gives the UI a consistent Riverpod access pattern and prevents new view code from depending directly on GetIt.
It is a temporary adapter: dependencies still come from GetIt today, but views can already use Riverpod providers and tests can override them. This keeps the migration incremental and makes removing GetIt later localized, without changing every consumer.

pubspec.yaml
9clg6: You are right I added it in last to migrate the view state to it, and I forget to do it. 
I modified `ContactsInvitationState`

lib/pages/contacts_tab/contacts_invitation_state.dart
9clg6: Agreed.
 Either is part of the legacy interactor contract and should not leak into Riverpod presentation state.

I will replace these fields with action-specific AsyncValues and keep the Either mapping contained in the ViewModel as a temporary migration adapter. When the interactors move from Stream<Either<...>> to Future<T> with typed exceptions, this adapter can be removed and the controller can use AsyncValue.guard() directly.

I was afraid that Either pattern removal could be a too big change

lib/domain/usecase/invitation/generate_invitation_link_interactor.dart
9clg6: Good catch. Replaced raw `response.data` logging with a shared `safeResponseDataSummary` helper used by both invitation interactors. It retains only map keys or runtime type, never response values. Added unit tests.

lib/pages/contacts_tab/contacts_invitation.dart
```diff
@@ -31,220 +27,172 @@ class ContactsInvitation extends StatefulWidget {
...
+      data: (link) async {
+        TwakeDialog.hideLoadingDialog(context);
+        if (link == null) {
           return;
         }
+        await Share.shareUri(link);
+        Navigator.of(context).pop();
       },
     );
   }
```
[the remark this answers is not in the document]
9clg6: Addressed in fd4632a. The `AsyncValue.data` handler is now synchronous; sharing runs in a dedicated method with error handling and a `mounted` check before navigation. Invitation-link-specific fallback messages are used.

lib/pages/contacts_tab/contacts_invitation.dart
```diff
@@ -31,220 +27,172 @@ class ContactsInvitation extends StatefulWidget {
...
           invitationId: invitationId,
-        )
-        .listen((state) {
-          Logs().d(
-            'ContactsInvitationController::_onStoreInvitationStatusInteractor',
-            state,
-          );
-        });
+        );
   }
```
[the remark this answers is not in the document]
9clg6: Addressed in fd4632a. `storeInvitationStatus` now returns whether the stream reached `StoreInvitationStatusSuccessState`, logs failure states, and the UI awaits that result before navigation. The user still receives invitation success feedback when server invitation succeeded; local persistence failures are explicit in logs. Added success and failure tests.

lib/domain/usecase/invitation/safe_response_data_summary.dart
```diff
@@ -0,0 +1,9 @@
+String safeResponseDataSummary(Object? data) {
+  if (data == null) {
+    return 'null';
+  }
+  if (data is Map) {
+    return 'Map(keys=${data.keys.join(',')})';
+  }
+  return data.runtimeType.toString();
+}
```
nqhhdev, line 9: Why didn't you create a class? 

`String safeResponseDataSummary(Object? data)` => wrong format

lib/pages/contacts_tab/providers/contacts_invitation_providers.dart
nqhhdev: IMO, this layer isn’t necessary for now. We can remove GetIt once we fully migrate to Riverpod.

lib/domain/usecase/invitation/safe_response_data_summary.dart
9clg6: It’s a stateless helper, so a top-level function seemed sufficient. We already use this pattern for similar helpers in the codebase. Is your concern specifically about placing it under domain/usecase? I can move it to a logging utility if that better matches the project structure.

lib/pages/contacts_tab/providers/contacts_invitation_providers.dart
9clg6: I see your point, but I’d keep this adapter. The view model tests already override these providers directly, and calling GetIt from the view model would couple the new Riverpod code to the legacy locator. With this layer, removing GetIt later stays localized to these three provider implementations.

lib/pages/contacts_tab/contacts_invitation_view.dart
```diff
@@ -88,14 +94,14 @@ class ContactsInvitationView extends StatelessWidget {
...
+                                builder: (context) {
+                                  final selectedContact = state.selectedContact;
                                   final isSelectedContact =
-                                      contact is PresentationPhoneNumber &&
-                                      contact.phoneNumber ==
+                                      selectedContact
+                                          is PresentationPhoneNumber &&
+                                      selectedContact.phoneNumber ==
                                           phoneNumber.phoneNumber;
                                   return Padding(
```
jbrnzl: Pre-existing and not blocking, you already improved it a lot, but I think it would have been great to extract a shared widget here: from what I see phone and mail cases are very similar

lib/pages/contacts_tab/contacts_invitation_view.dart
```diff
@@ -1,19 +1,25 @@
...
-import 'package:fluffychat/pages/contacts_tab/contacts_invitation.dart';
+import 'package:fluffychat/pages/contacts_tab/contacts_invitation_state.dart';
 import 'package:fluffychat/pages/contacts_tab/contacts_invitation_style.dart';
 import 'package:fluffychat/presentation/model/contact/presentation_contact.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:linagora_design_flutter/linagora_design_flutter.dart';
 import 'package:fluffychat/generated/l10n/app_localizations.dart';
 
 class ContactsInvitationView extends StatelessWidget {
```
jbrnzl: Should this be renamed to "ContactsInvitationPage" if "Page" suffix is the new convention for screens ?

lib/pages/contacts_tab/contacts_invitation_view.dart
nqhhdev: Totally agree

lib/pages/contacts_tab/contacts_invitation_view.dart
nqhhdev: I don't think so, we can keep it.

9clg6: Not blocking but yes, it will make the syntax more consistent

jbrnzl (approved):

lib/pages/contacts_tab/contacts_invitation_page.dart
```diff
@@ -0,0 +1,298 @@
...
+                                children: [
+                                  SizedBox(
+                                    width: 28,
+                                    height: 28,
+                                    child: CircularProgressIndicator.adaptive(
+                                      strokeWidth: 2,
+                                      backgroundColor:
+                                          LinagoraSysColors.material()
+                                              .onPrimary,
+                                    ),
```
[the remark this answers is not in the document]
9clg6: Thanks, the Material diagnosis is correct. However, `CircularProgressIndicator.adaptive` ignores `valueColor` on iOS/macOS and maps `backgroundColor` to the Cupertino indicator color, so applying the suggestion as-is would regress the Apple-platform tint. Fixed in f2dd28957 by using a non-adaptive `CircularProgressIndicator` with `color: onPrimary`, which keeps the intended contrast consistent across platforms. Targeted analysis and the invitation ViewModel/provider tests pass.

lib/pages/contacts_tab/contacts_invitation_view.dart
9clg6: Good point — done in 5efe913: phone and email now reuse `_InvitationContactItem`.

9clg6: Done

nqhhdev (approved):
LGTM

PR #3234, opened 2026-07-07
feat: add contacts pull to refresh

9clg6:
## Summary
- Add pull-to-refresh on the Contacts tab to retry contact synchronization.
- Keep the contacts list scrollable while empty or lightly populated so refresh remains available.
- Add a retry action on the empty contacts state.

https://github.com/user-attachments/assets/c4f4002c-b4da-4438-b4b4-cf570f17b346

https://github.com/user-attachments/assets/5a3c849e-24de-4024-8108-1016ed8e04d1

image

## Validation
- Tested on Mangue iOS.
- Verified Contacts tab loads and pull-to-refresh triggers a resync without blocking the list.
- Verified previous local contacts and invitation behavior from the base branch still works.

## Summary by CodeRabbit

* **New Features**
  * Added pull-to-refresh to the Contacts tab.
  * Added a “tap to retry” sync button in the empty contacts state when applicable.
* **Bug Fixes**
  * Improved loading/empty/search UI so the correct feedback is shown consistently.
  * Contact sync retry now safely handles errors without breaking the flow.
* **Tests**
  * Added widget coverage for empty search state while contacts are loading.
  * Added coverage for retry sync handling when subscription cancellation errors occur.


lib/pages/contacts_tab/empty_contacts_body.dart
```diff
@@ -34,6 +36,15 @@ class EmptyContactBody extends StatelessWidget {
             ),
             textAlign: TextAlign.center,
           ),
+          if (onRetrySyncContacts != null) ...[
+            const SizedBox(height: 16.0),
+            FilledButton.icon(
```
jbrnzl: Can we reuse component from design system here ? Or create/update an existing one to reuse here ?

9clg6: Yes, thanks

jbrnzl:
@9clg6 I tested and I have an infinite loading spin when I am looking for a contact that does not exist (I expect to have the empty list state instead)

https://github.com/user-attachments/assets/a67c502c-88c7-45c3-94e9-2dddd5918a17

9clg6:
> @9clg6 I tested and I have an infinite loading spin when I am looking for a contact that does not exist (I expect to have the empty list state instead)
> 
>  Enregistrement.de.l.ecran.2026-07-16.a.15.34.16.mov

Fixed, thanks for your test

jbrnzl:
@9clg6 please ping me when rebased I will review 🙏🏻

jbrnzl (dismissed):

nqhhdev (dismissed):
Tested, LGTM

nqhhdev:
pls resolve conflicts

lib/pages/contacts_tab/contacts_tab.dart
```diff
@@ -142,6 +142,18 @@ class ContactsTabController extends State<ContactsTab>
...
+    if (!mounted) {
+      return;
+    }
+
+    await retrySynchronizeContactsOnContactTab(
+      context: context,
+      client: Matrix.of(context).client,
+      matrixLocalizations: MatrixLocals(L10n.of(context)!),
+    );
+  }
```
nqhhdev, line 155: Do we need to try/catch to handle the error in this case?

test/pages/contacts_tab/widgets/contacts_empty_search_state_test.dart
nqhhdev, line 1: - This test is tautological — isLoadingContacts/hasVisibleContacts are stubbed directly on the mock, and the widgets under test only read those same getters. So the assertion just echoes back the stub, never exercises the real logic in ContactsViewControllerMixin. Invert the real getters and this test still passes unchanged. 

- isWaitingContacts is also stubbed but unused — neither widget reads it.

IMO: drive state through the real notifiers instead of stubbing the computed getters, so the assertion actually exercises real logic. WDYT

lib/pages/contacts_tab/contacts_tab.dart
9clg6: Handled in 5e6a05e8f: retry setup errors are now caught and logged, with a regression test.

test/pages/contacts_tab/widgets/contacts_empty_search_state_test.dart
9clg6: Updated in 5e6a05e8f: the widget test now drives the real mixin notifiers/getters instead of stubbing computed values.

jbrnzl (approved):

nqhhdev (approved):
LGTM

PR #3235, opened 2026-07-07
fix: load local contacts from homeserver well-known

9clg6:
## Summary
- Gate local contact display with the homeserver well-known invitation flag.
- Keep local contact lookup running independently from display visibility.
- Retry contact synchronization after well-known data becomes available.
- Avoid logging missing cached invitation status as an error because it is an expected state for contacts without stored invitations.

## Validation
- Tested on Mangue iOS.
- Verified the app starts past splash.
- Verified local contacts can be displayed when `enable_invitations` is enabled.
- Verified `InvitationStatusNotFound` no longer spams error logs.

On SaaS : 

https://github.com/user-attachments/assets/73ebae76-83da-40dd-9e05-7122a6bd8a6f

On LNG (invitations disabled) : 
IMG_6671

## Known follow-up
- Phonebook identity lookup can still fail with `401 M_SESSION_NOT_VALIDATED`; this is a separate token/session issue.

## Summary by CodeRabbit

* **New Features**
  * Improved well-known discovery using the current homeserver (with fallback) and refreshed invitation availability across contacts flows.
  * Added “keep alive” for the contacts tab and auto-retry when discovery data is missing.
  * Enabled phonebook results to participate in loading/search when available, with updated permission handling.

* **Bug Fixes**
  * Phonebook hash-details failures now retain discovered contacts for smoother recovery.
  * Avoided profile loading and tap actions for contacts with missing/empty Matrix IDs.
  * Reduced noisy error logging for expected “not found” cases.
  * Improved phonebook synchronization timing to prevent stuck syncing.

* **Tests**
  * Updated and added unit tests for the new failure payloads and sync retry behavior.

### DIFF
--- a/lib/presentation/mixins/wellknown_mixin.dart
+++ b/lib/presentation/mixins/wellknown_mixin.dart
@@ -1,4 +1,9 @@
+import 'dart:convert';
+
+import 'package:fluffychat/utils/custom_http_client.dart';
+import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:flutter/material.dart';
+import 'package:http/http.dart' as http;
 import 'package:matrix/matrix.dart';
 
 mixin WellKnownMixin {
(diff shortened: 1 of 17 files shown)


nqhhdev:
Pls add demo for both the web and mobile platform.

pm-McFly:
@9clg6 @tprudentova IMO having a visual queue indicating the synchronization is in progress and error state might be useful, wdyt?

tprudentova:
> @9clg6 @tprudentova IMO having a visual queue indicating the synchronization is in progress and error state might be useful, wdyt?

We used to have the sync progress bar previously, but it was a direct request from both Diana and the design team to remove it, cause it didn't really give users much info, but took a huge part of the screen :( So I'm not sure it's a good idea to add it back, but I'll check with them one more time

9clg6:
Any news @tprudentova ?

lib/presentation/mixins/wellknown_mixin.dart
```diff
@@ -10,18 +15,56 @@ mixin WellKnownMixin {
...
+    if (homeserver == null) {
+      return;
+    }
+    final result = await discoverFromHomeserver(homeserver);
+    if (result == null) return;
+    Logs().d('WellKnownMixin::getWellKnownInformation() well-known $result');
+    discoveryInformationNotifier.value = result;
+  }
+
+  static Future<DiscoveryInformation?> discoverFromHomeserver(
```
jbrnzl: @9clg6 Could we re-use `getWellKnownOrFallback` from #3251 in this PR ? I feel like reusing it will make the changes here more simple, if reusable in this case.

Rest of the PR looks good to me !

9clg6: Done in 6727dc674: `WellKnownMixin.getWellKnownInformation` now reuses `Client.getWellKnownOrFallback`, passing the current discovery information as fallback. I also removed the duplicated HTTP implementation and its now-redundant test; the mixin and extension tests pass locally.

jbrnzl (approved):

nqhhdev (approved):
LGTM

Issue #3223, opened 2026-07-06
[BUG] When I jump to a notification to my second account, I still see the 1st account avatar in the bottom bar


tprudentova:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

https://github.com/user-attachments/assets/a34c1850-8c97-434d-9800-64c7474a57ee

### Pre-Bug State

.

### To Reproduce

1. Log into two accounts in the mobile app
2. Go to account 1 
3. Receive a notification to account 2 
4. Jump to that notification 
5. You're in account 2, but the avatar at the bottom bar is still of account 1

### Expected results

When you switch accounts, the avatar in the bottom bar switches as well

### Context

- Device: Redmi Note 12 Pro
- OS: Androis
- Browser: [e.g. chrome, safari]
- App Version: 2.23.0
- Server: staging, linagora
- Account information:
- Others

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

PR #3236, opened 2026-07-08
TW-3223: Fix wrong bottom bar avatar after switching account from notification

nqhhdev:
## Ticket
**Related issue**
- #3223

## Resolved
- Web:
- Android:
- IOS:

https://github.com/user-attachments/assets/b59f52ed-e017-418e-8370-53dcfaf856f2

## Summary by CodeRabbit

* **New Features**
  * Improved account switching when opening notifications by carrying the newly activated account into the destination screen.
  * Destination navigation now refreshes using the correct active account after iOS notification redirects.
* **Bug Fixes**
  * Fixed cases where notification-based navigation could drop or ignore the newly selected account context during room/redirect handling.

### DIFF
--- a/lib/utils/background_push.dart
+++ b/lib/utils/background_push.dart
@@ -689,19 +689,20 @@ class BackgroundPush {
       if (_matrixState == null || roomId == null) {
         return;
       }
-      await client.roomsLoading;
-      await client.accountDataLoading;
+      final activeClient = switchedClient ?? client;
+      await activeClient.roomsLoading;
+      await activeClient.accountDataLoading;
       // ignore: unused_local_variable
       final isStory =
-          client
+          activeClient
               .getRoomById(roomId)
               ?.getState(EventTypes.RoomCreate)
               ?.content
               .tryGet<String>('type') ==
           ClientStoriesExtension.storiesRoomType;
-      if (client.getRoomById(roomId) == null) {
+      if (activeClient.getRoomById(roomId) == null) {
         Logs().v('[Push] Room $roomId not found, syncing...');
-        await client.waitForRoomInSync(roomId);
+        await activeClient.waitForRoomInSync(roomId);
       }
       _handleRedirectRoom(
         roomId,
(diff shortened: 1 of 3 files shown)


lib/utils/background_push.dart
```diff
@@ -932,11 +953,18 @@ class BackgroundPush {
...
+    final extra = switchedClient != null
+        ? SwitchActiveAccountBodyArgs(newActiveClient: switchedClient)
+        : null;
     if (eventId != null) {
-      TwakeApp.router.go('/rooms/$roomId?event=$eventId');
+      TwakeApp.router.go('/rooms/$roomId?event=$eventId', extra: extra);
       return;
     }
-    TwakeApp.router.go('/rooms/$roomId');
+    TwakeApp.router.go('/rooms/$roomId', extra: extra);
```
9clg6, line 969: We added 2 months ago (refacto: migrate routing to go_router_builder for type-safe navigation ([#2901]) `go_router_builder` could we use e.g `const RoomsRoute().go(context)` please ? Or it's not possible in this case ?

nqhhdev: Thank you for your concern, Not possible for this case — RoomRoute's $extra only accepts ChatRouterInputArgument, so SwitchActiveAccountBodyArgs would be silently dropped. That arg is meant for the shell route, which reads extra directly from the main router, so we need TwakeApp.router.go instead of the typed leaf route.

jbrnzl:
@nqhhdev I tested but unfortunately when I receive a notification for a new message and I click on it I land on the wrong account but it should have switched me on the account where I received this notification instead, not the one where I sent it from) so something is wrong, and the bottom bar avatar is wrong too

https://github.com/user-attachments/assets/3856be25-aa15-4750-9559-8bb3ed6623cd

lib/utils/background_push.dart
jbrnzl, line 649: @nqhhdev Maybe the wrong account redirect issue I have could come from here ? I mean maybe here the client should also be resolved from the payload no ? Anyway I think it's not blocking and I created a separate issue #3300

jbrnzl (approved):

9clg6 (approved):

Issue #3243, opened 2026-07-08
Refactor scroll to event mechanism


jbrnzl:
While investigating for #3222 I've seen that some part of the code can be simplified easily (for example duplicated code in chat.dart regarding reloading and centering) 

I'll come with more examples but I wanted to open a seperate issue, not include everything in the PR that is fixing something.

PR #3244, opened 2026-07-08
TW-3222(fix): recover jump to message on kept alive off-screen rows

### LINKED ISSUE #3222: [BUG] Can't jump to a source message of a reply (sometime issue)

### Describe the bug

If I reply to a message that includes a link, or any text, I can't jump to the source message

https://github.com/user-attachments/assets/6aae454b-144d-4b78-a95e-c5aa8815596c

### Pre-Bug State

N/A

### To Reproduce

1. Send a message 
2. Reply to that message
3. Try clicking on the source messsage -> can't jump to it 

It's a sometime issue, so you have to test multiple time to reproduce, source message should be rendered but without position in the view

### Expected results

User can jump to any source message of a reply

### Context

- Device: Honor
- OS: Windows
- Browser: Chrome
- App Version: 2.23.0
- Server: staging
- Account information:
- Others

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
## Ticket
Fixes #3222

## Root cause
Tapping a reply quote sometimes don't jump to the source message (with an `Unsupported operation: NaN` in console). This is happening because the jump path treats `GlobalObjectKey(eventId).currentContext != null` as "the row is rendered". But in a lazy sliver a row can be mounted but kept alive off the screen, and in this case `localToGlobal` returns `NaN` which absorb the jump. This is why this issue is not happening every time. (only when the target is kept-alive but off the screen)

## Solution
- Add hasFiniteViewportPosition that verify row is mounted and positioned on the screen.
- Gate the jump paths with it
- Guard centerRenderedMessage against a non-finite offset as a crash net

## Test recommendations
Please help me test by trying to jump to old or recent messages, with or without link.
If you have an issue, please compare with main to know if it's a regression related to this PR or a pre-existing bug.

### Before

https://github.com/user-attachments/assets/e1fbf733-f690-4439-b001-04e105b0ee87

### After

https://github.com/user-attachments/assets/c6f1d1b9-da1a-41fc-a6a4-f918f3d14f99

## More info

While looking at the issue I think some part of the pre-existing code can be improved so I created a follow up issue here #3243  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat message navigation so jumping to a message no longer tries to center or highlight items that are not yet usable on screen.
  * Made message scrolling more reliable by ignoring non-finite positions, reducing cases where “scroll to message” could fail or behave unexpectedly.
  * Updated message availability checks so the app waits for a message to be truly ready before attempting to focus it.

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -79,6 +79,7 @@ import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:fluffychat/utils/responsive/responsive_utils.dart';
 import 'package:fluffychat/utils/room_draft_storage.dart';
 import 'package:fluffychat/utils/room_status_extension.dart';
+import 'package:fluffychat/utils/scroll_controller_extension.dart';
 import 'package:fluffychat/utils/twake_snackbar.dart';
 import 'package:fluffychat/widgets/context_menu/context_menu_action.dart';
 import 'package:fluffychat/widgets/context_menu/context_menu_action_item_widget.dart';
@@ -1546,7 +1547,8 @@ class ChatController extends State<Chat>
     final targetEventId = timeline!.events[targetIndex].eventId;
     final itemContext = GlobalObjectKey(targetEventId).currentContext;
 
-    if (itemContext != null) {
+    if (itemContext != null &&
+        scrollController.hasFiniteViewportPosition(itemContext)) {
       await _centerAndHighlightMessage(
         itemContext: itemContext,
         targetIndex: targetIndex,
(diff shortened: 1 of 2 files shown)


nqhhdev:
I tested it on both mobile and web, it works well.

jbrnzl:
> I tested it on both mobile and web, it works well.

Thank you for testing !! 🙏🏻

nqhhdev (approved):
LGTM

9clg6 (approved):

PR #3249, opened 2026-07-09
Translate web push browser settings

9clg6:
## Summary
- Add Web Push browser settings translations for the locales available in the app language settings: French, Russian, and Vietnamese.
- Keep other locales falling back to English until they are supported in settings.

## Root cause
The new Web Push settings strings only existed in the English ARB file, so supported non-English settings locales fell back to English.

## Validation
- `flutter gen-l10n`
- `dart format lib/generated/l10n`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added new localized text for web push notifications in French, Russian, and Vietnamese.
  * Improved user-facing messages for enabling browser notifications and for cases where notification permission is denied or blocked.


nqhhdev:
I forced and rebased with l10n on weblate, so we can move on this ticket #3252

PR #3251, opened 2026-07-09
fix: keep well-known discovery across restarts by re-fetching from th…

jbrnzl:
## Root cause
The well-known is only fetched during login and kept in memory: any restart loses discovery informations, so well-known driven features (livekit call button for example) disappear until logout and login again, and well-known changes are not refreshed.

## Solution
Refetch the well-known on startup from the homeserver host and persist it through the SDK well_known cache served as offline fallback

## Summary by CodeRabbit

* **New Features**
  * Improved homeserver discovery using cached well-known information.
  * Added fallback handling when discovery services are unavailable.
  * Persisted discovery data can now be reused across app restarts.

* **Bug Fixes**
  * Prevented outdated homeserver checks from overwriting current connection state.
  * Improved recovery when homeserver checks fail by using available cached information.

* **Tests**
  * Added coverage for discovery requests, caching, fallback behavior, persistence, and correct homeserver targeting.

### DIFF
--- a/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart
@@ -1,0 +1,34 @@
+import 'package:matrix/matrix.dart';
+// ignore: implementation_imports
+import 'package:matrix/src/utils/request_and_cache.dart';
+
+extension ClientWellKnownExtension on Client {
+  /// Fetches the discovery information (`/.well-known/matrix/client`) from
+  /// the homeserver host, returning [fallback] when it cannot be fetched.
+  Future<DiscoveryInformation?> getWellKnownOrFallback({
+    DiscoveryInformation? fallback,
+    Duration cacheLifetime = const Duration(minutes: 5),
+  }) async {
+    try {
+      return await requestAndCache<DiscoveryInformation>(
+        () => MatrixApi(
+          homeserver: homeserver,
+          httpClient: httpClient,
+        ).getWellknown(),
+        fromJson: DiscoveryInformation.fromJson,
+        toJson: (wellKnown) => wellKnown.toJson(),
+        cacheKey: 'well_known',
+        cacheLifetime: cacheLifetime,
+        throwOnUpdateFailure: false,
+      );
+    } catch (e) {
+      Logs().w(
+        'ClientWellKnownExtension::getWellKnownOrFallback: could not fetch '
+        'well-known from the homeserver host, keeping previous discovery '
+        'information',
+        e,
+      );
+      return fallback;
+    }
+  }
+}
--- a/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart
+++ b/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart
@@ -3,8 +3,11 @@ import 'package:matrix/matrix.dart';
 import 'package:matrix/src/utils/request_and_cache.dart';
 
 extension ClientWellKnownExtension on Client {
+  static const _cacheKey = 'well_known';
+
   /// Fetches the discovery information (`/.well-known/matrix/client`) from
   /// the homeserver host, returning [fallback] when it cannot be fetched.
+  /// A returned fallback is persisted so it survives the next restart.
   Future<DiscoveryInformation?> getWellKnownOrFallback({
     DiscoveryInformation? fallback,
     Duration cacheLifetime = const Duration(minutes: 5),
@@ -17,7 +20,7 @@ extension ClientWellKnownExtension on Client {
         ).getWellknown(),
         fromJson: DiscoveryInformation.fromJson,
         toJson: (wellKnown) => wellKnown.toJson(),
-        cacheKey: 'well_known',
+        cacheKey: _cacheKey,
         cacheLifetime: cacheLifetime,
         throwOnUpdateFailure: false,
       );
@@ -28,6 +31,18 @@ extension ClientWellKnownExtension on Client {
         'information',
         e,
       );
+      if (fallback != null && isLogged()) {
+        // Persist the fallback so it survives the next restart
+        try {
+          await database.cacheCustomObject(_cacheKey, fallback.toJson());
+        } catch (e) {
+          Logs().w(
+            'ClientWellKnownExtension::getWellKnownOrFallback: could not '
+            'persist the fallback well-known',
+            e,
+          );
+        }
+      }
       return fallback;
     }
   }
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -1196,6 +1196,10 @@ class MatrixState extends State<Matrix>
 
     if (checkedSummary == null && discovery == null) return;
 
+    // The active client may have changed while awaiting: a late result must
+    // not overwrite the newer account's summary.
+    if (clientOrNull != newClient) return;
+
     loginHomeserverSummary = HomeserverSummary(
       discoveryInformation: discovery,
       versions:
--- a/lib/widgets/matrix.dart
+++ b/lib/widgets/matrix.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 import 'dart:io';
 import 'package:fluffychat/config/go_routes/app_routes.dart';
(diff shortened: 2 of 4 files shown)


9clg6:
Good direction: fetching well-known from the homeserver avoids the wrong MXID-domain requests, and using the SDK cache improves persistence.

However, I see a few consistency issues before merging:

- On restart, `checkHomeserver()` runs before reading the cached well-known. If the server is fully unreachable, it throws and the persisted cache is never used.
   `_getHomeserverInformation()` calls `checkHomeserver()` before `getWellKnownOrFallback()`:
   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1179-L1185

   `checkHomeserver()` performs `/versions` and `/login`; on failure it sets `homeserver = null` and rethrows:
   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/client.dart#L563-L597

   Therefore, with a fully unreachable server, execution stops before the persisted well-known cache is read.

--------

- MatrixState keeps a single loginHomeserverSummary shared by all clients. During account switching, a failed refresh for account B may reuse account A’s discovery data.
   `loginHomeserverSummary` is one global field:
   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L121

   It is used as fallback for any `newClient`, without checking account or origin:
   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1178-L1191

   Scenario: account A is active, switch to B, B has no cache and its fetch fails. Discovery A is then assigned together with versions/login flows from B.

--------

- A fallback returned after a failed fetch is not persisted, so it can be lost on the next restart.
   `requestAndCache()` writes only after a successful network response. Without cache, failure is rethrown:
   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/utils/request_and_cache.dart#L18-L38

   The extension catches that error and returns `fallback`, but performs no cache write:
   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart#L24-L31

   Therefore, this fallback survives only in memory.

--------

- Versions/login flows are fetched from the current homeserver, while the refreshed well-known may point to another `m.homeserver.base_url`. This can create a mixed snapshot.
   Versions/login flows are fetched from current homeserver A. The following well-known request may return `m.homeserver.base_url = B`, but `newClient.homeserver` is never updated or B validated:
   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1179-L1191

   Result: Matrix traffic continues against A while UI/configuration reads discovery pointing to B.

--------

- requestAndCache() has no in-flight deduplication, so two concurrent cache misses start two requests. MatrixState also has no guard before assigning loginHomeserverSummary, so a late result may overwrite a newer one.
   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/utils/request_and_cache.dart#L18-L27

   Two concurrent cache misses therefore start two requests. Their completion order also determines the final published global summary.

The current “restart” test reuses the same `Client` and in-memory database, and only removes the well-known endpoint:
https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/test/utils/matrix_sdk_extensions/client_well_known_extension_test.dart#L129-L152

So it does not cover a cold restart, a reopened production database, a fully offline server, account switching, or concurrent refreshes.

wdyt?

jbrnzl:
> Good direction: fetching well-known from the homeserver avoids the wrong MXID-domain requests, and using the SDK cache improves persistence.
> 
> However, I see a few consistency issues before merging:
> 
> * On restart, `checkHomeserver()` runs before reading the cached well-known. If the server is fully unreachable, it throws and the persisted cache is never used.
>   `_getHomeserverInformation()` calls `checkHomeserver()` before `getWellKnownOrFallback()`:
>   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1179-L1185
>   
>   `checkHomeserver()` performs `/versions` and `/login`; on failure it sets `homeserver = null` and rethrows:
>   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/client.dart#L563-L597
>   Therefore, with a fully unreachable server, execution stops before the persisted well-known cache is read.
> 
> * MatrixState keeps a single loginHomeserverSummary shared by all clients. During account switching, a failed refresh for account B may reuse account A’s discovery data.
>   `loginHomeserverSummary` is one global field:
>   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L121
>   
>   It is used as fallback for any `newClient`, without checking account or origin:
>   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1178-L1191
>   
>   Scenario: account A is active, switch to B, B has no cache and its fetch fails. Discovery A is then assigned together with versions/login flows from B.
> 
> * A fallback returned after a failed fetch is not persisted, so it can be lost on the next restart.
>   `requestAndCache()` writes only after a successful network response. Without cache, failure is rethrown:
>   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/utils/request_and_cache.dart#L18-L38
>   The extension catches that error and returns `fallback`, but performs no cache write:
>   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/utils/matrix_sdk_extensions/client_well_known_extension.dart#L24-L31
>   
>   Therefore, this fallback survives only in memory.
> 
> * Versions/login flows are fetched from the current homeserver, while the refreshed well-known may point to another `m.homeserver.base_url`. This can create a mixed snapshot.
>   Versions/login flows are fetched from current homeserver A. The following well-known request may return `m.homeserver.base_url = B`, but `newClient.homeserver` is never updated or B validated:
>   https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/lib/widgets/matrix.dart#L1179-L1191
>   
>   Result: Matrix traffic continues against A while UI/configuration reads discovery pointing to B.
> 
> * requestAndCache() has no in-flight deduplication, so two concurrent cache misses start two requests. MatrixState also has no guard before assigning loginHomeserverSummary, so a late result may overwrite a newer one.
>   https://github.com/famedly/matrix-dart-sdk/blob/6a7f917bf98a6bb97ae8878cd52141e1931301eb/lib/src/utils/request_and_cache.dart#L18-L27
>   Two concurrent cache misses therefore start two requests. Their completion order also determines the final published global summary.
> 
> The current “restart” test reuses the same `Client` and in-memory database, and only removes the well-known endpoint:
> 
> https://github.com/linagora/twake-on-matrix/blob/4d1eaa2ed674cc80b4f13f8e886d7a8f961b3eed/test/utils/matrix_sdk_extensions/client_well_known_extension_test.dart#L129-L152
> 
> So it does not cover a cold restart, a reopened production database, a fully offline server, account switching, or concurrent refreshes.
> 
> wdyt?

Thank you very much for the review, I really appreciate and the issues you mentioned are legit  ! As discussed I updated to take care of the point 1, 3 and the part of 5. (use cache when server unreachable, persist fallback to survives next restart and the fact that a late result may overwrite newer account summary. Please let me know what you think @9clg6 @nqhhdev :)

9clg6:
LGTM

9clg6 (approved):

nqhhdev (approved):
LGTM

Issue #3218, opened 2026-07-06
New UX and UI for device verification


tprudentova:
# Design 

https://www.figma.com/design/uLw3iWnOEGaRlJnfPz4Pv0/Twake--Matrix-?node-id=41647-9565&p=f&t=313XwGfBAYXL9VAY-0

# User story

When user connects a new device, verification should be performed automatically. However, if the automatic setup fails, user sees a modal to either retry automatic setup or verify the session manually

<img width="1004" height="642" alt="Image" src="https://github.com/user-attachments/assets/1d87bd7e-859b-42df-8fa8-b4689b981cc2" />

The modal includes the following buttons: 

- Retry automatically: forces another try of automatic setup. If the device is verified successfully, user sees a happy confirmation modal

<img width="1032" height="659" alt="Image" src="https://github.com/user-attachments/assets/2045828e-e93b-4f7f-9caf-07c8b0f2a722" />

- Use recovery key. Clicking it opens a modal for entering the recovery key. If the device is verified successfully, user sees a happy confirmation modal

<img width="957" height="615" alt="Image" src="https://github.com/user-attachments/assets/ef1b983d-a9db-4f83-9320-af175b438f46" />

- Use another device. Clicking it opens the device verification flow.

<img width="944" height="605" alt="Image" src="https://github.com/user-attachments/assets/17be8d6b-5bce-4254-ad14-b90eccbd8d5c" />

<img width="945" height="606" alt="Image" src="https://github.com/user-attachments/assets/603a7c9b-e599-4d04-90d9-b7c8706dd069" />

<img width="944" height="608" alt="Image" src="https://github.com/user-attachments/assets/287703f5-78e8-40c8-b37f-0d9dbfcde7c2" />

- Reset your identity (’Not possible to verify?’). Clicking it shows a verification modal, if user clicks Reset, their cryptographic identity is reset completely

<img width="1008" height="646" alt="Image" src="https://github.com/user-attachments/assets/49fd8837-8b73-4b00-ab1e-457066825de9" />

<img width="1006" height="647" alt="Image" src="https://github.com/user-attachments/assets/897d3035-b460-4c3a-a679-30a99312a242" />

In encrypted chats there is a banner if user has some unverified devices (including the current session). Clicking Verify redirect to Devices in Settings

<img width="1126" height="722" alt="Image" src="https://github.com/user-attachments/assets/9ec4d6d6-0a51-409d-a082-8ad744863261" />

[...]

PR #3257, opened 2026-07-14
[PART I] TW-3218: ADD device verification warning banner in chat screen

nqhhdev:
## Ticket
**Related issue**

- #3218

 What: Warn users when a device is unverified (can't decrypt messages).                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                               
  - New banner in chat screen (own session unverified) — dismissible on desktop, TODO verify-action for Part II                                                                                                                                                                                
  - New banner in device settings (any device unverified)                                                                                                                                                                                                                                      
  - Both use encryptToDevice == false as the trust check (not verified/directVerified)                                                                                                                                                                                                         
  - Refactored device_settings.dart from StatefulWidget → Riverpod (DevicesSettingsViewModel + sealed state), added GetDevicesInteractor                                                                                                                                                       
  - Unrelated: bootstrap_dialog.dart got mounted guards on 8 post-frame callbacks   


https://github.com/user-attachments/assets/bd7b7f46-51a3-43b9-8719-1bd12acbfe00



## Summary by CodeRabbit

* **New Features**
  * Added a device verification warning banner to encrypted chat views and the draft chat flow, with localized messaging and dismiss behavior.
  * Introduced a devices fetch flow and richer device settings UI state to drive banner visibility and loading/errors.
* **Bug Fixes**
  * Prevented bootstrap actions from running after the widget is disposed.
  * Updated device “enabled/verified” indicators to use per-device encryption capability (`encryptToDevice`).
* **Chores**
  * Updated localization for encryption/key verification warnings, and refreshed device settings wiring plus integration test targeting.


jbrnzl:
It would be amazing to have a little demo and PR description 👍🏻

nqhhdev:
> It would be amazing to have a little demo and PR description 👍🏻

Updated @jbrnzl

pm-McFly:
I have notified the Design team, but the "verify button" on the `current` device is a non sense.

image

How could you know that you are unverified?
On matrix you always trust yourself, but you don't trust others.

So current device is always green, but all `remotes` are either:

- green: trust bond between `current` and `remote` verified
- orange: `remote` device is not trusted by `current`

---

The figma should be updated, ping me if not ;)

nqhhdev:
> I have notified the Design team, but the "verify button" on the `current` device is a non sense.
>
> So current device is always green, but all `remotes` are either:
> 
> * green: trust bond between `current` and `remote` verified
> * orange: `remote` device is not trusted by `current`
> 
> The figma should be updated, ping me if not ;)

Thanks @pm-McFly, that's the legacy issue, 

Before:                                                                                                                                                                                                                                                                                      
  - Used verified / directVerified to decide when to show the warning.                                                                                                                                                                                                                       
  - Problem: these flags are always true for your own current device (the SDK auto-trusts yourself, you can never be "unverified" to yourself). So using this flag meant the banner for the current device could never show — pointless.                                                       
                                                                                                                                                                                                                                                                                               
  After (this PR):                                                                                                                                                                                                                                                                             
  - Switched to checking encryptToDevice == false.                                                                                                                                                                                                                                             
  - Meaning: this flag tells whether a device is actually receiving the room encryption keys — regardless of whether it's your own device or someone else's. If false, that device is at risk of failing to decrypt messages, even if it's your own current session.                           
  - That's why the banner can show up even for the current device — it's warning about "key sync risk," not "you haven't trusted yourself."

pm-McFly:
> After (this PR):
> 
> * Switched to checking encryptToDevice == false.
> * Meaning: this flag tells whether a device is actually receiving the room encryption keys — regardless of whether it's your own device or someone else's. If false, that device is at risk of failing to decrypt messages, even if it's your own current session.
> * That's why the banner can show up even for the current device — it's warning about "key sync risk," not "you haven't trusted yourself."

Okay for the banner, but the "verify" button on the current device row of the device view should not be present.

The verify button allows you to target which device you want the current one to be verified with.

---

| device name | verified |
| --- | --- |
| `current` | Not Applicable |
| Chrome Browser | `false` |
| Android Tablet | `true` |
| iPhone | `false` |

<details> <summary>**Web of Trust**</summary>

image

</details>

in this situation we should only see the verify button on the rows of:

- Chrome Browser
- iPhone

Nothing else.

If the user chooses to verify with iPhone, then the list gets updated to - after verification is successfully completed:

| device name | verified |
| --- | --- |
| `current` | Not Applicable |
| Chrome Browser | `false` |
| Android Tablet | `true` |
| iPhone | `true` |

<details> <summary>**Web of Trust**</summary>

image

</details>

nqhhdev:
> Okay for the banner, but the "verify" button on the current device row of the device view should not be present.
> 
> The verify button allows you to target which device you want the current one to be verified with.

1. Should the verify button key off verified, or off the same encryptToDevice signal as the banner? If a device is verified: true but encryptToDevice: false, does it get a button?                                                                                                          
2. Is "hide button when verified == true" actually intended, or should a verified device still be re-verifiable? Their example doesn't disambiguate — no verified device appears in the "should show" list, but they also never say to hide it.

cc: @pm-McFly

pm-McFly:
Good points, thx! 

> 1. Should the verify button key off verified, or off the same encryptToDevice signal as the banner? If a device is verified: true but encryptToDevice: false, does it get a button?

IMO the verify button/advisory should be off only if both `verified` and `encryptToDevice` are `true`. That for each row of the devices list. Meaning that if a device is not entirely trusted, it must be verified.

> 2. Is "hide button when verified == true" actually intended, or should a verified device still be re-verifiable? Their example doesn't disambiguate — no verified device appears in the "should show" list, but they also never say to hide it.

Question for @tprudentova: Can a user "re-verify" a device?
*Currently under discussion with product team*

tprudentova:
> Question for @tprudentova: Can a user "re-verify" a device? _Currently under discussion with product team_

Yep, added a design for that. Behaiour when user clicks/taps on some session 
https://www.figma.com/design/uLw3iWnOEGaRlJnfPz4Pv0/Twake--Matrix-?node-id=42540-18622&t=cniPOAEbwt4eKg7E-0

lib/domain/usecase/device_settings/get_devices_interactor.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:dartz/dartz.dart';
+import 'package:fluffychat/app_state/failure.dart';
+import 'package:fluffychat/app_state/success.dart';
+import 'package:fluffychat/domain/app_state/device_settings/get_devices_state.dart';
+import 'package:matrix/matrix.dart';
+
+class GetDevicesInteractor {
+  Stream<Either<Failure, Success>> execute({required Client client}) async* {
```
9clg6, line 8: Didn't we remove Either pattern ?

lib/domain/app_state/device_settings/get_devices_state.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:fluffychat/app_state/failure.dart';
+import 'package:fluffychat/app_state/success.dart';
+import 'package:matrix/matrix.dart';
+
+class GetDevicesLoading extends Success {
```
9clg6, line 5: Is it still relevant with the riverpod's success/error/loading pattern ?

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner.dart
```diff
@@ -0,0 +1,32 @@
...
+class ChatDeviceVerificationBanner extends ConsumerWidget {
+  final Client client;
+
+  const ChatDeviceVerificationBanner({super.key, required this.client});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final provider = chatDeviceVerificationBannerViewModelProvider(client);
+    final state = ref.watch(provider);
+    final controller = ref.read(provider.notifier);
```
9clg6, line 20: Prefer passing the family invocation directly to ref.watch and ref.read. This keeps the provider statically identifiable by Riverpod tooling. Calling the family twice with the same client still resolves to the same provider state, so this does not instantiate duplicate notifiers. : 

```
final state = ref.watch(
  chatDeviceVerificationBannerViewModelProvider(client),
);
final controller = ref.read(
  chatDeviceVerificationBannerViewModelProvider(client).notifier,
);
```

instead of 

```
    final provider = chatDeviceVerificationBannerViewModelProvider(client);
    final state = ref.watch(provider);
    final controller = ref.read(provider.notifier);
```

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
```diff
@@ -0,0 +1,53 @@
...
+  DevicesBannerState build(Client client) {
+    // Device keys can still be loading (or change later, e.g. after
+    // completing verification elsewhere), so re-evaluate on every sync
+    // rather than only once at build time.
+    _onSyncSubscription = client.onSync.stream.listen((_) => _refresh());
+    ref.onDispose(() => _onSyncSubscription?.cancel());
+    return _computeState();
+  }
+
+  final responsive = getIt.get<ResponsiveUtils>();
```
9clg6: Could you use the getIt bridge ? 
```
@Riverpod(keepAlive: true)
GetIt getIt(Ref ref) => GetIt.instance;
```
and after
`final responsive = ref.read(getItProvider).get<ResponsiveUtils>();`

@aparesy will do same in a followup PR for his PRs (so IDK if you already have it or not, you do it in this PR if already merged or in the alexandre's followup, WDYT ?)

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  void _refresh() => state = _computeState();
```
9clg6, line 25: To prevent useless rebuild, could we use 

```
void _refresh() {
  final newState = _computeState();
  if (newState == state) return;

  state = newState;
}
```
instead ?

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
9clg6: And tbh I never liked this ResponsiveUtils, because it depends on the BuildContext, and the BC should be exposed only in the UI, not in the VM so : 

```
return LinagoraBanner(
  message: L10n.of(context)!.deviceVerificationWaring,
  onDismiss: isMobile ? null : controller.onDismissBanner,
);
```
is more appropriate in a strict respect of Clean Architecture

lib/pages/chat_draft/draft_chat_view.dart
```diff
@@ -313,6 +314,7 @@ class DraftChatView extends StatelessWidget {
               show: controller.showAddContactBanner,
             ),
           ),
+          ChatDeviceVerificationBanner(client: Matrix.of(context).client),
```
9clg6: Since this is a new Riverpod component, should it obtain the active client from the planned matrixClientProvider instead of introducing another Matrix.of(context).client dependency? The existing Matrix.of(context) usages can remain legacy for now, but new Riverpod code should follow the migration architecture. If matrixClientProvider is not available yet, this PR may depend on the corresponding infrastructure work.

```
    final client = ref.watch(matrixClientProvider);

    final state = ref.watch(
      chatDeviceVerificationBannerViewModelProvider(client),
    );
```

for example, with 

```
@Riverpod(keepAlive: true)
Client matrixClient(Ref ref) {
  throw UnimplementedError(
    'matrixClientProvider must be overridden at application startup',
  );
}
```
+
```
ProviderScope(
  overrides: [
    matrixClientProvider.overrideWithValue(client),
  ],
  child: const TwakeApp(),
);
```

WDYT ?

@aparesy @jbrnzl

lib/pages/device_settings/device_settings_state.dart
```diff
@@ -0,0 +1,71 @@
+import 'package:equatable/equatable.dart';
+import 'package:matrix/matrix.dart';
+
+/// State of the device-verification warning banner shown atop the list.
+enum VerificationBannerVisibility { shown, dismissed }
+
+sealed class DevicesSettingsState extends Equatable {
```
9clg6: Instead of Equatable could we use @freezed as the Invitation Module did ?

lib/pages/device_settings/device_settings_view_model.dart
```diff
@@ -0,0 +1,139 @@
...
+import 'package:fluffychat/domain/usecase/device_settings/get_devices_interactor.dart';
+import 'package:fluffychat/pages/device_settings/device_settings_state.dart';
+import 'package:matrix/matrix.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'device_settings_view_model.g.dart';
+
+@riverpod
+class DevicesSettingsViewModel extends _$DevicesSettingsViewModel {
+  final GetDevicesInteractor _getDevicesInteractor = getIt
```
9clg6: Could we use Riverpod for DI ?

lib/pages/device_settings/device_settings_view_model.dart
```diff
@@ -0,0 +1,139 @@
...
+  Future<void> loadUserDevices(Client client) {
+    if (_devices != null) return Future.value();
+    return _loadInFlight ??= _loadUserDevices(client).whenComplete(() {
+      _loadInFlight = null;
+    });
+  }
+
+  Future<void> _loadUserDevices(Client client) async {
+    await _getDevicesInteractor.execute(client: client).forEach((either) {
+      either.fold(
```
9clg6: Don't we went remove Either pattern ? try/catch could do the same here WDYT ?

lib/pages/device_settings/user_device_list_item.dart
```diff
@@ -12,129 +13,101 @@ enum UserDeviceListItemAction { rename, remove, verify, block, unblock }
...
+    DeviceKeys? keys,
+    bool isOwnDevice,
+  ) async {
+    final action = await showModalActionSheet<UserDeviceListItemAction>(
+      context: context,
+      title: '${userDevice.displayName} (${userDevice.deviceId})',
+      actions: [
+        SheetAction(
+          key: UserDeviceListItemAction.rename,
+          label: L10n.of(context)!.changeDeviceName,
```
9clg6: it can be refacto in final l10n = L10n.of(context)!;

lib/domain/usecase/device_settings/get_devices_interactor.dart
nqhhdev: We're not removing Either from the interactor layer — see #3233, interactor stayed unchanged there too. Either is only fold()-ed inside the ViewModel and never leaks into state/UI Same pattern as contacts invitation.

lib/pages/device_settings/device_settings_view_model.dart
nqhhdev: Same answer as the interactor thread — `Either` from the interactor gets `fold()`-ed right here into `DevicesSettingsState`, which itself doesn't carry `Either` at all. This matches how #3233 handled the invitation interactors, so keeping it consistent rather than swapping to try/catch.

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
nqhhdev: `GetIt.instance` is already a global singleton with no reactive state, so wrapping it in a Riverpod provider (`ref.read(getItProvider)`) doesn't add anything — no override/dispose benefit, since the singleton itself never changes. Keeping `getIt.get<T>()` direct for now; happy to adopt the bridge later if @aparesy's PR lands and the team wants that consistency project-wide.

lib/pages/chat_draft/draft_chat_view.dart
nqhhdev: `matrixClientProvider` doesn't exist yet, and migrating off `Matrix.of(context)` here would ripple into a much wider change than this PR can safely test. Keeping `Matrix.of(context).client` — agreed this is fine as legacy usage per your note.

lib/pages/device_settings/device_settings_view_model.dart
nqhhdev: Not needed here — `GetDevicesInteractor` is stateless with no ref-scoped lifecycle requirement, so switching just moves where the dependency is declared without changing behavior. Mixing GetIt + Riverpod DI for the same kind of dependency also adds inconsistency rather than reducing it. GetIt remains the project's DI convention per CLAUDE.md, so keeping it as-is.

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
nqhhdev: Fixed — moved `isMobile` computation into the widget (`chat_device_verification_banner.dart`), the ViewModel no longer holds a `ResponsiveUtils` reference or touches `BuildContext`.

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner.dart
nqhhdev: Fixed — now passes the family provider directly to both `ref.watch` and `ref.read`.

lib/pages/device_settings/device_settings_state.dart
nqhhdev: Done — migrated to `@freezed` sealed union, consistent with the contacts invitation module.

lib/pages/device_settings/user_device_list_item.dart
nqhhdev: Done — hoisted `l10n` as a local in both `_openActionSheet` and `build`.

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
nqhhdev: Riverpod already skips notify when state = x sets an equal value, and DevicesBannerState extends Equatable — so this is already a no-op today, just implicit. Can make it explicit if you prefer.

lib/pages/chat_draft/draft_chat_view.dart
aparesy: I'd argue that since there are other call sites to the soon-to-be-deprecated provider at lines 234, 283, 392 in the same file, then it would be not that problematic to keep it that way in order to stay constant within the file and avoid external sources of errors. But yeah also since we want to make progress, it's not a bad idea to migrate now either 🤷 Depends if we prefer safety or efficiency I guess

lib/domain/usecase/device_settings/get_devices_interactor.dart
9clg6: Yes you are right, i didn't remembered if we migrate yet or not

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
9clg6: > `GetIt.instance` is already a global singleton with no reactive state, so wrapping it in a Riverpod provider (`ref.read(getItProvider)`) doesn't add anything — no override/dispose benefit, since the singleton itself never changes. Keeping `getIt.get<T>()` direct for now; happy to adopt the bridge later if @aparesy's PR lands and the team wants that consistency project-wide.

I don’t think the point of the bridge is to make GetIt.instance reactive or give it a Riverpod lifecycle. It is a migration boundary: migrated Riverpod code should no longer import or consume GetIt directly, so that GetIt remains an implementation detail behind providers.

That said, a typed bridge such as getDevicesInteractorProvider would be more useful than exposing the whole GetIt container through getItProvider: consumers could then be overridden independently in tests, and removing GetIt later would only require changing the provider implementation rather than every call site. This also matches the merged migration plan, which says that migrated modules must not directly import GetIt.

lib/pages/chat_draft/draft_chat_view.dart
9clg6: > `matrixClientProvider` doesn't exist yet, and migrating off `Matrix.of(context)` here would ripple into a much wider change than this PR can safely test. Keeping `Matrix.of(context).client` — agreed this is fine as legacy usage per your note.

Existing Matrix.of(context) usages can remain until their code paths are migrated, but we should apply a ratchet rule and avoid introducing new ones. If matrixClientProvider requires broader bootstrap work, let’s isolate that in a focused prerequisite PR rather than defer the architecture indefinitely.

lib/pages/device_settings/device_settings_view_model.dart
9clg6: This is not about GetDevicesInteractor being stateful or needing a ref-scoped lifecycle. The provider is the DI boundary for migrated Riverpod code: the ViewModel should depend on a typed Riverpod provider, while GetIt remains an implementation detail behind the temporary bridge. As we decided in the migration plan + the guidelines + invitation module already-merged PR (https://github.com/linagora/twake-flutter-guidelines/blob/e419768cfdb7c6bf43c62af60a5d1f693414e7c4/GUIDELINES.md#27-dependency-injection--migration-to-riverpod)

This is already the pattern used by the merged Invitation module in #3233: its stateless interactors are exposed through typed providers, and the ViewModel only uses ref.read(...InteractorProvider). It also allows tests to override each interactor directly through ProviderContainer.

The migration plan explicitly states that migrated modules must not directly import GetIt and that dependencies crossing the legacy/migrated boundary should go through bridge providers. So for consistency, I think we should expose GetDevicesInteractor through a typed provider here as well.

lib/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner_view_model.dart
nqhhdev: Updated

lib/pages/device_settings/device_settings_view_model.dart
nqhhdev: Updated

lib/pages/chat_draft/draft_chat_view.dart
9clg6: @nqhhdev

nqhhdev: @9clg6 @jbrnzl 


Good point in principle. But Client here is dynamic (changes on login/account switch via MatrixState.setActiveClient), not a static singleton like getItProvider. Wiring matrixClientProvider properly means overriding it through that whole switching flow — that's separate infra work, not something to bundle into this banner PR.

For now I'll keep client: controller.client and follow up with matrixClientProvider as its own PR once the infra is ready.

jbrnzl:
Little feedback after testing @nqhhdev : 
Capture d’écran 2026-07-29 à 11 48 30

lib/pages/chat_draft/draft_chat_view.dart
jbrnzl: Here is what I suggest: we do not block this Part 1 for that topic (maybe add a TODO), and we add this (the matrixClientProvider) in the scope of Part 3 so we do it but we can still move on ? What do you think @9clg6 @nqhhdev

nqhhdev: IMO: We can do it after finishing the feature because I don't want to block the release for this feature, and when we try migrating Riverpod, it also takes a lot of time to verify. So I will create a new ticket about migrating the Matrix client class to use Riverpod, wdyt?

jbrnzl: Ok fine for me if the ticket is created and referenced in a todo an is a priority right after the device verification changes

nqhhdev: New ticket #3302

lib/pages/device_settings/device_settings.dart
```diff
@@ -1,138 +1,320 @@
...
+            );
+          }
+          if (snapshot.connectionState != ConnectionState.done ||
+              state is DevicesSettingsInitial) {
+            return const MaxWidthBody(
+              child: Center(
+                child: CircularProgressIndicator.adaptive(strokeWidth: 2),
+              ),
+            );
+          }
```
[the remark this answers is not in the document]
nqhhdev: updated

lib/pages/chat_draft/draft_chat_view.dart
```diff
@@ -7,6 +7,7 @@ import 'package:fluffychat/pages/chat/blocked_user_banner.dart';
 import 'package:fluffychat/pages/chat/chat.dart';
 import 'package:fluffychat/pages/chat/chat_app_bar_title_style.dart';
 import 'package:fluffychat/pages/chat/chat_background.dart';
+import 'package:fluffychat/pages/chat/chat_device_verifycation_banner/chat_device_verification_banner.dart';
```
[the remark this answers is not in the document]
nqhhdev: updated

jbrnzl (approved):

9clg6 (approved):

PR #3272, opened 2026-07-17
fix: make video call button react to well-known updates

### LINKED ISSUE #3265: [BUG] Visio feature on and off

### Describe the bug

The camera button sometimes disappears, and the visio link bubbles sometimes change into usual messages with links

https://github.com/user-attachments/assets/5693f698-39c1-489a-b3ca-d26c34836000

### Pre-Bug State

.

### To Reproduce

I don't have specific steps, but sometimes it happens... It helps me to reload the app, but the button still doesn't come back immediately @pm-McFly maybe it's on your side? 

### Expected results

.

### Context

- Device: Redmi Note 12 Pro
- OS: Android
- App Version: 2.23.3
- Server: linagora
- Account information:
- Others

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
## Ticket
Fixes #3265 (front-end part, not deployment issue)

## Root cause
loginHomeserverSummary is a plain field on "MatrixState" fetched (or served from cache) but nothing rebuilds the chat app bar when fetched, so after a reload the call button only shows up after a rebuild (ex: room switch, new message, ...).

## Solution
Keep the summary in a Riverpod provider and watch it from the app bar, so the button appears as soon as discovery arrives. `MatrixState` keeps a getter/setter so legacy call sites don't change

## Demo

https://github.com/user-attachments/assets/b9b2e67d-1504-4886-8864-a98841708d73

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated chat video-call button and centralized the start-video-call action.
  * Video-call UI now appears only when chat messaging is allowed and the active server supports video calls.

* **Bug Fixes**
  * Video-call link handling and availability now consistently follow the active server’s configuration across chat UI and message rendering.

* **Refactor**
  * Moved homeserver discovery and video-call base URL lookup to Riverpod-backed state, reducing reliance on context-based access.

* **Tests**
  * Added Riverpod tests to verify homeserver summary provider initialization, updates, and keep-alive behavior.

### DIFF
--- a/lib/pages/chat/chat_view.dart
+++ b/lib/pages/chat/chat_view.dart
@@ -1,20 +1,17 @@
-import 'package:fluffychat/domain/model/extensions/homeserver_summary_extensions.dart';
 import 'package:fluffychat/pages/chat/chat.dart';
 import 'package:fluffychat/widgets/twake_components/unread_count_badge.dart';
 import 'package:fluffychat/pages/chat/chat_app_bar_title.dart';
 import 'package:fluffychat/pages/chat/chat_invitation_body.dart';
+import 'package:fluffychat/pages/chat/chat_video_call_button.dart';
 import 'package:fluffychat/pages/chat/chat_view_body.dart';
 import 'package:fluffychat/pages/chat/chat_view_style.dart';
 import 'package:fluffychat/pages/chat/events/message_content_mixin.dart';
-import 'package:fluffychat/providers/login_homeserver_summary_provider.dart';
-import 'package:fluffychat/utils/voip/video_call_helper.dart';
 import 'package:fluffychat/presentation/mixins/audio_mixin.dart';
 import 'package:fluffychat/resource/image_paths.dart';
 import 'package:fluffychat/utils/stream_extension.dart';
 import 'package:fluffychat/widgets/matrix.dart';
 import 'package:fluffychat/widgets/twake_components/twake_icon_button.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:fluffychat/generated/l10n/app_localizations.dart';
 import 'package:linagora_design_flutter/colors/linagora_state_layer.dart';
 import 'package:linagora_design_flutter/colors/linagora_sys_colors.dart';
@@ -164,32 +161,7 @@ class ChatView extends StatelessWidget with MessageContentMixin {
                             onPressed: controller.toggleSearch,
                             icon: const Icon(Icons.search),
                           ),
-                          Consumer(
-                            builder: (context, ref, __) {
-                              final videoCallBaseUrl = ref.watch(
-                                loginHomeserverSummaryProvider.select(
-                                  (summary) => summary?.videoCallBaseUrl,
-                                ),
-                              );
-                              if (!controller.canStartVideoCall(
-                                videoCallBaseUrl,
-                              )) {
-                                return const SizedBox.shrink();
-                              }
-                              return TwakeIconButton(
-                                icon: Icons.videocam_outlined,
-                                tooltip: L10n.of(context)!.startVideoCall,
-                                onTap: () => VideoCallHelper.start(
-                                  room: controller.room,
-                                  startedTitle: L10n.of(
-                                    context,
-                                  )!.videoCallStartedTitle,
-                                  baseUrl: videoCallBaseUrl,
-                                ),
-                                preferBelow: false,
-                              );
-                            },
-                          ),
+                          ChatVideoCallButton(controller),
                           if (controller.hasActionAppBarMenu)
                             Builder(
                               builder: (context) => TwakeIconButton(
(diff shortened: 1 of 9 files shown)


lib/pages/chat/chat_view.dart
```diff
@@ -162,21 +164,32 @@ class ChatView extends StatelessWidget with MessageContentMixin {
...
-                                startedTitle: L10n.of(
-                                  context,
-                                )!.videoCallStartedTitle,
-                                baseUrl: Matrix.of(
-                                  context,
-                                ).loginHomeserverSummary?.videoCallBaseUrl,
-                              ),
-                              preferBelow: false,
-                            ),
+                          Consumer(
```
9clg6: Could we extract this into a small ConsumerWidget, such as _VideoCallButton, to avoid the inline Consumer while keeping rebuilds scoped to the video-call action?

I would like in mid/short-term to start to refactor this page, by splitting everything in smaller module to make the code easier to read and make rebuild more efficient

test/providers/login_homeserver_summary_provider_test.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:fluffychat/domain/model/homeserver_summary.dart';
+import 'package:fluffychat/providers/login_homeserver_summary_provider.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:matrix/matrix.dart';
+
+void main() {
+  late ProviderContainer container;
+
+  setUp(() {
```
9clg6: I'm not sure about setUp, WDYT to use ProviderContainer.test() instead ? Example : 

```dart 
test('starts with no summary', () {
  final container = ProviderContainer.test();

  expect(container.read(loginHomeserverSummaryProvider), isNull);
});

test('set publishes the summary to watchers', () {
  final container = ProviderContainer.test();
  final seen = <HomeserverSummary?>[];

  container.listen(
    loginHomeserverSummaryProvider,
    (_, next) => seen.add(next),
    fireImmediately: true,
  );

  final summary = buildSummary();
  container.read(loginHomeserverSummaryProvider.notifier).set(summary);

  expect(seen, [null, summary]);
});
```

9clg6 (commented):
Nice work btw, it's good to see Riverpod around here 😎

9clg6 (approved):

lib/pages/chat/chat_view.dart
jbrnzl: Thank you for the review, I agree. I updated the test and extracted into a small ConsumerWidget, please let me know if everything is good for you now @9clg6

9clg6: LGTM

9clg6 (approved):

lib/widgets/matrix.dart
nqhhdev, line 1: - Why convert the whole widget to Riverpod's state type just for one setter to call ref.read(...)? Could write to the provider from outside instead (e.g. ProviderScope.containerOf(context, listen: false)), keeping the widget's original state type. 

- If we apply Riverpod for this class, we should migrate all functions and build structures for this path (I mean we should migrate all controller AND UI classes). Besides, it can take more time for testing and verifying to avoid any regression issues.

So should we keep Stateful for this widget, wdyt?

jbrnzl: Thank you for the questions @nqhhdev :)

1. Well, to be honest I also thought about it but this is not the recommended way to go: I have found this interesting discussion where riverpod author did not recommend context-based provider access as an API ([riverpod#134](https://github.com/rrousselGit/riverpod/issues/134)). Also the getter is read in 6 or 7 places inside MatrixState from what I remember, so we'd be re-deriving the container everywhere instead of using the supported ref...

2. Well imo it's the opposite: the full migration can be the real regression risk. Here only one field moves, nothing else changes. And the rest can migrate later, incrementally. With this PR I wanted to avoid migrating too much things as it can be a mistake I think, the goal here is just to fix a bug

Let me know what you think and if you agree/disagree 👍🏻

pm-McFly: @9clg6 any thoughts?

nqhhdev: @jbrnzl Could you add the comment TODO for this func `loginHomeserverSummary`, we will migrate it later, WDYT?

jbrnzl: @nqhhdev added :)

nqhhdev (approved):
LGTM

lib/widgets/matrix.dart
9clg6: large screens should generally be split into smaller, focused widgets. It improves readability and maintainability, and when provider subscriptions are colocated in those widgets, it also helps scope rebuilds to the parts that actually depend on the state (e.g ref.watch).

That said, converting Matrix to a ConsumerStatefulWidget does not introduce additional rebuilds by itself here, since we only use ref.read. It simply gives us access to Riverpod through the supported ref API while keeping the migration incremental

Issue #3274, opened 2026-07-20
[BUG] Search result highlighting does not support diacritics


aparesy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When using any search bar that supports diacritics (characters such as "à" or "é"), the highlighting of matching letters in orange only works when the match is done exactly, but not when a diacritic is matched with a common letter, or another diacritic

### Pre-Bug State

.

### To Reproduce

1. Go to the main search bar
2. Type the name of one of your contacts while replacing a non-diacritic letter with a diacritic (for example e -> é)

### Expected results

The matched special characters should be highlighted in orange

### Context

Observed on Android and web (Chrome)

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

https://github.com/user-attachments/assets/a4d050fc-80dd-4470-9b7f-c3258b03cbcb

</details>

aparesy:
Actually @tprudentova do we agree that this is really a behavior that we want to change; like do we still want to highlight the result even if there is not an exact match?
Because typically in Whatsapp it seems like the highlighting is diacritics-insensitive only for group/contacts names but not for messages

<img width="553" height="512" alt="Image" src="https://github.com/user-attachments/assets/0e02c856-90d4-43dd-9c3d-46842815e212" />

pm-McFly:
> Actually [@tprudentova](https://github.com/tprudentova) do we agree that this is really a behavior that we want to change; like do we still want to highlight the result even if there is not an exact match? Because typically in Whatsapp it seems like the highlighting is diacritics-insensitive only for group/contacts names but not for messages

As discussed today altogether; highlight should be on all matches/results. If diacritics is off, we highlight all (`zoé`, `zoe`, `zòé`...).
Users will be able to limit the greediness when we will introduce strict/exact match with double quotes

PR #3286, opened 2026-07-23
TW-3274: Search result highlighting does not support diacritics

aparesy:
## Ticket
- #3274 

## Root cause
Since the recent improvements of the search engine (see #3129), it was observed that the highlighting of the search results from the various impacted search bars did not adapt to support the more recent version of the search. For example, when "élie" appears as a successful match for "elie", the result is displayed but not highlighted as it should be.
This is because the highlighting as it stands before this PR follows separate logic from the search engine and only does a direct comparison of the needle with the haystack (while taking into account uppercase/lowercase letters though)  

## Solution                                                                                                                                                                                                                      
Our solution is to bring the highlighting logic closer to the search engine, notably by making `SearchEngine` the single source of truth for "what counts as a match":                                                           
                                                                                                                                                                                                                                   - Added `SearchEngine.matchRanges(needle, haystack)`, which returns the positions (UTF-16 offset ranges into the original string) of every match, computed through the same normalization pipeline the engine already uses for filtering (diacritic stripping + case folding). Offsets are mapped back rune-by-rune, so accents, decomposed Unicode, Vietnamese stacked diacritics and emoji don't shift the highlight.                                                                                                                                                                                                                                                                            - `HighlightText` now slices its text into spans at those ranges instead of running its own regex. It defaults to `diacriticSensitive: false` (the convention used by every filtering call site), so all existing usages (contacts, recent chats, display names, chat search and server search) get diacritic-aware highlighting without any call-site change.                                                                                            
- `FileNameText` (filename labels on file-message rows in chat search results) now delegates to `HighlightText` instead of building its own span tree, so file rows and text rows in the same results list highlight consistently.                                                                                                                                                                                                                    
- `substringToHighlight` (the "..." preview windowing for long messages) now locates the match via `matchRanges` as well, so it can window around a diacritic-insensitive match instead of silently falling back to the untruncated message.                                                                                                                                                                                                             - The legacy regex-based `String.buildHighlightTextSpans` had no remaining callers and was removed.                                                                                                                              
                                                                                                                                                                                                                                   
  ## Impact description                                                                                                                                                                                                            
  Display layer only: no search's *matching* behavior changes — result lists contain exactly the same items as before; only how matches are highlighted changes. Note that message search matching itself (encrypted-room local search and the server-search post-filter) is still diacritic-sensitive; aligning those matchers with `SearchEngine` is a separate follow-up, for which this PR is the display-side groundwork.                                   
                                                                                                                                                                                                                                   
  ## Test recommendations                                                                                                                                                                                                          
  Automated (all added in this PR):                                                                                                                                                                                                
  - `flutter test test/utils/search/match_ranges_test.dart` — 16 cases covering offset correctness: diacritics, Vietnamese stacked marks, pre-decomposed (NFD) input, emoji/surrogate pairs, trailing combining marks, multiple    
  occurrences, exact mode.                                                                                                                                                                                                         
  - `flutter test test/widget/highlight_text_test.dart test/widget/file_widget/base_file_tile_widget_test.dart` — widget tests asserting the actual rendered `TextSpan` split ("elie" highlights "Élie").                          
  - `flutter test test/string_extension_test.dart` — includes a regression test for preview windowing around a diacritic-insensitive match beyond the prefix window.                                                               
                                                                                                                                                                                                                                   
  Manual:                                                                                                                                                                                                                          
  - Contacts search: search "elie" against a contact named "Élie" → the name should now be highlighted (previously displayed unhighlighted).                                                                                                                                                                                                                                                       
  - Recent chats / main search: verify highlighted room and contact names still render correctly for plain ASCII queries (no regression).                                                                                          
                                      
                                                                                                                                                                                                                                   
  ## Resolved                                                                                                                                                                                                                      
  - Web:


aparesy:
For now this remains as a proof-of-concept that the highlighting can be potentially improved, but the deployment is suspended as it is too early to make that big of changes.

PR #3308, opened 2026-08-20
SimpleMatcher2

aparesy:
This version is very similar to the current `SearchEngine`, except we remove the highlighting since this allows for more flexibility in the way matches are done, typically we can make any transformation (like diacritic folding) to both the needle and the haystack without worrying about having to reverse the operation in order to retrieve a highlighted span.

Just like in https://github.com/aparesy/twake-on-matrix/pull/1 one of the biggest issues to me is that the server-side search stays far behind the SimpleMatcher in terms of features, which creates a discrepancy in the quality of search results throughout the app


aparesy:
Sorry for the mistake, this was supposed to go on my local fork :/

PR #3313, opened 2026-08-27
Enable first three progressive linter rules

9clg6:
## Ticket

Related discussion: https://github.com/linagora/twake-on-matrix/discussions/3311

## Root cause

Not a bug. This is the first implementation step of the progressive linter rollout proposed in Discussion #3311.

## Solution

Enable and resolve the first three rules from the First wave, in the proposed order:

1. `avoid_field_initializers_in_const_classes`
2. `avoid_unused_constructor_parameters`
3. `matching_super_parameters`

Each rule and its corresponding fixes are isolated in a dedicated commit so the diagnostic impact stays reviewable.

## Impact description

- Adds three low-impact linter rules to `analysis_options.yaml`.
- Resolves 1, 2, and 1 diagnostics respectively.
- Keeps runtime behavior unchanged.

## Test recommendations

- `fvm dart analyze --format machine`
- `fvm flutter test test/mixin/contacts_view_controller_mixin_test.dart`
- Verify the three commits follow the rule order from Discussion #3311.

Local results:

- Analyzer passes with 0 diagnostics for all three new rules.
- Targeted test passes: 18 tests.
- `git diff --check` passes.

## Pre-merge

Nothing else required.

## Resolved

No UI changes.

- Web: N/A
- Android: N/A
- IOS: N/A

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal constructor and field declarations without changing visible behavior.
  * Improved consistency across shared components.

* **Tests**
  * Streamlined contact search test fixtures while preserving existing coverage.

* **Chores**
  * Added static analysis checks for unused constructor parameters, consistent parameter forwarding, and const class field patterns.

PR #3321, opened 2026-08-31
Enable prefer_constructors_over_static_methods lint

9clg6:
## Summary

- enable `prefer_constructors_over_static_methods`, the next rule from Discussion #3311
- replace `MimeTypeUitls.instance` with a factory constructor
- migrate the four internal call sites without changing MIME resolution behavior

## Validation

- repository pre-commit hook (`flutter analyze` and `dart format`)
- `./scripts/code_analyze.sh` with Flutter 3.38.9

Stacked on #3320.
Follows #3313.
Related to #3311.

## Summary by CodeRabbit

* **Improvements**
  * MIME type detection continues to work consistently when uploading, dragging, or selecting files on the web.
  * Public room interfaces can now be created more efficiently without changing their behavior.

* **Maintenance**
  * Updated code quality checks to encourage more consistent constructor and assertion patterns.

### DIFF
--- a/lib/utils/mime_type_uitls.dart
+++ b/lib/utils/mime_type_uitls.dart
@@ -5,7 +5,7 @@ import 'package:mime/mime.dart';
 class MimeTypeUitls {
   MimeTypeUitls._();
 
-  static MimeTypeUitls get instance => MimeTypeUitls._();
+  factory MimeTypeUitls() => MimeTypeUitls._();
 
   String getTwakeMimeType(String path) {
     final mimeType = lookupMimeType(path);
--- a/lib/utils/mime_type_uitls.dart
+++ b/lib/utils/mime_type_uitls.dart
@@ -5,9 +5,7 @@ import 'package:mime/mime.dart';
 class MimeTypeUitls {
   MimeTypeUitls._();
 
-  factory MimeTypeUitls() => MimeTypeUitls._();
-
-  String getTwakeMimeType(String path) {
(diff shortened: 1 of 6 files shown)


lib/utils/mime_type_uitls.dart
```diff
@@ -5,7 +5,7 @@ import 'package:mime/mime.dart';
 class MimeTypeUitls {
   MimeTypeUitls._();
 
-  static MimeTypeUitls get instance => MimeTypeUitls._();
+  factory MimeTypeUitls() => MimeTypeUitls._();
 
   String getTwakeMimeType(String path) {
```
jbrnzl: Hello !
Because there is no state here maybe we don't need a factory ? Only a pure function like `static String getTwakeMimeType` sounds simpler to me. What's your opinion ?

nqhhdev: Agreed

9clg6: Agreed — there is no instance state here, so a factory was just a lint-shaped wrapper.

`getTwakeMimeType` is now a static helper (`MimeTypeUitls.getTwakeMimeType(...)`) in 079f517d3. The private constructor stays so the class remains a namespace.

nqhhdev:
Pls resolve conflicts

jbrnzl (approved):

nqhhdev (approved):

PR #3328, opened 2026-09-02
fix(test): handle unavailable recovery key

9clg6:
## Summary

- expose an explicit keyed state when no recovery key is available
- let the mobile E2E scenario skip only that expected state
- keep unexpected visibility timeouts failing the test

## Validation

- targeted Dart analyze: no issues
- diff check: clean
- the equivalent recovery-key change passed the mobile shard_3 run in https://github.com/linagora/twake-on-matrix/actions/runs/33399437458

Split from #3310 so that PR remains scoped to CI sharding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when a recovery key is unavailable in Security Settings.
  * Recovery key actions and related checks are now skipped when no key is available, avoiding misleading failures.
  * Added clearer detection of the recovery key’s unavailable state.

### DIFF
--- a/integration_test/robots/setting/settings_recovery_key_robot.dart
+++ b/integration_test/robots/setting/settings_recovery_key_robot.dart
@@ -17,12 +17,33 @@ class SettingsRecoveryKeyRobot extends HomeRobot {
     return $(SettingsKeys.recoveryKeyItem.key);
   }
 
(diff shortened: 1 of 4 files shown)


nqhhdev (approved):
LGTM

jbrnzl (approved):

Issue #3325, opened 2026-09-01
Create a shortcut to create a Broadcast


jbrnzl:
This ticket goal is to introduce Broadcast/Feed feature by adding a shortcut in the "new message" menu who will create a "public_channel" room with the right permissions.

<img width="250" alt="Image" src="https://github.com/user-attachments/assets/e4fa6510-febe-455b-8fc2-17c10549a342" />

Wordings and screens will be the same as for group chats for now.

PR #3334, opened 2026-09-03
TW-3325: introduce feeds as group with public channel preset

jbrnzl:
## Ticket
Implements #3325

## Solution

Adds a **New Feed** entry to the creation menu: in the FAB menu on web, and in the contact list on mobile.

The flow reuses the existing group creation screens as is (on purpose for this first version). Difference is the encryption toggle is hidden as a feed is not encrypted.

The room creation is done with the `public_channel` preset and a room type set to `app.twake.chat.feed`, so a feed created by twake chat can be identified from a group easily.
Currently permissions are: only admins can post, while everyone can react and undo their own reactions.

## Impact description

One new menu entry. Nothing changes for existing chats and groups.

A feed looks and behaves like a group for now: same chat list, same screens. The only visible difference is that non-admin members cannot write by default.

## Test recommendations

Creating a feed requires a homeserver that supports the `public_channel` preset. On a standard server it may fails with an explicit message ("This server doesn't support feeds yet"), and no room is created.

In the future, the best could be to have a fallback to create a feed thought `power_level_content_override`. Currently there is no way to fetch all the preset a server knows from an endpoint.

- Create a feed
- Check the creator can write, and an invited member cannot
- Check that a member can react, then remove their reaction
- Check regular group creation is unchanged

## Demo

https://github.com/user-attachments/assets/bcfc810d-9589-4d8d-87e7-0a8ba9547564

## Summary by CodeRabbit

* **New Features**
  * Added support for creating feeds from the chat list and new-chat screens.
  * Added feed-specific creation flows with naming, optional avatars, and navigation.
  * Added English and French translations for feed creation and unsupported-server messages.
  * Displays a dedicated message when the server does not support feeds.

* **Tests**
  * Added coverage for feed creation, navigation, error handling, and server compatibility.


lib/domain/usecase/room/create_new_group_chat_interactor.dart
```diff
@@ -55,6 +63,16 @@ class CreateNewGroupChatInteractor {
       }
     } catch (exception, stackTrace) {
       Logs().e('CreateNewGroupChatInteractor', exception, stackTrace);
+      if (createNewGroupChatRequest.isFeed &&
+          exception is MatrixException &&
+          exception.errorMessage.toLowerCase().contains('preset')) {
```
nqhhdev, line 68: How can we reuse it and matching .contains('preset') on error text breaks if Synapse rewords it.

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import 'package:twake_chat/config/feed_config.dart';
+
+extension ClientFeedExtension on Client {
+  /// Raw request rather than [createGroupChat] because
+  /// [CreateRoomPreset] is a closed Dart enum: the feed preset cannot be
+  /// custom.
+  ///
+  /// Throws a [MatrixException] on a homeserver that does not know the preset.
+  Future<String> createFeedRoom({
```
nqhhdev, line 11: IMO: should create interactor

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
```diff
@@ -0,0 +1,39 @@
...
+        'creation_content': {'type': FeedConfig.roomType},
+        if (feedName != null) 'name': feedName,
+        if (initialState != null)
+          'initial_state': initialState.map((state) => state.toJson()).toList(),
+        if (powerLevelContentOverride != null)
+          'power_level_content_override': powerLevelContentOverride,
+      },
+    );
+
+    final roomId = response['room_id'] as String;
```
nqhhdev, line 30: Can we create a class object?  wdyt

nqhhdev, line 30: Should add exception handling

lib/config/first_column_inner_routes.dart
```diff
@@ -34,6 +34,18 @@ class FirstColumnInnerRoutes {
         } else {
           return _defaultPageRoute(const NewGroupChatInfo(contactsList: {}));
         }
+      case 'innernavigator/newfeed':
```
9clg6, line 37: Enum could be relevant ?

lib/config/go_routes/app_routes.dart
```diff
@@ -279,6 +279,14 @@ class InvitationLinkWebRoute extends GoRouteData with $InvitationLinkWebRoute {
                 ),
               ],
             ),
+            TypedGoRoute<NewPrivateChatNewFeedRoute>(
+              path: 'newfeed',
```
9clg6, line 283: Enum could be relevant ?

lib/domain/model/room/room_extension.dart
```diff
@@ -39,6 +40,12 @@ extension RoomExtension on Room {
     return !isSpace && !isStoryRoom;
   }
 
+  /// Drives wording: what a user is able to do in the
+  /// room is set by its power levels
+  bool get isFeed =>
+      getState(EventTypes.RoomCreate)?.content.tryGet<String>('type') ==
```
9clg6, line 46: Enum could be relevant ?

lib/pages/new_group/new_group.dart
```diff
@@ -35,9 +37,15 @@ class NewGroupController extends ContactsSelectionController<NewGroup> {
   void moveToNewGroupInfoScreen() async {
     if (!FirstColumnInnerRoutes.instance.goRouteAvailableInFirstColumn()) {
       context.pushInner(
-        'innernavigator/newgroupchatinfo',
+        widget.isFeed
+            ? 'innernavigator/newfeedinfo'
```
9clg6, line 41: Same question

lib/pages/new_group/new_group_chat_info.dart
```diff
@@ -195,7 +208,9 @@ class NewGroupChatInfoController extends State<NewGroupChatInfo>
         if (failure is CreateNewGroupChatFailed) {
           await showConfirmAlertDialog(
             context: context,
-            message: L10n.of(context)!.inviteUserErrorMessage,
+            message: failure.exception is FeedNotSupportedByHomeserverException
+                ? L10n.of(context)!.feedNotSupportedByHomeserver
```
9clg6, line 212: Refacto possible here for final l10n = L10n.of

lib/pages/new_group/new_group_chat_info_view.dart
```diff
@@ -65,14 +65,16 @@ class NewGroupChatInfoView extends StatelessWidget {
...
-                      onChanged: (value) {
-                        newGroupInfoController.toggleEnableEncryption();
-                      },
-                    ),
+                    if (!newGroupInfoController.isFeed) ...[
+                      const SizedBox(height: 16),
+                      _EncryptionSettingTile(
+                        enableEncryptionNotifier:
+                            newGroupInfoController.enableEncryptionNotifier,
+                        onChanged: (value) {
```
9clg6, line 73: Wildcard possible here

lib/pages/new_group/new_group_chat_info_view.dart
```diff
@@ -121,7 +123,9 @@ class NewGroupChatInfoView extends StatelessWidget {
     return PreferredSize(
       preferredSize: const Size.fromHeight(NewGroupChatInfoStyle.toolbarHeight),
       child: TwakeAppBar(
-        title: L10n.of(context)!.newGroupChat,
+        title: newGroupInfoController.isFeed
+            ? L10n.of(context)!.newFeed
```
9clg6, line 127: Refacto possible here

lib/presentation/mixins/go_to_group_chat_mixin.dart
```diff
@@ -11,4 +11,12 @@ mixin GoToGroupChatMixin {
       const NewPrivateChatNewGroupRoute().push(context);
     }
   }
+
+  void goToNewFeed(BuildContext context) {
+    if (!FirstColumnInnerRoutes.instance.goRouteAvailableInFirstColumn()) {
+      context.pushInner('innernavigator/newfeed');
```
9clg6, line 17: We can use static access ?

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
```diff
@@ -0,0 +1,39 @@
...
+  /// [CreateRoomPreset] is a closed Dart enum: the feed preset cannot be
+  /// custom.
+  ///
+  /// Throws a [MatrixException] on a homeserver that does not know the preset.
+  Future<String> createFeedRoom({
+    String? feedName,
+    List<StateEvent>? initialState,
+    Map<String, dynamic>? powerLevelContentOverride,
+  }) async {
+    final response = await request(
```
9clg6, line 16: Is it possible to create a code-gen system to avoid to write all of that by hand ?

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
9clg6: Yes, with MatrixRepository

lib/domain/usecase/room/create_new_group_chat_interactor.dart
jbrnzl: Hi, I now check MatrixError.M_BAD_JSON but tell me if we have other ideas, I'm not sure on how we can do better :)

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
jbrnzl: Hi @nqhhdev I think it is a bit overkill cause it's only for a single field. By the way I just checked (by curiosity) and I see that the matrix sdk is doing the same in `createRoom` :) Would be better if the sdk allows custom presets

jbrnzl: Hi @nqhhdev I think it's ok here cause again I took example of the `createRoom` from the sdk, and from what I see, the spec requires a room_id in 200 responses. Anyway CreateNewGroupChatInteractor should still catch errors also.

lib/data/datasource_impl/feed/feed_datasource_impl.dart
```diff
@@ -0,0 +1,61 @@
...
+      );
+
+      final roomId = response.tryGet<String>('room_id');
+      if (roomId == null) {
+        throw const FormatException('No room_id in the response');
+      }
+
+      // Navigating to the room fails if it has not reached the sync yet.
+      if (_client.getRoomById(roomId) == null) {
+        await _client.waitForRoomInSync(roomId, join: true);
```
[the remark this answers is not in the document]
jbrnzl: waitForRoomInSync only awaits firstWhere on the client's sync streams, and the SDK never pushes errors into them, so it can't fail with an Exception after the room has been created. Do you agree ?

lib/domain/model/room/room_extension.dart
jbrnzl: I deleted isFeed cause we do not need it yet, let's go with enum if re-implemented

lib/presentation/mixins/go_to_group_chat_mixin.dart
jbrnzl: @9clg6 About route constants ? I agree but maybe we can do this in an other PR

lib/utils/matrix_sdk_extensions/client_feed_extension.dart
jbrnzl: I updated, please let me know what you think about my implementation @9clg6 // @nqhhdev

lib/data/datasource_impl/feed/feed_datasource_impl.dart
```diff
@@ -0,0 +1,71 @@
...
+  final Client _client;
+
+  @override
+  Future<String> createFeed({String? name, String? avatarUrl}) async {
+    try {
+      final roomId = await _createFeedRoom(name: name, avatarUrl: avatarUrl);
+
+      // Navigating to the room fails if it has not reached the sync yet.
+      if (_client.getRoomById(roomId) == null) {
+        await _client.waitForRoomInSync(roomId, join: true);
```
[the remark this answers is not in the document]
jbrnzl: waitForRoomInSync doesn't reject: it only awaits firstWhere on the client's sync streams, and the SDK catches sync request errors instead of pushing them into those streams, as you confirmed in the previous thread. While sync is stopped, the wait just stays pending, and it completes once sync resumes, since the room already exists on the server

lib/data/datasource_impl/feed/feed_datasource_impl.dart
```diff
@@ -0,0 +1,71 @@
...
+    return roomId;
+  }
+
+  FeedException _toFeedException(Exception exception, StackTrace stackTrace) {
+    // The preset is the only non-standard field of this request, so a
+    // M_BAD_JSON means the homeserver does not know it.
+    if (exception is MatrixException &&
+        exception.error == MatrixError.M_BAD_JSON) {
+      Logs().w('FeedDatasourceImpl::createFeed', exception, stackTrace);
+      return const FeedNotSupportedByHomeserverException();
```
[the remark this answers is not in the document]
jbrnzl: On Synapse, the only M_BAD_JSON this request can hit is the unknown preset (_room_preset_config). The other M_BAD_JSON checks in room creation cover fields we don't send (room_version, power levels, retention, ACL)

nqhhdev (approved):
LGTM

lib/presentation/mixins/go_to_group_chat_mixin.dart
9clg6: Okay np

lib/config/first_column_inner_routes.dart
9clg6: @jbrnzl

9clg6 (approved):

PR #3348, opened 2026-09-13
ci(e2e): run the Android integration suite on parallel x86_64 emulators with an isolated Synapse

9clg6:
## Context

The Android functional E2E suite ran on Firebase Test Lab (`MediumPhone.arm`):

- **every FTL virtual Android device is ARM**, and FTL has recurring incidents of "execution delay on ARM based virtual devices" (confirmed on the status dashboard) — observed as 2–8 min device-allocation waits and shards cancelled without logcat;
- the shards were **serialized through a shared staging account** (`concurrency: matrix-ci-shared-account`), so parallel receiver logins failed with Matrix API rate limits;
- `--timeout 20m` / 25m watchdog plus `fail-fast: true` meant every shard hit the timeout and 4 of the 5 shards were hidden.

## Changes

- **GitHub-hosted x86_64 emulator (KVM)**: one job per shard, **in parallel**, without `fail-fast`. Build and run happen in the same job (`patrol test`).
- **Per-job isolated local Synapse** (same approach as the web workflow): `integration-server-synapse.sh` + `integration-test-provision-synapse.sh`; connection URLs rewritten to `10.0.2.2` (the emulator's host-loopback alias) and `TestUserServer=localhost`, so the shared staging account is no longer needed at all.
- **Scheme-aware mobile API helper** (HTTP supported for the local Synapse) and debug-only `usesCleartextTraffic`.
- **Mobile fixtures**: generic `prepareMobileRoomFixture`; migrated the scenarios that searched for the missing staged `SearchByTitle` room.
- **Test fixes**:
  - retry the message long-press until the menu opens (the app ignores long-presses while the event `isSending`);
  - remove the `/` from the "Pin/unpin" test name (it crashed `AndroidTestOrchestrator`, which uses the test name as a file name);
  - point the "non-existing" DM account at an unresolvable domain;
  - relax the `create_direct` room-count assertion (`<=`);
  - derive the `chat_dm_leave` homeserver from `CurrentAccount`;
  - poll for the receiver join before failing the fixture.
- **App fix**: make `ProfileInfoBodyView` scrollable (17/27 px RenderFlex overflow exposed by the profile test).
- **Local exclusions** (`requiresBackend` / `requiresHardware`, driven by `LOCAL_HARNESS`): contacts visibility (staging TOM `/_twake` backend), dm-leave (staging DM semantics), image-retry (hardware Wi-Fi toggle), copy (native clipboard), external-MXID search (federation). These stay covered on staging/web.

**Performance stays on physical FTL** (`oriole`) — unchanged.

## Result

Run **34713106012**: **5/5 shards green** (~24 tests), jobs 23–31 min in parallel (including ~15 min build), with no ARM incidents and no account contention.

## Note

This PR **includes the long-press fix from #3347**, which is therefore superseded and closed.

## Follow-ups (non-blocking)

- Consider mocking the `_twake` module to cover contacts visibility locally.
- Decide on a staging job for the backend-dependent tests.

## Summary by CodeRabbit

- **New Features**
  - Profile pages now scroll when content exceeds the screen height.
  - Mobile integration tests support dynamically created chat rooms and fixtures.
  - Integration tests can run in parallel across multiple Android emulator shards.
  - Integration testing supports local backend environments and configurable test targets.

- **Bug Fixes**
  - Improved reliability when opening message menus and waiting for chat participants.
  - Improved support for local HTTP-based test environments.
  - Updated test coverage for platform-specific, backend-dependent, and hardware-dependent behavior.


integration_test/robots/chat_group_detail_robot.dart
```diff
@@ -153,10 +153,18 @@ class ChatGroupDetailRobot extends CoreRobot
...
 
   Future<PullDownMenuRobot> openPullDownMenu(String message) async {
-    await $(
+    // Long-press is ignored while the event is still sending, so retry until
+    // the menu opens instead of pressing once right after the bubble appears.
+    final messageFinder = $(
       MessageContent,
-    ).containing(find.textContaining(message)).longPress();
-    await $.waitUntilVisible($(PullDownMenu));
+    ).containing(find.textContaining(message));
```
nqhhdev, line 160: add `findRichText: true`, same RichText/HtmlMessage path as `getText` above. Without it longPress can hang before the retry loop helps. Wdyt?

nqhhdev (approved):

9clg6:
## Review comments addressed (357b6a220)

| Finding | Change |
|---|---|
| `MemberMatrixID` missing → empty `User2MaxtrixAddress` | Both provisioning blocks now fail with `::error::` instead of writing an empty value |
| `patrol_test_target` interpolated into the shell script | Passed via step `env` (`PATROL_TEST_TARGET`) to avoid script injection |
| `find.textContaining` misses standalone `RichText` | `findRichText: true` added in `MessageMenuRobot` and `ChatGroupDetailRobot` long-press finders |
| Missing `.env.cicd` error pointed at the wrong recovery step | Message now references the fixture provisioning step |
| Matrix-address search only asserted a non-empty count | Now also asserts the fixture room title is returned |

All six CodeRabbit threads resolved. The remaining CodeScene `Bumpy Road Ahead` on `chat_group_open_profile_scenario.dart` is unrelated to this PR's changes and left as-is.

integration_test/robots/chat_group_detail_robot.dart
9clg6: You're right, thanks

nqhhdev (approved):
LGTM

jbrnzl (approved):

PR #3369, opened 2026-09-17
fix(contacts): restore ToM configuration from well-known for existing sessions

9clg6:
### Context
The ToM configuration (ToM server base URL, identity server, auth URL) is persisted locally **only at login** (`setUpToMServicesInLogin` → `_storeToMConfiguration`). `_retrieveLocalToMConfiguration` only *reads* it.

A valid Matrix session can outlive that local data (for example when the app container is reset while the session stays in the keychain, or after a corrupted database). The app then stays logged in but with `twakeSupported == false`.

### Problem
With no ToM configuration:
- the ToM `DynamicUrlInterceptors` has no base URL,
- every `/_twake/*` request is sent with an empty host and fails, e.g. `Invalid argument(s): No host specified in URI /_twake/addressbook`,
- and nothing ever retries/re-derives the configuration.

Concretely, the ToM address book (`GET /_twake/addressbook`) never loads, so the contacts list shows no ToM contacts and no merge, until the user logs out and logs in again. Pull-to-refresh cannot help because the base URL is never set.

### Solution
Re-derive the ToM configuration from the homeserver **well-known** discovery information (`.well-known/matrix/client` → `app.twake.chat` / `t.server`) when it is missing:

- new `MatrixState._ensureToMServicesFromSummary`, called at the end of `_getHomeserverInformation` (startup / active client change):
  - if the ToM interceptor has no base URL and the summary exposes a ToM server, re-run `setUpToMServices`, persist the configuration with `_storeToMConfiguration`, and refresh the address book so the merge appears without a manual refresh;
  - it returns early when the ToM server is already configured, so the normal login path is untouched.
- `GetTomContactsInteractor` now skips the address book request when no ToM server is configured, so a non-Twake homeserver no longer logs a bogus "No host specified" error.
- `GetTomContactsInteractor` is now also declared through a Riverpod provider (`lib/pages/contacts_tab/providers/contacts_providers.dart`), following the transitional pattern used for interactors that are still registered in GetIt (see `bootstrap_providers.dart`), so Riverpod consumers can `ref.read` it instead of reaching into GetIt.

This is generic: it fixes the configuration for **all** `/_twake/*` endpoints (address book, invitations, recovery words, user info), not only contacts.

### Impact
No behaviour change when the ToM configuration is present (early return). Only sessions that lost their local ToM configuration recover automatically.

### Test recommendations
- Normal login on a Twake homeserver: contacts (ToM + phonebook) still load, no extra requests (early return).
- Reproduce the fixed case: log in, then clear the local Hive `tomConfigurationsBox` (or restore a container without it) and restart → the app should log `_ensureToMServicesFromSummary: restoring ToM configuration from well-known` and reload the address book.
- On a non-Twake homeserver: no more `No host specified in URI /_twake/addressbook`.

### Notes
- Independent of the flutter_gen migration and of the contacts module rework.
- The well-known is already fetched by `_getHomeserverInformation`, so no extra network call is added.

## Summary by CodeRabbit

* **New Features**
  * Contact services now refresh automatically from the signed-in account’s server configuration.
  * Address-book synchronization can begin without requiring users to log in again when server settings change.
  * Contacts can be accessed through the app’s shared contact service integration.

* **Bug Fixes**
  * Contacts remain unavailable gracefully when no contact service is configured.
  * Contact synchronization no longer overwrites active updates.
  * Web message menu actions and reply interactions now appear more reliably.


jbrnzl (approved):
Tested, working well on my side!

nqhhdev (approved):

Issue #3357, opened 2026-09-16
Handling messages that exceed maximum character limit


tprudentova:
## Purpose

Allow users to send messages that exceed maximum character length by automatically splitting them into several messages.

## Entry Point

The user types in a message exceeding the maximum character limit and tries to send it.

## Behaviour

- The message is automatically split into several messages.
- Each part contains exactly the maximum number of characters possible, except for the final part.
- No characters should be added, removed, or modified during splitting.
- Messages are sent sequentially, in their original order (first part, second, third etc.).
- The user does not need to manually split the message.

### One part fails to send

- If one part fails, the remaining parts must still be sent according to their sending order.
- The failed part must display the existing message error state with the Retry option.
- Successfully sent parts remain sent and must not be resent automatically.

### Message length equals the limit

- A message exactly at the maximum character limit is sent as a single message.

### Message exceeds the limit by one character

- The message is split into two messages: one containing the maximum number of characters and one containing the remaining character.

## Design

- No new UI or design is required.
- Existing message bubbles and existing message error states are reused.

## Technical requirements

- The maximum character limit must use the existing limit already defined for messages (please, define it in the ticket below).
- Parts must not be sent in parallel if doing so can result in them appearing out of order.
- Existing message sending and error handling should be reused where possible.

## Acceptance Criteria

- [x] A message exceeding the maximum character limit is automatically split into multiple messages.
- [x] Each resulting message contains no more than the maximum allowed number of characters.
- [x] Splitting follows a strict character limit.
- [x] Each part contains the maximum possible number of characters, except the final part.
- [x] No characters are added, removed, or modified during splitting.
- [x] Message parts are sent sequentially in their original order.
- [x] A message exactly at the maximum character limit is sent as one message.
- [x] A message exceeding the limit by one character is split into two messages.
- [x] If one part fails to send, the other parts are still sent according to the sending order.
- [x] A failed part displays the existing message error state.
- [x] Successfully sent parts remain sent and are not resent automatically.
- [x] The failed part retains its original content, includes a Retry button and can be resent.

jbrnzl:
Criteria 7 is in my opinion in contradiction with criteria 10. I did implement the 10th instead, that seems way natural to me

jbrnzl:
> The maximum character limit must use the existing limit already defined for messages (please, define it in the ticket below)

Technically the only constraint is 65k bytes on an event on synapse side.
[...]

PR #3370, opened 2026-09-17
TW-3357: split long messages

### LINKED ISSUE #3031: [BUG] Can not send a message Error in main - Instance of 'minified:Z0'

### Describe the bug

I copied / pasted a big JSON to @jbrnzl , but then, I got the "click to retry to send". And I got this error in my console : "!!!CRITICAL!!! [Matrixl Error in main - Instance of 'minified:Z0'" 

Image

I can't share the json output because there is a lot of Personal informations in it. 

### Pre-Bug State

.

### To Reproduce

.

### Expected results

.

### Context

- Device: chrome
- Server: linagora.com

### Screenshots or Video

<details open>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### Additional information

- Frequency
- Others

jbrnzl:
## Ticket
Implements #3357 
Fixes #3031

## Solution
A text longer than 4096 characters is now split into several messages. Before, it ended as an error bubble with the ability to retry but with the error `"!!!CRITICAL!!! [Matrixl Error in main - Instance of 'minified:Z0'"`. Which is staying without way to fix as it's related to the message size.

image

- `MessageSplitter` cuts the text every 4096 characters. It only cuts between graphemes, so an emoji or an accented letter is never broken. **That is why `characters` becomes a direct dependency**.
- `SendTextMessageInteractor` sends the parts one by one and waits for each one before sending the next, so the order is kept. If a part fails it keeps the existing error state with its "Retry button", and the next parts are still sent. The reply is attached to the first part only. Edits and commands are not split.

**Why 4096 ?**: [EXPLANATION HERE](https://github.com/linagora/twake-on-matrix/issues/3357#issuecomment-5713025989)

## Impact description

- Each part is a normal message: n parts means n notifications, and markdown cut in the middle of a link or a code block renders differently in the two parts
- An edit longer than the limit still fails like before

## Test recommendations

- Send a very long text with emojis and check nothing is broken at the cuts.
- Reply to a message with a long text: only the first part shows the reply.
- Go offline while the parts are being sent: the failed part shows Retry, and retrying only resends that part.

## Demo

https://github.com/user-attachments/assets/e6745558-2601-4171-80c7-647077f90323

## Summary by CodeRabbit

* **New Features**
  * Long text messages are automatically split into sequential messages when they exceed the 4,096-character limit.
  * Splitting preserves emojis, flags, and other combined characters without breaking them.
  * Replies remain attached to the first message segment.
  * Edited messages and recognized commands continue to be sent as a single message.
  * Unrecognized commands are safely handled as plain text when messages are split.


lib/domain/usecase/room/send_text_message_interactor.dart
```diff
@@ -0,0 +1,56 @@
...
+        text,
+        inReplyTo: inReplyTo,
+        editEventId: editEventId,
+        parseCommands: isCommand,
+      );
+      return;
+    }
+
+    final List<String> parts = splitter.split(text);
+    for (final (int index, String part) in parts.indexed) {
```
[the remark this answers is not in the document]
jbrnzl: I think this edge case is OK...I mean who will ever complain about that

lib/domain/usecase/room/send_text_message_interactor.dart
```diff
@@ -0,0 +1,56 @@
...
+        );
+      } on Exception catch (error, stackTrace) {
+        // The part is already in error in the timeline, keep going.
+        Logs().e(
+          'SendTextMessageInteractor::execute(): part '
+          '${index + 1}/${parts.length} failed in room ${room.id}',
+          error,
+          stackTrace,
+        );
+      }
```
nqhhdev, line 47: A second message can be sent while this loop is awaiting each part, so it may appear between parts of a long message. Could sends be serialized per room or the composer disabled until splitting finishes?

Example: the user sends a long message split into parts A, B, and C, then sends “Thanks” before the split finishes. “Thanks” could appear between A and B.

Wdyt? Could you please add a small demo for this case?

lib/domain/services/message_splitter.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:characters/characters.dart';
+
+/// Splits a text into parts of at most [maxLength] UTF-16 code units, cutting
+/// only between grapheme clusters so an emoji is never broken. A grapheme
+/// longer than [maxLength] stays whole in its own part.
+final class MessageSplitter {
```
9clg6, line 6: I don't think it's necessary to have a usecase here, maybe a tool is enough

lib/domain/usecase/room/send_text_message_interactor.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:matrix/matrix.dart';
+import 'package:twake_chat/domain/usecase/room/message_splitter.dart';
+
+/// Sends a composer text, split into several messages when it is longer than
+/// [MessageSplitter.maxLength]. Parts are sent one after the other, the reply
+/// goes on the first one, and a failed part does not stop the next ones.
+/// Edits and commands are never split.
+final class SendTextMessageInteractor {
+  const SendTextMessageInteractor({required this.splitter});
```
9clg6: If you use a tool instead of a interactor, you can free the constructor from his parameter

lib/pages/chat/providers/chat_providers.dart
```diff
@@ -0,0 +1,12 @@
...
+import 'package:twake_chat/domain/usecase/room/message_splitter.dart';
+import 'package:twake_chat/domain/usecase/room/send_text_message_interactor.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'chat_providers.g.dart';
+
+@riverpod
+SendTextMessageInteractor sendTextMessageInteractor(Ref ref) =>
+    const SendTextMessageInteractor(
+      splitter: MessageSplitter(maxLength: AppConfig.maxMessageLength),
```
9clg6: If maxMessageLength is a const maybe it could be more relevant to use the const directly in the tool and not from a parameter

lib/domain/usecase/room/send_text_message_interactor.dart
jbrnzl: Well I agree, I updated to take this scenario into consideration even if it is minor in my opinion. I will not be able to add a demo for this case as it is very specific and difficult to record

lib/domain/usecase/room/send_text_message_interactor.dart
jbrnzl: I moved it out of usecase but I think we should keep the injection, cause it lets me test the interactor with a custom maxLength. I took example on https://github.com/linagora/twake-flutter-guidelines/blob/main/GUIDELINES.md#26-core--standalone-modules

lib/pages/chat/providers/chat_providers.dart
jbrnzl: Same for this one I prefer to keep the param to test it easily

9clg6 (approved):

nqhhdev (approved):

PR #3373, opened 2026-09-17
fix(web): restore asset caching with our own service worker

9clg6:
Restores https://github.com/linagora/twake-on-matrix/pull/3316 after it was reverted in https://github.com/linagora/twake-on-matrix/pull/3372.

cc @Crash-- @jbrnzl @pm-McFly @nqhhdev @9clg6

Same change as #3316: precache via `web/push_sw.js`, manifest generated at build time by `scripts/generate-sw-manifest.py`.

## Summary (from #3316)

- Move precaching into `web/push_sw.js`, alongside the existing Web Push handlers, driven by a path/MD5 manifest generated at build time by `scripts/generate-sw-manifest.py`.
- Stop unregistering service workers on every load in `web/index.html`, and register this one instead.
- Exclude `config.json` from the precache manifest, along with source maps, pre-compressed `.gz` twins and dotfiles.

Open point from #3316: @pm-McFly preferred a bash-generated never-cached manifest over the Python script and artifact injection. That is still unresolved and should be discussed here before merge.


pm-McFly:
@Crash-- I try since yesterday to make the SW work. I have built the Docker image and I am running a local docker compose to validate.

But, it seems that each time I F5 the app, all assets are re-downloaded anyway.

First app boot:

image

After F5 the page:

image

pm-McFly:
Okay after tinkering a bit and checking the browser settings, I see that I only get 304 on reload

image

jbrnzl:
> Okay after tinkering a bit and checking the browser settings, I see that I only get 304 on reload
> 
> image

So that's some good news!!
What you guys @9clg6 @nqhhdev think about the discussion on https://github.com/linagora/twake-on-matrix/pull/3316 ?
What's the simplest option ?

